PR #4, opened 2025-05-23
Fix silent data loss risk in IndexedDBStore

jacwright:
## Summary
- Add warning when snapshot creation is skipped due to old pending changes
- Force snapshot creation if too many committed changes accumulate (3x threshold)
- Prevents unbounded memory growth when pending changes prevent normal cleanup

## Test plan
- [x] Existing IndexedDB tests pass
- [x] Verified warnings are logged and forced snapshots are created when needed

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/client/IndexedDBStore.ts
+++ b/src/client/IndexedDBStore.ts
@@ -306,6 +306,28 @@ export class IndexedDBStore implements PatchesStore {
           }),
           committedChanges.delete([docId, 0], [docId, lastRev]),
         ]);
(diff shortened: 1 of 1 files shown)


jacwright:
Closing this PR as the approach was incorrect. The force snapshot creation mixed pending and committed state, which would cause data corruption. Creating a new PR with the proper fix that synchronizes rebased pending changes back to the store.

PR #22, opened 2026-04-08
Add updateBranch to PatchesBranchClient

jacwright:
## Summary
- Exposes `updateBranch(branchId, metadata)` on `PatchesBranchClient` for updating branch metadata (e.g. name)
- The server, transport, and sync layers already supported this; only the client-facing class was missing it
- Optimistically updates local `branches` state after the API call

## Test plan
- [x] Added tests for both online (BranchAPI) and offline (BranchClientStore) modes
- [x] All 26 tests pass
- [x] Build passes


GravityRabbitTwo (approved):
Looks good to me.

PR #34, opened 2026-04-21
fix(client): pure-echo skip + import() preserves optimistic ops + LWW content-keyed echo detection

RyanHavoc:
## Summary

Three coordinated client-side fixes to keep optimistic UI updates from getting stomped during sync. Together they eliminate the visible "text jumps mid-typing" bug observed in the dabble-writer app.

### 1. Pure-echo state-emit skip (original commit)

`OTDoc.applyChanges` and `LWWDoc.applyChanges` were emitting a redundant subscriber notification when the server broadcast back our own committed change. Easy-signal compares with `===`; `_recomputeState()` always built a fresh top-level object even when the data hadn't changed, triggering Vue stores to push the (now stale) state back into focused inputs.

Detect "every server change in this batch is a confirmation of one of our own pending changes (no foreign concurrent ops, no fresh local changes mixed in)" and skip the state setter in that case.

### 2. `import()` preserves outstanding optimistic ops (new)

`OTDoc.import()` and `LWWDoc.import()` were resetting state and wiping `_optimisticOps = []`. Sync recovery and cross-tab snapshot broadcast paths legitimately call `import()` while a user is mid-typing — wiping in-flight ops silently regressed the input back to the snapshot value.

`import()` now:
- ignores stale snapshots whose `rev < _committedRev` (caller is behind us);
- preserves `_optimisticOps` and re-applies them on top of the imported state via `applyPatch(..., { strict: true })`;
- drops any optimistic op that fails to re-apply (e.g. parent path no longer exists in the new snapshot).

This is a strict improvement — every existing caller of `import()` either has no optimistic ops (initial load) or wants them preserved (recovery, broadcast).

### 3. LWW echo detection keyed on op content, not change id (new)

`LWWDoc._inFlightChangeIds: Set<string>` only worked when the server-confirmed Change id matched the one we tracked locally. But `LWWAlgorithm.getPendingToSend()` builds a *fresh* sending Change with a new id when packaging pending ops — the server commits and echoes that fresh id back, which never matched. Result: every server echo on an LWW doc fell through to recompute and emitted a spurious notification.

Replaced with `_inFlightOpKeys: Set<string>` keyed by `JSON.stringify(op)` (op content). The LWW algorithm preserves original timestamped ops end-to-end, so content fingerprints survive the change-id reissue. Orphan keys (ops consolidated away in the algorithm's pending store) are bounded by session length and act as harmless extra positives. False-positive risk requires a foreign client to emit byte-identical ops including the `ts` value — astronomically unlikely.

## Files

- `src/client/OTDoc.ts` — pure-echo skip in `applyChanges`; `import()` preservation + rev guard.
- `src/client/LWWDoc.ts` — content-keyed `_inFlightOpKeys` with pure-echo skip; `import()` preservation + rev guard.
- `tests/client/OTDoc.spec.ts` — pure-echo batch coverage (4 tests) + new `import preserves optimistic ops` block (3 tests: preserve on fresher snapshot, drop incompatible op, ignore stale rev).
- `tests/client/LWWDoc.spec.ts` — pure-echo batch coverage (5 tests) + new `import` cases (3 tests) + LWW reissued-id regression test that triggered bug #3 (1 test).

## Test plan

- [x] `npm test` — 1863/1863 pass (was 1856 before this PR's additions).
- [x] `npm run type:check` clean.
- [x] `npm run lint` clean.
- [x] `npm run format` clean.
- [x] `npm run build` produces clean dist (verified `_inFlightOpKeys` present, no debug logging).
- [x] Smoke-tested via `npm link` against dabble-writer worktree: rapid typing in a doc title field no longer jumps; cross-tab snapshot rebroadcast no longer wipes in-flight characters.


RyanHavoc:
## Self-review (sanity check)

Re-read the diff against the surrounding code (`OTDoc`, `LWWDoc`, `BaseDoc`, `LWWAlgorithm.handleDocChange`, `LWWAlgorithm.getPendingToSend`, `consolidateOps`). Overall the three fixes are sound, but I want to flag a few things I missed in the original write-up:

### Issue 1 — LWW echo skip silently bypassed for combinable ops (@inc / @bit / @max / @min)

The PR description claims op CONTENT survives the change-id reissue end-to-end. That's true for `replace` (the only case the new tests cover), but **not** for combinable ops.

Trace for two `@inc 1` ops typed rapidly:

1. First `@inc 1` → `handleDocChange` → `timedOps = [{op:'@inc', value:1, ts:T1}]`. `existingOps=[]`, so `consolidateOps` is a passthrough. Doc receives the timedOp; `_inFlightOpKeys` gets `{@inc,1,T1}`.
2. Second `@inc 1` → `timedOps = [{op:'@inc', value:1, ts:T2}]`. `existingOps=[{@inc,1,T1}]`. `consolidateFieldOp` returns `{...incoming, op, value: combine(1,1)} = {@inc, value:2, ts:T2}`. Store now holds the **combined** op. Doc receives the **original** timedOp `{@inc,1,T2}`; `_inFlightOpKeys` gets `{@inc,1,T1}` and `{@inc,1,T2}`.
3. `getPendingToSend` reads `pendingOps = [{@inc, value:2, ts:T2}]` from the store and sends that.
4. Server echoes `{@inc, value:2, ts:T2}`. We check `_inFlightOpKeys.has({@inc,2,T2})` → **false**. `isPureEcho = false`. Recompute fires.

Net effect: spurious recompute persists for any sequence of combinable ops. Not a regression (was already broken), and the symptom is much less visible than `replace` because combinable ops are rarely user-keystroke-driven, but the PR doesn't actually fix this case despite the framing.

Two options:
- (a) Acknowledge the limitation in the docstring on `_inFlightOpKeys` and add a vitest case showing the spurious recompute is *expected* for `@inc` so future-me doesn't think it's a regression.
- (b) Fix it by also tracking the **post-consolidation** op key — e.g. have `LWWAlgorithm.handleDocChange` pass the consolidated `opsToSave` to a separate `addInFlight()` method on the doc, in addition to passing `timedOps` via `applyChanges`.

I'd lean (a) — (b) leaks algorithm internals into the doc API and the symptom barely matters in practice.

### Issue 2 — Orphan key accumulation in `_inFlightOpKeys` for typing into the same field

For a normal typing burst into a title field:

- Type 'h' → `_inFlightOpKeys` += `{replace,'h',T1}`.
- Type 'he' → store-side consolidation **replaces** the prior `{replace,'h',T1}` with `{replace,'he',T2}`. Doc-side, we add `{replace,'he',T2}` but never remove `{replace,'h',T1}`.
- Type 'hel' → similar; `{replace,'h',T1}` and `{replace,'he',T2}` are now both orphans.

When the server echoes `{replace,'hel',T3}`, we delete that key. The two earlier orphans linger forever (within the session).

The PR doc says "bounded by session length and act as harmless extra positives." That's true for correctness — the `ts` requirement makes false-positive matches astronomically unlikely. But for a long Dabble writing session (hours of typing into a manuscript) this could grow to tens of thousands of entries per doc. JSON-stringified op keys are ~50–100 bytes each; not catastrophic but not great either.

Cheap mitigation: cap `_inFlightOpKeys.size` and evict the oldest N when over the cap. Even simpler: clear it whenever `_optimisticOps.length === 0 && hasPending === false` (steady state — nothing in flight, safe to drop everything).

### Issue 3 — `OTDoc.applyChanges` echo-skip relies on `[].every() === true` non-being-triggered

```169:170:src/client/OTDoc.ts
      const priorPendingIds = new Set(this._pendingChanges.map(c => c.id));
      const isPureEcho = serverChanges.every(c => priorPendingIds.has(c.id));
```

`serverChanges` is currently always non-empty when this branch is reached (`changes[0].committedAt > 0` is the entry condition; `serverEndIndex` is then either −1 or ≥ 1). So today `isPureEcho` can't be a vacuous true. But if anyone ever refactors the slicing above, `[].every()` would silently skip a recompute.

Add a defensive `serverChanges.length > 0 &&` to the `isPureEcho` expression, or add a `// invariant: serverChanges.length > 0` comment on the line above. Cheap insurance.

### Things that are correct but I want to record I checked:

- `OTDoc.import` preserving `_optimisticOps` while wiping `_pendingChanges = snapshot.changes` — correct in practice because the algorithm-side `pendingOps` store is the source of truth and survives independently. Verified by tracing the case where Tab A imports a fresh snapshot from Tab B mid-typing: state ends up as `newSnapshot + optimisticOps`, and the eventual local-confirm rebuilds `_pendingChanges` properly.
- `JSON.stringify(op)` key stability — relies on V8/JSC consistent insertion-order key serialization for the same op-construction code path. The LWW algorithm always builds ops with the same field order (`{ ...op, ts: timestamp }` in `handleDocChange`, and the store round-trips through IDB JSON which preserves insertion order). Safe.
- LWW pure-echo cross-tab behavior — Tab B's `_inFlightOpKeys` gets populated when the hub broadcasts Tab A's local change as a `committedAt:0` event (line 184–186 fires unconditionally). So when the server echo arrives at Tab B, `isPureEcho` works there too. Nice emergent property.

### Summary

Land it, but I'd want issues 1 and 3 addressed before tagging a release — at minimum a comment + test case for issue 1 to lock in the limitation, and the defensive non-empty check for issue 3. Issue 2 is fine to defer until someone reports memory growth in a long Dabble session.

RyanHavoc:
Self-review follow-up commit (32bb2ab) addresses three of the findings I flagged:

**Combinable-op echo blind spot — actually fixed, not just documented.** Added an optional `{ inFlightOpsOverride, retiredInFlightOps }` options bag to `LWWDoc.applyChanges`. `LWWAlgorithm.handleDocChange` now passes `opsToSave` (the consolidated ops that actually ship) as the override and the prior pending op at each affected path as `retiredInFlightOps`. The in-flight tracker now matches what the server will echo for `@inc`/`@bit`/`@max`/`@min`, and prior orphan keys at consolidated paths are pruned in the same call.

**Orphan key accumulation — fixed by the same mechanism.** `retiredInFlightOps` retires the previous key at each path on every consolidation step, so memory is bounded by O(distinct paths with pending writes) instead of O(keystrokes per session).

**`[].every() === true` defensive guard — added to OTDoc.** Currently redundant (the outer branch only runs when `changes[0].committedAt > 0`, guaranteeing at least one server change) but defends against future refactor weakening that invariant. Comment explains why. LWWDoc's analogous check is already gated by `hasServerChanges` so no additional guard needed.

**Tests added** in `tests/client/LWWDoc.spec.ts`:
- `detects pure echo for combinable ops via inFlightOpsOverride (consolidated ops)` — drives `@inc 1 + @inc 1` → server-echoes `@inc 2`, asserts no subscriber callback fires.
- `retiredInFlightOps prunes orphan keys to bound memory under fast typing` — types four chars into one field, asserts the latest echo skips AND that an orphan-aged echo falls through (proving the retirement actually mutated the set).

External callers (Worker-Tab sync, tests) leave `options` undefined and get the legacy `c.ops`-based behaviour unchanged.

All 1865 patches tests pass; type/lint/format clean.

Process lesson logged to memory.md: when introducing a generic-looking mechanism, write a test per op-class (replace / combinable / soft / remove) before claiming it's general.

matalina (approved):

PR #38, opened 2026-05-05
DAB-472: WebRTC awareness over SSE+REST via PatchesREST multiplexing

jacwright:
## Summary

WebRTC signaling can now ride on the same SSE stream as document sync. Apps using `PatchesREST` get presence without opening a second `EventSource` or pulling in WebSocket just for the handshake.

- `WebRTCTransport` constructor widened from `WebSocketTransport` to a structural `SignalingTransport` (`ClientTransport` + `connect`/`state`/`onStateChange`).
- New `PatchesRESTSignalingTransport` adapts a `PatchesREST` instance as a `SignalingTransport`, sending via `POST /signal/:clientId` and receiving via a new `signal` SSE event.
- New `SSESignalingService` (server-side) sends signaling frames through `SSEServer.sendToClient()`, which writes client-targeted events bypassing subscription routing. Unlike `notify()`, signaling is **not buffered through expiry** — stale handshakes (gone peers, moot ICE candidates) are deliberately dropped.
- `SignalingService` moved to `src/net/signaling/` since it's no longer WebSocket-specific; the old import path re-exports for backward compatibility.

## Acceptance criteria

- [x] An app can construct `WebRTCAwareness` with a `PatchesRESTSignalingTransport` and have presence working between two browsers, no WebSocket involved.
- [x] Signaling rides on the same SSE stream as document sync (no second `EventSource`).
- [x] `PatchesWebSocket` + `WebSocketSignalingService` path still works (no regression — all 1898 existing tests pass).
- [x] `clientId` is shared between SSE doc sync and signaling, so peer addressing is consistent.
- [x] Tests cover SSE-side signal relay, signal POST endpoint handling, `WebRTCTransport` constructed with a non-WebSocket transport.
- [x] `awareness.md` and `README.md` updated with the SSE+REST wiring example alongside the WebSocket one.

## Multi-tab / hub note (out of scope, captured for context)

This implementation gives **one peer per `PatchesREST` instance** = one peer per browser when used behind a tab-election hub. That's the right model: a human has one cursor and one keyboard, so showing per-tab presence is fiction, not feature. Apps that want different behaviour can choose later (e.g. per-tab WebSocket awareness alongside SSE doc sync), but the data path here doesn't enforce a policy.

## Test plan

- [x] `npm run type:check`
- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm run test` (1898 passing, 12 new)
- [ ] Manual verification with two browsers using `PatchesREST` once a consumer wires up the `POST /signal/:clientId` route in `pup`.


docs/awareness.md
```diff
@@ -218,7 +254,35 @@ websocketServer.on('connection', async ws => {
...
+const clientId = req.auth.clientId;
+const stream = sse.connect(clientId, req.headers['last-event-id']);
+await signaling.onClientConnected(clientId);
+
+req.on('close', async () => {
+  sse.disconnect(clientId);
+  await signaling.onClientDisconnected(clientId);
+});
+
+// POST /signal/:clientId — raw JSON-RPC body. The `fromId` MUST come from auth,
```
RyanHavoc: **Security:** this example takes `clientId` straight from the URL and trusts it. Unlike the WebSocket flavor (where `clientId` is the server-assigned return from `onClientConnected()` and stays bound to the socket), here an authenticated user can pass *any* other user's `clientId` and forge `peer-signal` traffic on their behalf — including spoofing `from: <victimClientId>` to a third party. WebRTC SDP/ICE forgery isn't catastrophic alone, but combined with a coercion-to-TURN attack it's a plausible path to IP/relay metadata leaks and DoS against in-flight handshakes.

The docs are how integrators learn the contract. Suggest making the auth check explicit in the example, e.g.:

```ts
if (authedClientId !== req.params.clientId) return res.status(403).end();
await signaling.handleClientMessage(req.params.clientId, body);
```

And calling out in prose that this is a security requirement, not optional.

src/net/rest/SSESignalingService.ts
```diff
@@ -0,0 +1,44 @@
...
+ * app.post('/signal/:clientId', async (c) => {
+ *   const fromId = c.get('auth').clientId; // authenticated, not URL-derived
+ *   const body = await c.req.text();
+ *   const handled = await signaling.handleClientMessage(fromId, body);
+ *   return c.json({ ok: handled });
+ * });
+ *
+ * // On SSE stream close:
+ * await signaling.onClientDisconnected(clientId);
+ * ```
```
RyanHavoc, line 34: `sendToClient` returns `boolean` but it's discarded here. Combined with `SignalingService.clients` being a duplicate of the SSE liveness state (kept in sync only via the framework's explicit `onClientConnected` / `onClientDisconnected` calls), the two can drift: a client may still be in the set while `client.writer` is already null.

The consequence in `SignalingService.handleClientMessage`:

1. `clients.has(to)` passes.
2. `await this.send(to, outbound)` silently drops the frame.
3. `await this.respond(fromId, id, 'ok')` lies that the relay succeeded.

Two options:

- Bubble the boolean up so `handleClientMessage` can `respondError(fromId, id, 'Target unreachable')`.
- Or have `SSESignalingService.send` call `onClientDisconnected(id)` when `sendToClient` returns false, so the set self-heals and other peers get the standard disconnect notification.

A third (cleaner) alternative: override `getClients()` in `SSESignalingService` to derive from `sse.getConnectionIds()`, eliminating the drift class entirely. The base class already supports overriding `getClients`/`setClients`.

src/net/rest/SSESignalingService.ts
```diff
@@ -0,0 +1,44 @@
...
+ * URL `:clientId`, client A can POST to `/signal/B` and impersonate B in the
+ * WebRTC mesh, redirecting peer connections. Treat the URL parameter as
+ * untrusted input and bind sender identity to auth.
+ *
+ * ```typescript
+ * const sse = new SSEServer();
+ * const signaling = new SSESignalingService(sse);
+ *
+ * // GET /events/:clientId — after creating the SSE stream:
+ * const clientId = c.get('auth').clientId; // authenticated, not URL-derived
```
RyanHavoc: Minor: this JSDoc example uses Hono style (`return c.json({ ok: handled })`) and returns a body, while the `docs/awareness.md` example uses Express style (`res.status(204).end()`) with no body. Pick one. Also note that `204 No Content` shouldn't have a body, but `c.json({...})` does — so the two examples disagree on the response shape, not just the framework. Mixed signals for new integrators.

src/net/signaling/SignalingService.ts
```diff
@@ -0,0 +1,171 @@
...
+    const welcome: JsonRpcRequest = {
+      jsonrpc: '2.0',
+      method: 'peer-welcome',
+      params: {
+        id,
+        peers: Array.from(this.clients).filter(pid => pid !== id),
+      },
+    };
+
+    this.send(id, welcome);
```
RyanHavoc: Fire-and-forget. The base contract says `send` is `void | Promise<void>`, so this swallows any rejection from an async implementation. Synchronous in `SSESignalingService` today so it's harmless, but a future async implementation (queue-backed, batched, retrying) would have unhandled rejections here. Pre-existing in the moved code, but worth fixing while the file is being relocated — `await this.send(id, welcome);` is a one-character change.

src/net/rest/PatchesREST.ts
```diff
@@ -347,6 +363,33 @@ export class PatchesREST implements PatchesConnection {
...
+   * POSTs a raw JSON-RPC string to `/signal/:clientId`. Used as the upstream
+   * half of the multiplexed signaling channel: receive happens via the `signal`
+   * SSE event (see {@link onSignal}).
+   *
+   * The body is sent verbatim — callers pass an already-stringified JSON-RPC
+   * message. The endpoint forwards it to {@link SignalingService.handleClientMessage}.
+   */
+  async sendSignal(raw: string): Promise<void> {
+    const headers = await this._getHeaders();
+    const response = await globalThis.fetch(`${this._url}/signal/${this.clientId}`, {
```
RyanHavoc, line 386: `sendSignal` POSTs regardless of `_state`. If `WebRTCTransport` tries to send before `patches.connect()` resolves, the server has no SSE client to relay *to* — the message reaches `SignalingService.handleClientMessage` but the sender's clientId isn't in the set yet, so the relay fails silently (or, with the fix from the `SSESignalingService.send` comment, gets a clearer error).

Not a real bug given the documented call order (`await patches.connect()` before `awareness.connect()`), but cheap insurance: either early-return / throw when `_state !== 'connected'`, or document the precondition on the JSDoc.

tests/net/rest/SSEServer.spec.ts
```diff
@@ -229,6 +229,69 @@ describe('SSEServer', () => {
...
+      const decoder = new TextDecoder();
+      // Skip the initial retry: message
+      await reader.read();
+      // Read with a microtask boundary — there should be no further events.
+      const racer = new Promise<{ done: boolean; value?: Uint8Array }>(resolve => {
+        Promise.resolve().then(() => resolve({ done: true }));
+      });
+      const result = await Promise.race([reader.read(), racer]);
+      reader.releaseLock();
+      if (!result.done && result.value) {
```
RyanHavoc: This conditional makes the assertion a placebo: if the microtask wins the race (`result.done === true`), the `if` block doesn't run and the test passes without asserting anything — so even if signaling events *were* being replayed, this test could green. Two cleaner options:

- Drop the conditional and assert `expect(result.done).toBe(true)` (the buffer should be empty, so the read should not produce a value before the microtask resolves).
- Or skip the race entirely: assert against the buffer state directly (e.g. expose a length getter for tests, or check that no `event: signal` line appears in a fixed-size read of N bytes after reconnect).

The rest of the `sendToClient` test block is solid — this is just the one test that's load-bearing for "signaling is not buffered" and currently isn't actually load-bearing.

RyanHavoc (commented):
Inline comments below. Top-level take: architecture is right and tests cover the happy paths. Main blocker for me is the auth gap in the docs example for `POST /signal/:clientId` — see the comment on `docs/awareness.md`. The `send` return-value swallowing on `SSESignalingService` is also worth fixing now while the abstraction is fresh. The rest is polish.

src/net/rest/SSESignalingService.ts
jacwright: Took option (c) in 4c6833a — `SSESignalingService` now overrides `getClients()` to return `new Set(this.sse.getConnectionIds())` and `setClients()` as a no-op. Eliminates the drift class entirely: `handleClientMessage` will see fresh writer-liveness state and respond with 'Target not connected' instead of relaying into the void.

One adjustment in the base class: line 49 was reading `this.clients` directly for the welcome's peer list, which would have stayed empty under the override (since `setClients` is a no-op). Changed to `Array.from(clients)` (local var from `getClients`) so the override actually composes. No semantic change for the existing default impl.

src/net/rest/SSESignalingService.ts
jacwright: Done in 4c6833a — both examples now use Express style with `res.status(204).end()`. JSDoc body-and-status mismatch fixed at the same time.

src/net/signaling/SignalingService.ts
jacwright: Done in 4c6833a — `await this.send(id, welcome);`.

src/net/rest/PatchesREST.ts
jacwright: Went silent rather than throw, in 4c6833a. Reasoning: `sendSignal` is async, so a `throw` becomes a rejected promise that `JSONRPCClient.call` doesn't await — it'd surface as an unhandled rejection rather than rejecting the call. Silent also matches the `sendToClient` server-side behavior (we already chose not to buffer signaling because stale frames are harmful). Documented the precondition and the rationale on the JSDoc.

tests/net/rest/SSEServer.spec.ts
jacwright: Done in 4c6833a — replaced the conditional with `expect(result.done).toBe(true)`. If buffered signaling regresses, the read will win the race against the microtask and the assertion will fail.

docs/awareness.md
jacwright: Done in 4c6833a — the `if (req.auth.clientId !== req.params.clientId) return res.status(403).end();` check is now explicit in both code blocks, with the security note rewritten as a non-optional requirement.

RyanHavoc (approved):

PR #42, opened 2026-05-14
DAB-491: canonicalize @txt op.value as Op[] array

jacwright:
## Summary

Fix an inconsistency where `@txt` op `value` could land in three shapes (array, `Delta` instance, `{ ops }`), causing `text.transform` to silently drop ops and crashing the 3→2 migration sync with `TypeError: ops.map is not a function`.

- `JSONPatch.text()`, `text.compose()`, and `text.invert()` now emit a bare `Op[]` array (the canonical shape).
- `text.apply()` and `text.transform()` defensively accept all three input shapes via a shared `toOps()` helper, so existing serialized data still works.
- Documents at the target path (the result of *applying* `@txt`) remain `{ ops: [...] }` — that's the doc, not the op `value`.

Linear: DAB-491

## Test plan

- [x] `npm test` (1883 tests pass)
- [x] `npm run type:check`
- [x] `npm run lint`
- [x] `npm run format:check`
- [x] Updated `proxy.spec.ts` and `compose.spec.ts` to the array form; added a compose regression test covering legacy `{ ops }` input.

### DIFF
--- a/src/json-patch/JSONPatch.ts
+++ b/src/json-patch/JSONPatch.ts
@@ -22,6 +22,7 @@ import { applyPatch } from './applyPatch.js';
 import { composePatch } from './composePatch.js';
 import { invertPatch } from './invertPatch.js';
 import { bitmask } from './ops/bitmask.js';
+import { toOps } from './ops/text.js';
 import { transformPatch } from './transformPatch.js';
 import type { ApplyJSONPatchOptions, JSONPatchOp, JSONPatchOpHandlerMap } from './types.js';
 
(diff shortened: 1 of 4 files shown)


RyanHavoc:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

RyanHavoc (approved):

PR #47, opened 2026-05-24
DAB-503: Add Patches.applySnapshot and PatchesDoc.flush public APIs

jacwright:
## Summary

Two new public APIs in `@dabble/patches` to support transports that deliver snapshots out-of-band from `openDoc`:

- **`Patches.applySnapshot(docId, snapshot)`** — handles the open / opening / closed states consistently so transports don't have to. Closes a race where a snapshot arriving during an in-flight `openDoc` was silently dropped (the multi-tab hub-spoke and WebRTC mesh cases). Opening docs stash the snapshot (highest-rev wins for out-of-order delivery); open docs import when `snapshot.rev > committedRev`; unknown docs drop. PatchesSync's two snapshot routes now go through this method instead of `getOpenDoc(id)?.import(...)`.
- **`PatchesDoc.flush()`** — resolves when the per-doc change queue has settled and `_optimisticOps` is drained (confirmed or rolled back). Replaces the private-field-poking `waitForOpenDocQuiescent` pattern that dabble-writer-3.0 carries today. Standalone (un-wired) docs return immediately; wired-but-cleared queues (post-closeDoc with in-flight work) yield to the macrotask queue so IndexedDB callbacks can drain.

Both methods unblock DAB-426 (dabble-writer-3.0's pup-client integration) and let ~80 LOC of app-level race handling get deleted there.

## Review iterations

Went through three rounds of `/code-review`. Notable surfaced-and-fixed issues:
- Equal-rev `import()` would have wiped `OTDoc._pendingChanges` / `LWWDoc._inFlightOpKeys` — strict `>` guard chosen over `>=`.
- `_openingSnapshots` cleared in `finally` on every exit (not preserved on failure) to avoid a dead window where post-failure snapshots got dropped and a memory leak for abandoned opens.
- `flush()` uses `setTimeout(0)` (not `Promise.resolve()`) so the spin path yields to macrotasks instead of starving IndexedDB callbacks.

Four pre-existing patches bugs surfaced during the review are tracked in [DAB-507](https://linear.app/dabblewriter/issue/DAB-507) — not in scope for this PR.

## Test plan

- [ ] `npm run type:check`, `npm run lint`, `npm run format:check`, `npm run test` all pass (currently 1900 tests green)
- [ ] DAB-426 consumer can delete `pendingSnapshots`, `consumePendingSnapshot`, `waitForOpenDocQuiescent` and route through the new APIs without behavioral regression
- [ ] Manually exercise: open a doc in two tabs; broadcast snapshot from one mid-`openDoc` in the other; verify the spoke tab opens at the broadcast rev instead of stale store state

Closes DAB-503.

### DIFF
--- a/docs/Patches.md
+++ b/docs/Patches.md
@@ -164,6 +164,27 @@ await patches.deleteDoc('old-shopping-list');
 
 Closing docs frees memory and ensures pending changes are persisted.
 
+### Applying External Snapshots
+
+`PatchesSync` handles snapshots that flow over its own connection. Any other transport that delivers snapshots out-of-band (multi-tab broadcast hubs, WebRTC peer mesh, custom server-push channels) has a race: a snapshot can land mid-`openDoc`, before the doc is in `getOpenDoc()`. Looking the doc up and calling `doc.import()` silently drops the snapshot in that window.
+
+`applySnapshot` makes the window disappear:
+
+```typescript
+// Hub tab broadcasts a fresh snapshot; spoke tab applies it.
+hub.onSnapshot((docId, snapshot) => {
+  patches.applySnapshot(docId, snapshot);
+});
+```
+
+What it does, based on the doc's state:
+
+- **Already open** — imports immediately if `snapshot.rev > doc.committedRev`. Equal-rev and older snapshots are dropped (a `doc.import` at equal rev would reset internal pending-state that the user may still be filling).
+- **`openDoc` in flight** — stashes in a single slot, keeping the highest-rev snapshot seen so out-of-order delivery doesn't lose newer state. `openDoc` drains the slot before resolving.
+- **Neither open nor opening** — dropped. Patches doesn't hold onto snapshots for docs it isn't managing.
+
+Idempotent, no-throw, and safe to call from any transport callback. Use it instead of the `getOpenDoc(id)?.import(snapshot)` pattern.
+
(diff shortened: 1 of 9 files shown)


RyanHavoc (commented):
Read through and it's clean — the race analysis in the PR body matches what the code does, and the trade-offs are documented in JSDoc where it matters.

A couple of small notes, no blockers:

- **`_setFlushAwaiter` getter pattern across reopen.** The getter resolves `_changeQueues.get(docId)` on each iteration, so a `flush()` awaited across a `closeDoc` → `openDoc` cycle on the same docId could end up resolving against the new queue tail rather than the original work. JSDoc on `PatchesDoc.flush` already warns about this, so it's a known sharp edge — fine.

- **Strict `>` rev guard.** Right call given that `doc.import()` resets `_pendingChanges` / `_inFlightOpKeys` from `snapshot.changes`. The equal-rev case (duplicate broadcast with stripped changes) would silently wipe in-flight ops. Worth keeping the comment that explains this — it's load-bearing for anyone tempted to "relax" it later.

- **`setTimeout(0)` macrotask yield.** The 4ms HTML-spec clamp note is accurate and the choice is right — `Promise.resolve()` would starve IndexedDB callbacks. Good to have it spelled out.

RyanHavoc (approved):

PR #49, opened 2026-05-26
fix: DAB-507: four pre-existing Patches bugs from DAB-503 review

jacwright:
## Summary

Fixes [DAB-507](https://linear.app/dabblewriter/issue/DAB-507/dabblepatches-4-pre-existing-bugs-surfaced-during-dab-503-review) — four pre-existing bugs in `@dabble/patches` surfaced during the DAB-503 review. All four predate DAB-503; its diff did not introduce or worsen them. Patch bump to **0.9.10**.

### Bug 1: snapshot delivery wipes in-memory pending state
`PatchesSync` fresh-fetch (`PatchesSync.ts:488`) and `docReloadRequired` (`:555`) called `patches.applySnapshot(docId, { ...snapshot, changes: [] })`. For an open doc with `snapshot.rev > committedRev`, `applySnapshot` invoked `doc.import(snapshot)`, and:
- `OTDoc.import` set `_pendingChanges = []` — wiping the in-memory view of pending while the algorithm's store still had them. Visible regression of pending characters mid-typing.
- `LWWDoc.import` called `_inFlightOpKeys.clear()` — breaking pure-echo detection so the next server echo of the user's own ops triggered a spurious `_recomputeState()` with a fresh state identity (visible cursor jump for editor subscribers).

**Fix:** re-read the snapshot via `algorithm.loadDoc(docId)` (which derives the snapshot from the store, including pending) and pass that to `applySnapshot`.

### Bug 2: `openDoc` failure orphans `trackDocs` side-effects
`openDoc` did `await this.trackDocs([docId])` before `loadDoc`. If anything between trackDocs and `this.docs.set` threw, the doc was left in `trackedDocs`, persisted by the algorithm, and subscribed via `onTrackDocs` — but never in `this.docs`. "Tracked but never opened" zombie until retry.

**Fix:** snapshot `wasTracked = this.trackedDocs.has(docId)` before trackDocs; in `catch`, if we added it, unwind via `untrackDocs`. Best-effort — an untrack failure doesn't mask the original openDoc error.

### Bug 3: `Patches.close()` doesn't await in-flight `openDoc`
`close()` cleared maps synchronously, then awaited algorithm closures. An `openDoc` resuming after its last `await` could call `this.docs.set(docId, ...)` against a now-emptied map with a doc bound to closed algorithm internals.

**Fix:** added `_openingPromises` map populated by `openDoc` via a deferred (cleared in `finally`). `close()` awaits `Promise.allSettled([..._openingPromises.values()])` before clearing state.

### Bug 4: `deleteDoc` / `untrackDocs` tombstone the wrong algorithm store
For a closed doc opened on a non-default algorithm, both `deleteDoc` (`Patches.ts:316`) and `untrackDocs` (`:164`) fell back to `defaultAlgorithm` since `this.docs.get(docId)` was `undefined`. Result: tombstone or untrack written in the wrong store while the real data sat untouched in the algorithm that actually owned it.

**Fix:** extracted `_resolveAlgorithmForDoc` that scans each algorithm's `listDocs(true)` when the doc isn't open (mirroring the lookup in `openDoc`). Used in both `deleteDoc` and `untrackDocs`.

### Out of scope (worth filing separately)
While verifying Bug 1, found that `LWWInMemoryStore.saveDoc` and `LWWIndexedDBStore.saveDoc` both wipe `pendingOps` and `sendingChange` despite their docstrings claiming otherwise. PatchesSync's fresh-fetch and `docReloadRequired` paths therefore cause data loss for LWW docs with concurrent local writes — pre-existing, distinct from this issue's four bugs.

## Test plan
- [x] `npm run type:check`
- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm run test` (1900 tests pass)
- [ ] Manual: open OT doc, type a few chars, force `docReloadRequired` — verify pending characters survive in `OTDoc._pendingChanges`
- [ ] Manual: open LWW doc, type while server roundtrip in flight — verify no cursor jump on next echo
- [ ] Manual: `algorithms: { ot, lww }`, `defaultAlgorithm: 'ot'` — open `'x'` with `algorithm: 'lww'`, change, closeDoc, deleteDoc — verify LWW store gets the tombstone, OT store has no record

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@dabble/patches",
-  "version": "0.9.9",
+  "version": "0.9.10",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@dabble/patches",
-      "version": "0.9.9",
+      "version": "0.9.10",
       "license": "MIT",
       "dependencies": {
         "@dabble/delta": "^1.2.5",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dabble/patches",
-  "version": "0.9.9",
+  "version": "0.9.10",
   "description": "Immutable JSON Patch implementation based on RFC 6902 supporting operational transformation and last-writer-wins",
   "author": "Jacob Wright <jacwright@gmail.com>",
   "bugs": {
--- a/src/client/LWWInMemoryStore.ts
+++ b/src/client/LWWInMemoryStore.ts
@@ -92,14 +92,18 @@ export class LWWInMemoryStore implements LWWClientStore {
 
   /**
    * Saves the current document state to storage.
-   * Clears committed fields (subsumed by the snapshot) but preserves pending ops.
+   * Clears committed fields (subsumed by the snapshot) but preserves pending ops
+   * and the in-flight sending change — those represent local edits not yet
+   * accepted by the server and would otherwise be silently dropped when
+   * PatchesSync re-saves a freshly-fetched snapshot.
    */
   async saveDoc(docId: string, docState: PatchesState): Promise<void> {
+    const existing = this.docs.get(docId);
     this.docs.set(docId, {
       snapshot: { state: docState.state, rev: docState.rev },
       committedFields: new Map(),
-      pendingOps: new Map(),
-      sendingChange: null,
+      pendingOps: existing?.pendingOps ?? new Map(),
+      sendingChange: existing?.sendingChange ?? null,
       committedRev: docState.rev,
     });
   }
--- a/src/client/LWWIndexedDBStore.ts
+++ b/src/client/LWWIndexedDBStore.ts
@@ -204,12 +204,15 @@ export class LWWIndexedDBStore implements LWWClientStore {
 
   /**
    * Saves the current document state to storage.
-   * Clears committed fields (subsumed by the snapshot) but preserves pending ops.
+   * Clears committed fields (subsumed by the snapshot) but preserves pending ops
+   * and the in-flight sending change — those represent local edits not yet
+   * accepted by the server and would otherwise be silently dropped when
+   * PatchesSync re-saves a freshly-fetched snapshot.
    */
   @blockable
   async saveDoc(docId: string, docState: PatchesState): Promise<void> {
-    const [tx, snapshots, committedOps, pendingOps, sendingChanges, docsStore] = await this.db.transaction(
-      ['snapshots', 'committedOps', 'pendingOps', 'sendingChanges', 'docs'],
+    const [tx, snapshots, committedOps, docsStore] = await this.db.transaction(
+      ['snapshots', 'committedOps', 'docs'],
       'readwrite'
     );
 
@@ -219,8 +222,6 @@ export class LWWIndexedDBStore implements LWWClientStore {
       docsStore.put<TrackedDoc>({ docId, committedRev: rev, algorithm: 'lww' }),
       snapshots.put<Snapshot>({ docId, state, rev }),
       this.deleteFieldsForDoc(committedOps, docId),
-      this.deleteFieldsForDoc(pendingOps, docId),
-      sendingChanges.delete(docId),
     ]);
 
     await tx.complete();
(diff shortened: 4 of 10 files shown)


matalina (approved):
Surgical, well-documented PR. Each fix has a clear root-cause comment in the code, and the design choices (deferred for `close()` to await opens, `_resolveAlgorithmForDoc` mirroring openDoc's lookup, `wasTracked` snapshot before trackDocs) are exactly what these bugs needed. Approving — a couple of items worth addressing in follow-up.

**Correctness**

- Bug 1 fix is incomplete for LWW docs. Re-reading via `algorithm.loadDoc(docId)` is the right move, but the "out of scope" note flags that `LWWInMemoryStore.saveDoc` and `LWWIndexedDBStore.saveDoc` wipe `pendingOps` / `sendingChange` before `loadDoc` runs. So for LWW docs `loadDoc` returns a snapshot with no pending, `applySnapshot` calls `doc.import()` with empty `changes`, and `LWWDoc._inFlightOpKeys` still gets cleared — the cursor-jump regression Bug 1 was meant to fix isn't actually fixed for LWW users. Recommend filing the LWW store fix as the next follow-up and either noting in the changelog that LWW is still affected, or holding the 0.9.10 release until both ship. Otherwise the commit message will misleadingly claim Bug 1 is resolved.
- `@singleInvocation(true)` on openDoc correctly dedupes concurrent calls for the same docId, so the `_openingPromises.set(docId, …)` overwrite concern is moot. Worth a one-liner ("safe under @singleInvocation: only one body per docId") to head off the question for the next reviewer.
- `_resolveAlgorithmForDoc` is O(algorithms × docsPerAlgorithm) and `untrackDocs` calls it inside a for-loop over docIds — for a batch untrack of N docs across A algorithms that's N × A `listDocs(true)` round-trips. One scan up-front producing a `Map<docId, algorithm>` would be A round-trips total. Fine at current call-site sizes (1–10 docs typical), but worth a comment if you keep the per-doc version.
- Best-effort untrack in the openDoc catch path swallows untrackDocs errors silently. If untrack itself fails, the doc remains in `trackedDocs` *and* potentially in the algorithm's store — the very "tracked but never opened" zombie state the fix is meant to prevent. At minimum emit via `onError` so callers and tests can observe.
- Multi-algorithm same-docId resolution: `_resolveAlgorithmForDoc` returns the first algorithm whose `listDocs(true)` claims the docId, with iteration order = `Object.values(this.algorithms)` insertion order. Well-defined under the invariant that a docId lives in exactly one algorithm, but if that invariant ever breaks (partial migration leaving stale records in both stores) resolution becomes silently non-deterministic. Worth either asserting the invariant or documenting the precedence.

**Tests**

Light for a four-bug fix — only Bug 1 got a test update. Nothing covers:

- *Bug 2*: openDoc failure-path unwind via untrackDocs. Mock `algorithm.loadDoc` or `createDoc` to throw after `trackDocs`, assert `trackedDocs.has(docId) === false`. Also a pre-tracked case to verify we don't untrack a doc tracked before this open.
- *Bug 3*: `close()` awaiting in-flight openDoc. Race-y but doable with a controlled algorithm mock holding a deferred — call `close()`, resolve the deferred, assert ordering + no `docs.set` against cleared map.
- *Bug 4*: deleteDoc / untrackDocs for a closed doc on a non-default algorithm. Open with `algorithm: 'lww'`, closeDoc, deleteDoc, assert the LWW store got the tombstone and OT did not.

These are exactly the kind of subtle bugs that regress silently without tests. The manual test plan covers them but manual-only is fragile.

**Cleanup**

- `// eslint-disable-next-line @typescript-eslint/no-empty-function` + `let resolveOpening: () => void = () => {};` is awkward. Cleaner with definite-assignment:
  ```ts
  let resolveOpening!: () => void;
  this._openingPromises.set(docId, new Promise<void>(r => { resolveOpening = r; }));
  ```
- The `if (fullSnapshot)` guard after `algorithm.loadDoc` is defensive but the doc was just saved one line up — loadDoc returning null there would be a store bug. Fine to keep, just noting the guard hides what would be a real defect.
- `_resolveAlgorithmForDoc`'s comment says "mirroring openDoc at lines 206-213" — line numbers rot fast; better as "mirroring openDoc's algorithm-detection loop."

PR #50, opened 2026-05-28
client: refcount openDoc/closeDoc

RyanHavoc:
## Summary

Adds reference counting to `ManagedDoc` so multiple independent consumers can each `openDoc` / `closeDoc` the same path without one teardown ripping the doc out from under another.

- `openDoc` increments `refCount` on every call (including when an existing doc is returned).
- `closeDoc` decrements and only tears the doc down when the count hits zero.
- `deleteDoc` passes `{ force: true }` to evict regardless of outstanding refs — when a doc is being deleted, every consumer must release.

**Caller contract:** pair every `openDoc` with exactly one `closeDoc`.

### Why

Pup-client's Vue composables (`usePupDoc`, `usePupDocs`) and any future consumer that opens the same content path concurrently were racing: the first component to unmount would call `closeDoc` and tear down a doc the other was still rendering. Refcounting at the `Patches` layer is the right place to fix this — composables shouldn't need their own bookkeeping for shared underlying docs.

## Test plan

- [ ] CI green (lint, tests, build)
- [ ] Verify in pup-client branches-refactor: two components binding the same content path mount/unmount in any order without "doc closed" errors
- [ ] `deleteDoc` still evicts even with outstanding refs

### DIFF
--- a/src/client/Patches.ts
+++ b/src/client/Patches.ts
@@ -32,11 +32,15 @@ export interface PatchesOptions {
   docOptions?: PatchesDocOptions;
 }
 
-// Internal doc management structure
+// Internal doc management structure. `refCount` tracks how many callers
+// currently hold the doc open via `openDoc`; `closeDoc` only tears the doc
+// down when the count hits zero, so independent consumers can each manage
+// their own open/close lifecycle without stomping on one another.
 interface ManagedDoc<T extends object> {
   doc: PatchesDoc<T>;
   algorithm: ClientAlgorithm;
   unsubscribe: Unsubscriber;
+  refCount: number;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/client/Patches.ts
```diff
@@ -360,19 +384,29 @@ export class Patches {
...
-      managed.unsubscribe();
-      this.docs.delete(docId);
-      this._changeQueues.delete(docId);
-      if (untrack) {
-        await this.untrackDocs([docId]);
-      }
+    if (!managed) return;
+    if (!force) {
+      if (managed.refCount > 0) managed.refCount -= 1;
+      if (managed.refCount > 0) return;
```
jacwright, line 403: Non-blocking: the new refcount semantics have no direct test coverage. The existing "return existing document if already open" test now silently leaves `refCount` at 2, and nothing asserts the core contract. Worth adding a few cases:

- two `openDoc` + one `closeDoc` → doc stays open (`unsubscribe` not called, `getOpenDoc` still defined)
- two `openDoc` + two `closeDoc` → torn down once
- `deleteDoc` evicts with outstanding refs (the `force: true` path)
- a stray extra `closeDoc` past zero is a harmless no-op

This is the subtle part of the change, so locking it down with tests would be valuable.

jacwright (approved):
Looks correct and the doc comments are excellent. I traced the concurrency paths and they hold up:

- Concurrent first-time opens share one `_openDocOnce` promise (singleInvocation), then each caller increments `refCount` independently, so N callers get N refs.
- `closeDoc` only tears down at zero; `deleteDoc` force-evicts regardless.
- I checked the one race I was worried about (a `close()`/`closeDoc` landing between `_openDocOnce` resolving and the post-await `this.docs.get(docId)!` increment, which would blow the `!`). It doesn't manifest: `close()` awaits `Promise.allSettled([...])`, whose extra microtask hop lets `openDoc`'s single-await continuation increment first. So the non-null assertion is safe in practice.

All 45 tests in `Patches.spec.ts` pass on the branch and CI is green. Approving. One non-blocking note inline about test coverage.

src/client/Patches.ts
RyanHavoc: Added coverage.

PR #52, opened 2026-06-08
Support reading an OT document at a revision via getDoc(docId, rev)

jacwright:
## What

Lets the server read an OT document **as of an earlier revision** through the existing `getDoc`:

```ts
const stream = await server.getDoc(docId, 123);
// {"state":...,"rev":M,"changes":[...]} — version at/before 123 + changes up to rev 123
```

Without `rev`, behavior is unchanged (latest state + all changes since the last version).

## Why

`getSnapshotAtRevision` already encodes "state at a revision" logic, but it isn't reachable from any public entry point, and `getDoc`/`getSnapshotStream` ignored the concept. Consumers (Pup, app clients) that want point-in-time reads had no API for it short of pulling all `_changes` and replaying client-side. This threads the existing capability through the method everyone already calls.

## How

- `getSnapshotStream(store, docId, rev?)` — selects the latest `main` version at or before `rev` and bounds `listChanges` with `endBefore: rev+1`. Same streaming envelope as before.
- `OTServer.getDoc(docId, rev?)` and the `PatchesServer` interface gain the optional `rev`.
- `LWWServer.getDoc(docId, rev?)` throws if `rev` is passed — LWW stores fields, not per-revision history.

## Tests

- `getSnapshotStream`: latest envelope, bounded-to-rev, and null-state-when-no-version.
- `OTServer.getDoc` forwards `rev` to `getSnapshotStream`.
- `LWWServer.getDoc` rejects `rev`.
- Full suite green (1922 passed); type/lint/format clean; build clean.

## Follow-up

Unblocks a Pup change that exposes `GET /docs/:docId?rev=` by passing the param straight through to `getDoc` (pending a release bump of `@dabble/patches`).

### DIFF
--- a/src/algorithms/ot/server/getSnapshotAtRevision.ts
+++ b/src/algorithms/ot/server/getSnapshotAtRevision.ts
@@ -9,7 +9,7 @@ async function getLatestMainVersion(store: OTStoreBackend, docId: string, before
   const versions = await store.listVersions(docId, {
     limit: 1,
     reverse: true,
-    startAfter: beforeRev ? beforeRev + 1 : undefined,
+    startAfter: beforeRev != null ? beforeRev + 1 : undefined,
     origin: 'main',
     orderBy: 'endRev',
   });
@@ -42,7 +42,7 @@ export async function getSnapshotAtRevision(
   // Get *all* changes since that version up to the target revision (if specified)
   const changesSinceVersion = await store.listChanges(docId, {
     startAfter: versionRev,
-    endBefore: rev ? rev + 1 : undefined,
+    endBefore: rev != null ? rev + 1 : undefined,
   });
 
   return {
@@ -70,7 +70,10 @@ export async function getSnapshotStream(
 ): Promise<ReadableStream<string>> {
   const { rawState, versionRev } = await getLatestMainVersion(store, docId, rev);
   const statePayload: string | ReadableStream<string> = rawState ?? 'null';
-  const changes = await store.listChanges(docId, { startAfter: versionRev, endBefore: rev ? rev + 1 : undefined });
+  const changes = await store.listChanges(docId, {
+    startAfter: versionRev,
+    endBefore: rev != null ? rev + 1 : undefined,
+  });
 
   return concatStreams(`{"state":`, statePayload, `,"rev":${versionRev},"changes":`, JSON.stringify(changes), '}');
 }
--- a/src/algorithms/ot/server/getSnapshotAtRevision.ts
+++ b/src/algorithms/ot/server/getSnapshotAtRevision.ts
@@ -58,11 +58,19 @@ export async function getSnapshotAtRevision(
  *
  * The version state (typically the largest payload) flows through as a raw
  * string from the store without parsing. Changes are stringified from the array.
+ *
+ * When `rev` is given, the snapshot reflects the document as of that revision:
+ * the latest version at or before `rev`, plus the changes from there up to `rev`.
+ * Omitted, it streams the current state with all changes since the latest version.
  */
-export async function getSnapshotStream(store: OTStoreBackend, docId: string): Promise<ReadableStream<string>> {
-  const { rawState, versionRev } = await getLatestMainVersion(store, docId);
+export async function getSnapshotStream(
+  store: OTStoreBackend,
+  docId: string,
+  rev?: number
+): Promise<ReadableStream<string>> {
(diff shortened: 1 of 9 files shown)


src/algorithms/ot/server/getSnapshotAtRevision.ts
```diff
@@ -58,11 +58,19 @@ export async function getSnapshotAtRevision(
...
-  const { rawState, versionRev } = await getLatestMainVersion(store, docId);
+export async function getSnapshotStream(
+  store: OTStoreBackend,
+  docId: string,
+  rev?: number
+): Promise<ReadableStream<string>> {
+  const { rawState, versionRev } = await getLatestMainVersion(store, docId, rev);
   const statePayload: string | ReadableStream<string> = rawState ?? 'null';
-  const changes = await store.listChanges(docId, { startAfter: versionRev });
+  const changes = await store.listChanges(docId, { startAfter: versionRev, endBefore: rev ? rev + 1 : undefined });
```
RyanHavoc: **Bug: `rev=0` silently returns the latest document instead of the empty/initial state.**

The truthiness guard `rev ? rev + 1 : undefined` treats `0` (falsy) the same as "no rev passed": it selects the *latest* main version and returns *all* changes. So `getDoc(docId, { rev: 0 })` returns the full current doc rather than the state as of rev 0 — silent wrong data, not an error.

This is reachable: pup#88's validator accepts `rev=0` (it only rejects `n < 0`), so `GET /docs/:docId?rev=0` is an accepted request that returns incorrect results.

Same issue applies to `getLatestMainVersion`'s `startAfter: beforeRev ? beforeRev + 1 : undefined` (line 12) and the parsing variant `getSnapshotAtRevision` (line 45) — both are pre-existing latent bugs that this PR now exposes through a public, user-reachable path.

Suggest a nullish check throughout, matching how `LWWServer` already guards:
```ts
endBefore: rev != null ? rev + 1 : undefined
startAfter: beforeRev != null ? beforeRev + 1 : undefined
```
Worth a `rev=0` test too — it would have caught this.

src/server/LWWServer.ts
```diff
@@ -84,9 +84,11 @@ export class LWWServer implements PatchesServer {
    * flowing through without parsing.
    *
    * @param docId - The document ID.
+   * @param options - `rev` is unsupported for LWW (no per-revision history); passing it throws.
    * @returns A ReadableStream of JSON string chunks.
    */
-  async getDoc(docId: string): Promise<ReadableStream<string>> {
+  async getDoc(docId: string, options?: GetDocOptions): Promise<ReadableStream<string>> {
+    if (options?.rev != null) throw new Error('LWW documents do not support reading at a specific revision');
```
RyanHavoc, line 91: Note the inconsistency this creates with OT: LWW uses `!= null` here, so it correctly rejects `rev: 0` as "a rev was supplied." But the OT path (`getSnapshotStream`) uses a truthiness check and treats `rev: 0` as "not supplied." Same input, two interpretations. Switching the OT guards to `!= null` (see comment on `getSnapshotAtRevision.ts`) resolves both that bug and this inconsistency in one move. This line is correct as written.

RyanHavoc (commented):
Reviewed alongside pup#88. Design is right — threading the existing `getLatestMainVersion`/`getSnapshotAtRevision` capability through the public `getDoc` is the clean move, and tests/docs/release ordering are all handled. One real correctness bug around `rev=0`, plus a cross-implementation inconsistency. Details inline.

src/algorithms/ot/server/getSnapshotAtRevision.ts
jacwright: Fixed in `81c6852`: switched all three guards (`getLatestMainVersion` startAfter, `getSnapshotAtRevision` endBefore, `getSnapshotStream` endBefore) to `!= null`, so `rev=0` now bounds to the empty pre-history state instead of returning latest. Added `rev=0` tests on both `getSnapshotAtRevision` and `getSnapshotStream` (`a4017d6`). The LWW `!= null` guard was already correct, so OT and LWW now agree on `rev=0` semantics.

src/server/LWWServer.ts
jacwright: Agreed, left this line as-is. Switching the OT guards to `!= null` in `81c6852` brings the two paths into agreement: OT `rev=0` now bounds to the empty pre-history state and LWW `rev=0` still rejects (it has no per-rev history), so `rev=0` is consistently treated as "a rev was supplied" on both sides.

PR #54, opened 2026-06-13
fix(ot): version/rev integrity — orphan versions, snapshot rev cap, merge-base clamp

RyanHavoc:
## Background

On a duplicated, size-pushing project a client got stuck re-committing the full project state as a root-replace. Each attempt rebased to a no-op server-side (the content was already committed) but still **minted an `offline-branch` version stamped at a rev that never persisted** — 334 orphan versions at `endRev=295` on a doc whose change log topped out at **294**. `getDoc` then reported the head as **295** (it trusted the highest-endRev version), so branches were created at `branchedAtRev=295` and every merge threw `baseRev (295) > currentRev (294)`. And the client never cleared the no-op'd pending change, so it resent forever — fuelling the storm.

This PR fixes the root causes (create-side, read-side, client-side) plus the merge-time symptom.

## Changes

**1. Don't mint versions for changes that never persist** (`commitChanges`)
The offline/batch path versioned the *incoming* (pre-transform) changes. When an offline change rebased to a no-op it wasn't saved, yet a version was still created at its claimed rev. Now the offline session is versioned from the changes that actually persist (their post-transform revs), and **no version is created when nothing persists**.

**2. A snapshot can never sit ahead of the change log** (`getSnapshotAtRevision` / `getLatestMainVersion`)
Version selection had no upper bound, so an orphan version (`endRev > currentRev`) was picked as the snapshot base — making `getDoc` report a phantom rev and drop the real tail of changes. It now bounds the search by `getCurrentRev` (and the caller's target rev). Also corrects a latent bug where a target rev selected a version *after* it rather than at/before it.

**3. Drop pending changes the server rebased away** (`PatchesSync` / `OTAlgorithm` / `OTClientStore`)
A sent change that rebases to a no-op is never echoed back, so the rebase-by-id removal can't clear it and a root-replace never reduces to empty — the client resent it on every flush forever. After a commit, the client now drops pending changes that were sent but absent from the server's response (new `OTClientStore.dropPendingChanges` + optional `ClientAlgorithm.dropResolvedPending`), and re-syncs the open doc from the store when it does. This is what actually stops the retry storm.

**4. Clamp the merge base to the source tip** (`OTBranchManager.mergeBranch`)
Defense for branch records already written with an inflated `branchedAtRev`: clamp `branchStartRevOnSource = min(branchedAtRev, getCurrentRev(sourceDocId))`. No-op for healthy branches.

## Tests
- `commitChanges`: a no-op offline change creates no version and saves nothing; a persisting offline change is versioned from its rebased rev, not the claimed rev.
- `getSnapshotAtRevision`: never selects a version ahead of the committed tip.
- `dropResolvedPending`: drops a rebased-away sent change, keeps an echoed-back one, leaves unrelated pending intact; `PatchesSync` flush wiring drops + re-syncs the open doc.
- `OTBranchManager`: merge clamps the base when `branchedAtRev` is ahead of the tip.
- Full suite green (1933 tests); type-check, lint, format clean.

## Follow-ups (not in this PR)
- **pup `FirestoreOTStore.listVersions`** ignores the `origin` filter (defense-in-depth; needs an `(origin, endRev)` composite index). The `getCurrentRev` cap (#2) already makes `getDoc` correct without it.
- Data cleanup: purge the ~334 orphan `offline-branch@295` versions on the affected prod doc (direct Firestore delete; no `deleteVersion` API).

### DIFF
--- a/src/algorithms/ot/server/commitChanges.ts
+++ b/src/algorithms/ot/server/commitChanges.ts
@@ -151,12 +151,16 @@ export async function commitChanges(
 
       // Fast-forward: nothing committed after baseRev, so the incoming changes save
       // verbatim. Their revs are final, so version them directly (origin 'main').
+      // Save before versioning: if `saveChanges` throws a RevConflictError (a
+      // concurrent commit landed first), nothing was versioned and the retry falls
+      // through to the transform branch cleanly — no version minted from the
+      // pre-transform changes is left stranded ahead of the real log.
       if (isOfflineOrBatch && committedChanges.length === 0) {
+        await store.saveChanges(docId, incomingChanges);
         if (!offlineSessionsHandled) {
           await handleOfflineSessionsAndBatches(store, sessionTimeoutMillis, docId, incomingChanges, 'main');
           offlineSessionsHandled = true;
         }
-        await store.saveChanges(docId, incomingChanges);
         return { catchupChanges: [], newChanges: incomingChanges, docReloadRequired };
       }
 
@@ -169,6 +173,9 @@ export async function commitChanges(
       );
 
       if (transformedChanges.length > 0) {
+        // Save before versioning (same ordering as the fast-forward branch) so a
+        // RevConflictError on save never leaves a version behind to retry against.
+        await store.saveChanges(docId, transformedChanges);
         // Version the offline/batch session from the changes that ACTUALLY persisted
         // (their post-transform revs), never the pre-transform claimed revs. When an
         // offline change rebases to a no-op it isn't saved — versioning the claimed
@@ -179,7 +186,6 @@ export async function commitChanges(
           await handleOfflineSessionsAndBatches(store, sessionTimeoutMillis, docId, transformedChanges, origin);
           offlineSessionsHandled = true;
         }
-        await store.saveChanges(docId, transformedChanges);
       }
 
       // Return catchup changes and newly transformed changes separately
--- a/src/algorithms/ot/server/commitChanges.ts
+++ b/src/algorithms/ot/server/commitChanges.ts
@@ -143,27 +143,21 @@ export async function commitChanges(
         return { catchupChanges: committedChanges, newChanges: [], docReloadRequired };
       }
 
-      // 4. Handle offline-session versioning (once — version creation is not idempotent):
+      // 4. Offline-session versioning applies when:
       // - batchId present (multi-batch uploads)
       // - or the first change is older than the session timeout (single-batch offline)
       const isOfflineTimestamp = serverNow - incomingChanges[0].createdAt > sessionTimeoutMillis;
-      if (isOfflineTimestamp || batchId) {
-        const canFastForward = committedChanges.length === 0;
+      const isOfflineOrBatch = isOfflineTimestamp || !!batchId;
 
+      // Fast-forward: nothing committed after baseRev, so the incoming changes save
+      // verbatim. Their revs are final, so version them directly (origin 'main').
+      if (isOfflineOrBatch && committedChanges.length === 0) {
         if (!offlineSessionsHandled) {
-          // In historicalImport mode, always use 'main' origin
-          const origin = options?.historicalImport ? 'main' : canFastForward ? 'main' : 'offline-branch';
-
-          // Create versions for offline sessions (metadata + changes, no state)
-          await handleOfflineSessionsAndBatches(store, sessionTimeoutMillis, docId, incomingChanges, origin);
+          await handleOfflineSessionsAndBatches(store, sessionTimeoutMillis, docId, incomingChanges, 'main');
           offlineSessionsHandled = true;
         }
-
-        // Fast-forward: no transformation needed, save changes directly
-        if (canFastForward) {
-          await store.saveChanges(docId, incomingChanges);
-          return { catchupChanges: [], newChanges: incomingChanges, docReloadRequired };
-        }
+        await store.saveChanges(docId, incomingChanges);
+        return { catchupChanges: [], newChanges: incomingChanges, docReloadRequired };
       }
 
       // 5. Transform the incoming changes against committed changes (stateless — no state loaded)
@@ -175,6 +169,16 @@ export async function commitChanges(
       );
 
       if (transformedChanges.length > 0) {
+        // Version the offline/batch session from the changes that ACTUALLY persisted
+        // (their post-transform revs), never the pre-transform claimed revs. When an
+        // offline change rebases to a no-op it isn't saved — versioning the claimed
+        // rev here would mint an orphan version pointing past the committed log, which
+        // poisons getDoc's reported rev and never gets cleaned up.
+        if (isOfflineOrBatch && !offlineSessionsHandled) {
+          const origin = options?.historicalImport ? 'main' : 'offline-branch';
+          await handleOfflineSessionsAndBatches(store, sessionTimeoutMillis, docId, transformedChanges, origin);
+          offlineSessionsHandled = true;
+        }
         await store.saveChanges(docId, transformedChanges);
       }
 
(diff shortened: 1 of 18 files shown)


jacwright (changes requested):
Really strong PR — the root-cause story (no-op offline commit mints an `offline-branch` version past the change log → `getDoc` trusts the highest-`endRev` version → inflated `committedRev`/`branchedAtRev` → merge throws) is exactly right, and the new client retry-storm fix (`dropResolvedPending`) goes after the actual source of those no-op commits. Fixes #1 (version only what persists) and #3 (merge-base clamp) are sound and well-tested.

One blocking issue before merge, plus a couple of non-blocking notes.

## 🔴 Blocking — fix #2 inverts the version query against pup (production)

`getLatestMainVersion` now calls `listVersions({ reverse: true, endBefore: upperBound + 1, orderBy: 'endRev' })`, intending an absolute `endRev <= upperBound`. The patches in-memory reference store (`LWWMemoryStoreBackend.listVersions`) treats `endBefore` as absolute, so the tests pass. But the **production store treats `startAfter`/`endBefore` as cursors that flip under `reverse`**, and pup's own tests enshrine it:

- `pup/tests/store/FirestoreOTStore.spec.ts:112` — *"treats startAfter as an upper bound when reversed"*: `{ startAfter: 1001, reverse: true }` → `where('endRev', '<', 1001)`.
- `pup/tests/store/FirestoreOTStore.spec.ts:137` — *"flips endBefore to a lower bound when reversed"*: `{ endBefore: 20, reverse: true }` → `where('endRev', '>', 20)`.

So against `FirestoreOTStore`, `endBefore: upperBound + 1` becomes `where('endRev', '>', upperBound + 1)` — selecting versions **above** the bound. For a healthy doc (`currentRev` 250, versions ≤ 250) that matches **nothing** → `versionRev` 0 → every `getDoc` drops its snapshot base and **replays the entire change log**. On the size-pushing projects this bug is about, that's a severe perf/cost regression. The PR's unit tests can't catch it because they mock the store with the opposite (absolute) semantics.

Note pup's comment at `:109`: the regression test was written for the *old* `getLatestMainVersion` call — `{ startAfter: rev + 1, reverse: true }` — precisely because that yields the `endRev <= rev` bound we want. The old `startAfter` form was correct against prod; this PR swapped it for `endBefore`, which is the inverted form.

### Fix (small, and no pup change required)
Keep using `startAfter`, just add the `currentRev` cap:
```ts
startAfter: upperBound + 1,   // not endBefore
```
Against pup this is `where('endRev', '<', upperBound + 1)` = `endRev <= upperBound` — excludes the orphan@295, keeps the latest legit version, and reuses the existing `(origin, endRev)` composite index. Then:
1. Fix `LWWMemoryStoreBackend.listVersions` to apply the same reverse-cursor flip, so the reference store matches the real one (right now it's the outlier).
2. Update the `getSnapshotAtRevision` tests to expect `startAfter`, not `endBefore`.
3. Clarify the `ListVersionsOptions` doc — `startAfter`/`endBefore` are cursors relative to sort order; the current "strictly before this value" line plus the contradictory `reverse` note are what let these two implementations drift apart.
4. Add one conformance test both `LWWMemoryStoreBackend` and `FirestoreOTStore` run against the same fixtures — the whole bug is that two implementations of one interface disagree and only the mock is exercised.

(If we'd rather standardize on *absolute* semantics instead, that's a pup change — drop the `desc ? … ` flips in `FirestoreOTStore.listVersions` and rewrite those three pup tests — but it touches the hot getDoc query and is higher risk than conforming patches to pup. I'd go with the patches-side fix above.)

## 🟡 Non-blocking

**`OTBranchManager` clamp — log when it fires.** A triggered clamp means a corrupted/renumbered source doc, i.e. exactly the data-integrity case worth surfacing. A `console.warn` when `branch.branchedAtRev > sourceCurrentRev` keeps these from merging silently.

**`commitChanges` fast-forward + conflict-retry residual.** In the FF branch you version from `incomingChanges` and set `offlineSessionsHandled = true` before `saveChanges`. If `saveChanges` throws `RevConflictError`, the retry falls into the transform branch but `offlineSessionsHandled` is already set, so the session is never re-versioned from the actually-persisted `transformedChanges`; the version minted from pre-transform changes survives. Narrow (concurrent commit during an offline/batch flush) and strictly better than before, so fine to defer — worth a follow-up.

**`dropResolvedPending` looks correct** — "absent from the commit response ⟺ rebased to a no-op" holds because OT `commitChanges` always echoes survivors inline (catchup + new). The re-import on drop keeps the open doc consistent. Nice and well-covered.

RyanHavoc:
Thanks for the thorough review — all points addressed in `b0f0e2f`.

## 🔴 Blocking — inverted version query

Switched `getLatestMainVersion` from `endBefore: upperBound + 1` to `startAfter: upperBound + 1`. Under `reverse`, that's the upper bound (`endRev <= upperBound`) against `FirestoreOTStore` — excludes the orphan, keeps the latest legit version, reuses the existing `(origin, endRev)` index, no pup change.

Then the conformance work you called for:
1. `LWWMemoryStoreBackend.listVersions` now applies the reverse-cursor flip, so the reference store matches pup instead of being the outlier. (Bonus: this also fixes `handleOfflineSessionsAndBatches`, which already used the `startAfter`+`reverse` idiom to find the parent main version — it now resolves "last main version *before* the batch" correctly against the in-memory store.)
2. `getSnapshotAtRevision` tests updated to expect `startAfter`; the orphan regression mock now emulates the reverse-cursor semantics instead of absolute.
3. Rewrote the `ListVersionsOptions` doc — `startAfter`/`endBefore` are cursors relative to sort order, and `reverse` flips which side of each bound they select.
4. Added a `reverse-cursor semantics (FirestoreOTStore conformance)` block in `LWWMemoryStoreBackend.spec.ts` mirroring pup's two enshrined cases (`startAfter`→upper bound, `endBefore`→lower bound) plus the snapshot-query shape and orphan exclusion. The truly *shared* cross-store conformance harness would need to live in pup where `FirestoreOTStore` is — happy to add it there as a follow-up if you'd like.

## 🟡 Non-blocking — both done, not deferred

- **Clamp logging**: `OTBranchManager.mergeBranch` now `console.warn`s when `branchedAtRev > sourceCurrentRev`, surfacing the corrupted/renumbered-source case rather than merging silently.
- **FF + conflict-retry residual**: `commitChanges` now saves *before* versioning in both branches, so a `RevConflictError` on a fast-forward save no longer strands a version minted from pre-transform changes — the retry's transform branch versions from the actually-persisted changes. `createVersion` builds metadata purely from the passed array (no store read-back), so the reorder is safe; also strictly prevents orphan versions on any save failure, same spirit as fix #1. Strengthened the existing retry test to assert the session is versioned from the post-transform rev.

Full suite (1937 tests), `type:check`, `lint`, and `format:check` all green.

jacwright (approved):
Approving — the blocking issue is fully resolved and verified.

Fix #2 now uses `startAfter: upperBound + 1` (reverse cursor), which matches what `FirestoreOTStore` already implements and indexes for `getLatestMainVersion` — so the snapshot query is correct against the production store with no pup change, no new index, no deploy risk. Excited that the supporting work all landed too:

- `LWWMemoryStoreBackend.listVersions` now applies the same reverse-cursor flip, so the reference store and prod store agree instead of diverging.
- Added `reverse-cursor semantics (FirestoreOTStore conformance)` tests that mirror pup's `FirestoreOTStore.spec.ts` assertions — exactly the cross-implementation coverage that was missing.
- `ListVersionsOptions` doc rewritten to describe cursors-relative-to-sort, dropping the contradictory "strictly before this value" line that let the two stores drift.
- Bonus: the `OTBranchManager` clamp now warns when it fires (`branchedAtRev … is ahead of source … clamping`), so corrupted/renumbered docs surface instead of merging silently.

Verified locally on `fix/branch-merge-base-clamp`: getSnapshotAtRevision, LWWMemoryStoreBackend, OTBranchManager, commitChanges specs — 133 passed.

The remaining `commitChanges` fast-forward + conflict-retry versioning residual is fine to leave as a follow-up (narrow window, strictly better than before). Nice work on the whole chain — root cause (client retry storm) through to the merge-time symptom.

PR #60, opened 2026-06-23
fix: latch isLoaded before notifying syncStatus subscribers

jacwright:
**TL;DR:** Empty documents (no content yet) could appear to "load" forever in apps using the Vue/Solid integrations, so users couldn't add their first change. Fixed.

`updateSyncStatus` set `syncStatus.state` (notifying subscribers synchronously) before `_checkLoaded()` latched `isLoaded`. So a `syncStatus` subscriber saw the new status with stale `isLoaded === false`; for an empty rev-0 doc nothing else flips `isLoaded`, so it never finishes loading. Docs with content latch via snapshot import (committedRev > 0), so only empty docs hit this. Wrapping the writes in `batch()` defers notifications until all are applied, so subscribers see consistent final values.

Regression test added in `tests/client/PatchesDoc.spec.ts` (verified to fail without the fix).

Companion app-side fix: dabblewriter/dabble-admin#59.

### DIFF
--- a/src/client/BaseDoc.ts
+++ b/src/client/BaseDoc.ts
@@ -1,4 +1,4 @@
-import { ReadonlyStoreClass, signal, store, type Store } from 'easy-signal';
+import { batch, ReadonlyStoreClass, signal, store, type Store } from 'easy-signal';
 import { applyPatch } from '../json-patch/applyPatch.js';
 import { createJSONPatch } from '../json-patch/createJSONPatch.js';
 import type { JSONPatchOp } from '../json-patch/types.js';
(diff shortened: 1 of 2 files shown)


node_modules
```diff
@@ -0,0 +1 @@
+/Users/jacob/Projects/dabble/patches/node_modules
```
RyanHavoc: This PR commits a `node_modules` **symlink** pointing at an absolute path on your machine:

```
node_modules -> /Users/jacob/Projects/dabble/patches/node_modules
```

Looks like an accidental `git add` — it'll be a dangling symlink for everyone else and on CI. Could you pull it out before merge (`git rm --cached node_modules`) and confirm `node_modules` is in `.gitignore`?

The actual change — latching `isLoaded` inside `batch()` before subscribers run, plus the regression test — is spot on and exactly the right root-cause fix. Just want to keep this stray symlink out of the tree.

PR #61, opened 2026-06-24
fix(sync): prevent silent data loss/duplication on the SSE+REST path

RyanHavoc:
## TL;DR

Fixes several ways a syncing hiccup could silently lose or duplicate a writer's text:

- A connection that quietly dies and **never reconnects** (you keep typing, nothing reaches the server).
- A missed update that **freezes the document** so you stop seeing collaborators'/other-device changes — while the app still says "synced".
- A recovery path that could **double-paste what you just typed**.

It also makes the server's keep-alive **visible** to the client so the app can tell "idle but healthy" from "dead", and adds a watchdog that rebuilds a half-open stream.

These are client/transport-correctness fixes in `@dabble/patches`. The OT server semantics are unchanged. Consumers (DW3) need a **0.12.1 release** to pick them up.

## Changes

- **SNAPIMP-1 — `OTDoc.import()` content duplication.** During sync recovery, `import()` set `_pendingChanges = snapshot.changes` and then re-applied the surviving `_optimisticOps` on top. If a just-typed change had already been persisted into the snapshot's pending set (its echo not yet shifted off the optimistic queue), it was applied twice → duplicated text for non-idempotent ops. Now de-dups optimistic ops whose ops already appear in `snapshot.changes` (consumed FIFO, structural match). Matches the live "reverts then duplicates and adds characters" report.
- **SSE-1 — stuck on `connecting`.** A fatal post-handshake SSE error (`EventSource.readyState === CLOSED`, no auto-reconnect) left the connection non-null and the client stuck `connecting` forever (`connect()` no-ops on a non-null EventSource). Now tears down + nulls the stream and surfaces `error` so the consumer's reconnect path rebuilds it.
- **SSE-3 — half-open stream + invisible heartbeat.** Added a client liveness watchdog: if no SSE frame (event **or** heartbeat) arrives within 75s, close + reconnect. `SSEServer`'s heartbeat is now an observable `event: heartbeat` (was a `:` comment EventSource ignores), so the watchdog can distinguish a healthy-but-idle stream from a dead one.
- **SSE-2 — frozen `committedRev` on a gap.** `PatchesSync._receiveCommittedChanges` now recovers a non-contiguous "Missing changes" gap by calling `syncDoc` (pull the authoritative tail), mirroring `applyMergeChanges`' fallback, instead of only emitting `onError` and silently dropping the tail.

## Tests

- `OTDoc.spec.ts`: SNAPIMP-1 repro (was red `['a','b','b']`, now green).
- `PatchesREST.spec.ts`: fatal-error teardown + reconnect; watchdog fires on silence; heartbeats keep a stream alive.
- `PatchesSync.spec.ts`: gap recovers via `syncDoc`; non-gap errors still only `onError`.
- `SSEServer.spec.ts`: heartbeat assertion updated to the observable event.

Full suite green (1958), type/lint/format clean.

## Release

Manual: needs a **0.12.1** publish on merge, then DW3 bumps its dependency (see the paired DW3 PR).

### DIFF
--- a/src/net/PatchesSync.ts
+++ b/src/net/PatchesSync.ts
@@ -1,5 +1,6 @@
 import { isEqual } from '@dabble/delta';
 import { batch, ReadonlyStoreClass, signal, store, type Store, type Unsubscriber } from 'easy-signal';
+import { MissingChangesError } from '../algorithms/ot/client/applyCommittedChanges.js';
 import { breakChangesIntoBatches, type SizeCalculator } from '../algorithms/ot/shared/changeBatching.js';
 import { BaseDoc } from '../client/BaseDoc.js';
 import type { BranchClientStore } from '../client/BranchClientStore.js';
@@ -669,10 +670,10 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
       await this._applyServerChangesToDoc(docId, serverChanges);
     } catch (err) {
       // A non-contiguous server change (we missed an earlier event — a transient SSE drop the
-      // browser's replay didn't fully cover) throws "Missing changes from the server …". Without
-      // recovery the tail is dropped and `committedRev` freezes silently, so the client stops
-      // converging while believing it is up to date. Pull the authoritative tail via syncDoc
-      // (getChangesSince), mirroring applyMergeChanges' gap fallback. Keep onError for telemetry.
+      // browser's replay didn't fully cover) throws a MissingChangesError. Without recovery the
+      // tail is dropped and `committedRev` freezes silently, so the client stops converging while
+      // believing it is up to date. Pull the authoritative tail via syncDoc (getChangesSince),
+      // mirroring applyMergeChanges' gap fallback. Keep onError for telemetry.
       if (this._isMissingChangesGap(err)) {
         try {
           await this.syncDoc(docId);
@@ -686,9 +687,9 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
     }
   }
 
-  /** True when an error is the "missing changes" revision-gap thrown by applyCommittedChanges. */
+  /** True when an error is the revision-gap thrown by applyCommittedChanges. */
   private _isMissingChangesGap(err: unknown): boolean {
-    return err instanceof Error && err.message.includes('Missing changes from the server');
+    return err instanceof MissingChangesError;
   }
 
   /**
--- a/src/net/PatchesSync.ts
+++ b/src/net/PatchesSync.ts
@@ -668,8 +668,27 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
     try {
       await this._applyServerChangesToDoc(docId, serverChanges);
     } catch (err) {
+      // A non-contiguous server change (we missed an earlier event — a transient SSE drop the
+      // browser's replay didn't fully cover) throws "Missing changes from the server …". Without
+      // recovery the tail is dropped and `committedRev` freezes silently, so the client stops
+      // converging while believing it is up to date. Pull the authoritative tail via syncDoc
+      // (getChangesSince), mirroring applyMergeChanges' gap fallback. Keep onError for telemetry.
+      if (this._isMissingChangesGap(err)) {
+        try {
+          await this.syncDoc(docId);
+          return;
+        } catch (syncErr) {
+          this.onError.emit(syncErr as Error, { docId });
+          return;
+        }
+      }
       this.onError.emit(err as Error, { docId });
     }
+  }
+
+  /** True when an error is the "missing changes" revision-gap thrown by applyCommittedChanges. */
+  private _isMissingChangesGap(err: unknown): boolean {
+    return err instanceof Error && err.message.includes('Missing changes from the server');
   }
 
   /**
--- a/src/net/rest/PatchesREST.ts
+++ b/src/net/rest/PatchesREST.ts
@@ -22,6 +22,16 @@ import { normalizeIds } from './utils.js';
 const SESSION_STORAGE_KEY = 'patches-clientId';
 const REQUEST_TIMEOUT_MS = 30_000;
 const CONNECT_TIMEOUT_MS = 30_000;
+/**
+ * If no SSE frame (a committed-change/signal/resync event OR the server's periodic
+ * `heartbeat` event) arrives within this window, the watchdog treats the stream as
+ * half-open and forces a reconnect. Must exceed the server heartbeat interval
+ * (default 30s) with margin for jitter/proxies — a half-open EventSource can stay
+ * `connected` forever without ever firing `onerror`, silently freezing the client.
+ */
+const LIVENESS_TIMEOUT_MS = 75_000;
+/** How often the liveness watchdog checks for a stalled stream. */
+const LIVENESS_CHECK_MS = 15_000;
 
 /**
  * Options for creating a PatchesREST instance.
@@ -71,6 +81,10 @@ export class PatchesREST implements PatchesConnection {
   private options: PatchesRESTOptions;
   private shouldBeConnected = false;
   private onlineUnsubscriber: Unsubscriber | null = null;
+  /** Timestamp (ms) of the last SSE frame received (event or heartbeat). 0 until first open. */
+  private _lastEventAt = 0;
+  /** Interval handle for the half-open-stream watchdog; null while not connected. */
+  private livenessTimer: ReturnType<typeof globalThis.setInterval> | null = null;
 
   constructor(url: string, options?: PatchesRESTOptions) {
     this._url = url.replace(/\/$/, ''); // Strip trailing slash
@@ -139,6 +153,8 @@ export class PatchesREST implements PatchesConnection {
       }, CONNECT_TIMEOUT_MS);
 
       es.onopen = () => {
+        this._noteTraffic();
+        this._startLivenessWatchdog();
         this._setState('connected');
         if (!settled) {
           settled = true;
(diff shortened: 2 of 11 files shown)


src/net/rest/SSEServer.ts
```diff
@@ -380,7 +380,10 @@ export class SSEServer {
 
   private _startHeartbeat(): void {
     this.heartbeatInterval = globalThis.setInterval(() => {
-      const heartbeat = ': heartbeat\n\n';
+      // Named event (not a `:` comment) so the EventSource client can observe it and
+      // distinguish a healthy-but-idle stream from a half-open one. A comment is invisible
+      // to EventSource handlers, which left clients unable to detect a silently dead stream.
+      const heartbeat = 'event: heartbeat\ndata: \n\n';
```
jacwright, line 394: Rollout-skew blocker. This swaps the old `:` comment heartbeat for a named `event: heartbeat`. Live pup still runs @dabble/patches 0.9.13 (pins `^0.9.13`), whose SSEServer emits `": heartbeat\n\n"` - an SSE comment EventSource silently discards. A client built on this PR runs the 75s liveness watchdog, which only resets on observable frames (named events + `addEventListener('heartbeat')`). Against today's pup, an idle doc gets zero observable frames for 75s, so the watchdog tears the stream down to `error`. Gate the release on server-first deploy: bump pup's @dabble/patches and deploy the named-heartbeat SSEServer before any watchdog client ships (DW3 pins `^0.12.0`).

src/net/rest/PatchesREST.ts
```diff
@@ -363,6 +392,45 @@ export class PatchesREST implements PatchesConnection {
...
+   * force a reconnect: tear the stream down and surface `error`, which drives the
+   * consumer's reconnect path (re-subscribe + `syncAllKnownDocs` catch-up).
+   */
+  private _startLivenessWatchdog(): void {
+    this._stopLivenessWatchdog();
+    this.livenessTimer = globalThis.setInterval(() => {
+      if (!this.shouldBeConnected || onlineState.isOffline || !this.eventSource) return;
+      if (Date.now() - this._lastEventAt <= LIVENESS_TIMEOUT_MS) return;
+      this._teardownStream();
+      this._setState('error');
```
jacwright, line 447: This watchdog path (and the fatal-CLOSED branch in `onerror`) ends at `_teardownStream()` + `_setState('error')` with no reconnect. WebSocketTransport self-schedules reconnect with backoff (`_scheduleReconnect`); PatchesREST does not. I traced the consumer too: DW3 #680's `error` transition only calls `reportSyncError` (Sentry), and the sole `connect()` trigger is `onlineState.onOnlineChange`, which won't fire on a stable connection. So a fatal error or watchdog teardown surfaces `isSyncDegraded` but never rebuilds the stream - the watchdog detects a half-open stream and then doesn't reconnect. Where is the rebuild meant to happen? Either mirror WebSocketTransport's reconnect here, or add a `connect()`-on-`error` supervisor in #680 (and test it).

src/net/rest/PatchesREST.ts
```diff
@@ -22,6 +22,16 @@ import { normalizeIds } from './utils.js';
...
 const REQUEST_TIMEOUT_MS = 30_000;
 const CONNECT_TIMEOUT_MS = 30_000;
+/**
+ * If no SSE frame (a committed-change/signal/resync event OR the server's periodic
+ * `heartbeat` event) arrives within this window, the watchdog treats the stream as
+ * half-open and forces a reconnect. Must exceed the server heartbeat interval
+ * (default 30s) with margin for jitter/proxies — a half-open EventSource can stay
+ * `connected` forever without ever firing `onerror`, silently freezing the client.
+ */
+const LIVENESS_TIMEOUT_MS = 75_000;
```
jacwright, line 37: `LIVENESS_TIMEOUT_MS` (75s) is decoupled from the server's configurable `heartbeatIntervalMs` (SSEServer default 30s), in a separate module with no shared source of truth. Raising the server interval above ~37s would false-fire the watchdog on healthy idle streams. Benign at the default - worth a comment stating the invariant (timeout >= ~2x heartbeat interval).

src/net/PatchesSync.ts
```diff
@@ -668,10 +668,29 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
...
+          return;
+        }
+      }
       this.onError.emit(err as Error, { docId });
     }
   }
 
+  /** True when an error is the "missing changes" revision-gap thrown by applyCommittedChanges. */
+  private _isMissingChangesGap(err: unknown): boolean {
+    return err instanceof Error && err.message.includes('Missing changes from the server');
```
jacwright: Gap recovery keys on the free-text substring `'Missing changes from the server'`; the sole producer is a plain `Error` in applyCommittedChanges. A reword there silently disables gap recovery and reverts to the silent tail-drop this PR fixes, with no compile-time link. Prefer a typed error checked via `instanceof`, or a shared message constant used by both sites.

jacwright (changes requested):
Reviewed the four fixes against the PR branch plus live pup and DW3 #680 (not just this repo). The OT/sync logic is sound in isolation: SNAPIMP-1's optimistic-op dedup, SSE-2's gap recovery via `syncDoc` (reuses the contiguity check; watermark only advances on success), and the watchdog/teardown lifecycle all hold up, with focused tests.

Two cross-repo issues to resolve before merge/release:

**Blocker - heartbeat rollout skew.** The heartbeat changes from a `:` comment to a named `event: heartbeat`, but live pup runs patches 0.9.13 (comment heartbeat). A new client's 75s watchdog only counts observable frames, so on an idle doc against today's pup it tears down to `error`. Needs server-first deploy ordering: pup must ship the named-heartbeat build before any watchdog client. (Inline on SSEServer.ts.)

**Major - no reconnect on `error`.** Both fatal paths end at teardown + state `error`; nothing rebuilds. WebSocketTransport self-reconnects; PatchesREST doesn't, and DW3 #680 only reports to Sentry (no `connect()`), with the sole connect trigger being offline->online. So the watchdog detects a dead stream but never reconnects. (Inline on PatchesREST.ts.)

**Minors:** free-text coupling in `_isMissingChangesGap`; `LIVENESS_TIMEOUT_MS` decoupled from server `heartbeatIntervalMs`; and release hygiene - the branch is 0.12.0 but that appears already published (DW3 resolves `^0.12.0`; main is 0.11.0), so this needs to ship as 0.12.1.

Nice work tracking these down - the fixes themselves are right; it's the deploy ordering and the reconnect handoff to the consumer that need closing.

src/net/rest/PatchesREST.ts
RyanHavoc: Fixed in b932365 — `PatchesREST` now self-reconnects, mirroring `WebSocketTransport`.

`_scheduleReconnect()` runs exponential backoff (1s → ×1.5 → 30s cap), single-flight, gated on `shouldBeConnected && !isOffline`. It is now invoked from every dead-stream path: the fatal `CLOSED` branch in `onerror`, the watchdog teardown, the initial-handshake failure, and the connect-timeout — each does teardown → `error` → `_scheduleReconnect()`. A healthy `onopen` resets the backoff and cancels any pending reconnect; `disconnect()` and the offline transition cancel it.

The eventual `connected` re-emit drives the consumer’s existing `syncAllKnownDocs` (re-subscribe + pull authoritative tail), so the rebuild is fully in-transport and #680 needs no `connect()`-on-`error` supervisor. Added tests: reconnect after a fatal error, reconnect after watchdog teardown, and no-reconnect-after-`disconnect()`.

src/net/rest/PatchesREST.ts
RyanHavoc: Added the invariant to the constant’s doc comment: `LIVENESS_TIMEOUT_MS` must stay `>= ~2x heartbeatIntervalMs` (plus jitter/proxy margin), with a note that raising the server interval above ~37s means raising this in step. Left them as separate constants since they live in different layers (client transport vs. `SSEServer`); the comment is the shared source of truth.

src/net/PatchesSync.ts
RyanHavoc: Fixed — `applyCommittedChanges` now throws a typed `MissingChangesError` (carrying `expectedRev`/`gotRev`/`sinceRev`), and `_isMissingChangesGap` checks `err instanceof MissingChangesError`. Compile-time link now, so rewording the message can’t silently disable gap recovery. The `applyCommittedChanges.spec` message assertion still holds (same string); the `PatchesSync` gap test and its module mock were updated to use the real class.

src/net/rest/SSEServer.ts
RyanHavoc: Resolved by keeping the strict watchdog and gating on server-first deploy, as you suggested. Documented the hard ordering in three places: a `WIRE-FORMAT CHANGE — REQUIRES SERVER-FIRST DEPLOY ORDERING` comment right here on the heartbeat, a callout in `docs/sse-rest.md`, and the PR release notes — pup must ship the named-heartbeat `SSEServer` build (bump pup’s `@dabble/patches`) before any watchdog client.

Worth flagging: now that `PatchesREST` self-reconnects (the Major above), this is a *hard* gate, not just cosmetic — against today’s comment-heartbeat pup an idle doc would tear down → reconnect every ~75s in a loop until pup is upgraded. So the deploy sequence is: publish 0.12.1 → bump+deploy pup → only then ship the DW3 client.

RyanHavoc:
@jacwright thanks for the cross-repo trace — all four addressed in b932365 (full suite green, 1961 tests; type/lint/format clean):

- **Major — reconnect.** `PatchesREST` now self-reconnects with exponential backoff (mirrors `WebSocketTransport`) from every dead-stream path (fatal `CLOSED`, watchdog teardown, initial-handshake failure, connect timeout). Rebuild is fully in-transport; the `connected` re-emit drives `syncAllKnownDocs`, so #680 needs no supervisor. +3 tests.
- **Blocker — heartbeat skew.** Kept the strict watchdog; documented the hard server-first deploy ordering (SSEServer comment + `docs/sse-rest.md` + release notes below). With self-reconnect this is now a hard gate, so the sequence is **publish 0.12.1 → bump+deploy pup → ship DW3**.
- **Minor — typed gap error.** `MissingChangesError` (instanceof) replaces the substring match.
- **Minor — liveness invariant.** Documented `LIVENESS_TIMEOUT_MS >= ~2x heartbeatIntervalMs`.
- **Release.** Bumped to **0.12.1** (0.12.0 was already published).

Ready for another look.

jacwright (approved):
Re-reviewed at b932365 — all four points from the prior review are addressed; verified each against the code and ran the affected specs (223 passing).

- **Major (no reconnect on `error`):** PatchesREST now self-reconnects with exponential backoff (1s -> x1.5 -> 30s cap), single-flight, gated on intent + online, and canceled on `disconnect()`/offline. `connect()` guards against a double EventSource, and reaching `connected` drives `PatchesSync.syncAllKnownDocs()` so the tail missed during the dead window is pulled. New tests cover the backoff reconnect, watchdog-triggered reconnect, and no-reconnect-after-`disconnect()`. Resolved.
- **Typed gap error:** `MissingChangesError` (structured `expectedRev`/`gotRev`/`sinceRev`); `_isMissingChangesGap` now uses `instanceof`. Producer -> consumer coupling is type-checked. Resolved.
- **`LIVENESS_TIMEOUT_MS`:** invariant (>= ~2x server `heartbeatIntervalMs`) documented. Resolved.
- **Release:** bumped to 0.12.1. Resolved.

**Blocker (heartbeat rollout skew):** resolved as a documented deploy-ordering gate (SSEServer.ts comment + docs/sse-rest.md) — the right call, since patches can't change old-pup behavior. One operational reminder for release: deploy the named-heartbeat server (bump + deploy pup's @dabble/patches) BEFORE shipping any watchdog client, or idle beta docs flap to `error` every ~75s against current pup (0.9.13). Note the self-reconnect makes that a ~75s flap loop rather than a one-time error, so the ordering matters.

Approving. Nice turnaround.

PR #62, opened 2026-06-25
fix(sync): spoke-stable change ids for safe, idempotent write retries

RyanHavoc:
## TL;DR
Closes the deferred write-side data-loss path behind "pasted text disappears." A spoke→hub change submit that times out (tab-election 30s "Call timed out" / leader death) can now be **safely retried** instead of discarding the just-typed text — because the change id is minted on the spoke (stable across retries) and the hub/server dedup it by id.

## Why this was deferred
The change `id` is minted on the **hub** (`OTAlgorithm._createChangesFromOps` → `createChangeId(rev)`), so a blind retry re-mints a *new* id the server can't recognise → duplicate. That's why the client discarded the text via `rollbackOptimistic()` rather than retrying.

## What changed (additive, backward compatible)
- **`data/change.ts`** — `createChange(baseRev, rev, ops, metadata?, id?)` honours an explicit, rev-independent stable id (verified nothing parses the rev out of the id — `inc.from(rev)` lives only in `change.ts`; every id consumer treats it as opaque).
- **`client/OTAlgorithm.ts`** — `handleDocChange(…, id?, isRetry?)`: on a retry, `_findChangeById` checks pending first (the dominant case — a "Call timed out" means the hub was slow, so the change is still pending) then a bounded recent-committed window, and returns the existing change instead of re-minting. Otherwise mints with the supplied id. The existing `commitChanges` id-dedup (`commitChanges.ts:138-144`) is the server-side backstop.
- **`client/ClientAlgorithm.ts`** — interface gains the two optional params.
- **`client/LWWAlgorithm.ts`** — accepts + ignores them (LWW resolves by timestamp/path).

The spoke-side retry loop that uses this lives in the consuming app (Dabble Writer 3.0 — separate PR).

## Tests
- `tests/client/OTAlgorithm-stableId.spec.ts` (8): mints with the stable id; a retry returns the already-accepted change (pending + committed-tail) with no duplicate; the guard is retry-gated; terminal cases mint fresh.
- Full suite **1969 green**; type/lint/format clean.

## Release
Bumps **0.12.1 → 0.12.2** (manual publish — patches has no release-please). The DW3 PR pins `^0.12.2`, so this must publish before that one's CI/merge.

## Residual (documented)
One uncovered compound case — original change *confirmed* AND the hub *restarts* between confirm and retry AND `baseRev` advanced past the original rev — falls to pup's `baseRev`-windowed id-dedup, which can miss it. Doubly-rare; vs today's unconditional discard it's a large improvement. Follow-up could widen the pup dedup window.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@dabble/patches",
-  "version": "0.12.0",
+  "version": "0.12.2",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@dabble/patches",
-      "version": "0.12.0",
+      "version": "0.12.2",
       "license": "MIT",
       "dependencies": {
         "@dabble/delta": "^1.2.5",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dabble/patches",
-  "version": "0.12.1",
+  "version": "0.12.2",
   "description": "Immutable JSON Patch implementation based on RFC 6902 supporting operational transformation and last-writer-wins",
   "author": "Jacob Wright <jacwright@gmail.com>",
   "bugs": {
--- a/src/client/ClientAlgorithm.ts
+++ b/src/client/ClientAlgorithm.ts
@@ -54,13 +54,21 @@ export interface ClientAlgorithm {
    * @param ops The JSON Patch ops to process
    * @param doc The open doc instance, or undefined if in Worker (no docs)
    * @param metadata Metadata to attach to the change
+   * @param id Optional caller-supplied stable change id. Lets an upstream caller (e.g. a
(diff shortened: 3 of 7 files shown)


jacwright (approved):
Verified the design end-to-end against source: `createChange` honours the optional id (rev-derived fallback intact), `_findChangeById` checks pending first then a bounded committed-tail window, and `commitChanges` (138-144) is the server backstop. Confirmed the id is treated as opaque everywhere (`inc.from(rev)` lives only in `change.ts`; changes order by rev, pending keyed by `[docId, rev]`), so a plain spoke-minted id is safe. `OTIndexedDBStore.listChanges` is live with matching `startAfter` semantics, so the committed-tail fallback works on the hub. Tests are thorough, CI green. LGTM.

PR #63, opened 2026-06-26
feat(ot-server): version on change count to bound cold-load replay

RyanHavoc:
## TL;DR

Versioning currently only fires on an **inactivity gap** (`sessionTimeoutMinutes`). A document edited in a **continuous high-rate stream** (changes seconds apart) never hits that gap, so it can pile up tens of thousands of un-versioned changes. Cold loads then replay the whole change log on top of a stale snapshot — and that can grow large enough that the document **can no longer be loaded at all**. This adds a **count-based versioning trigger** so a snapshot is taken at least every N changes regardless of timing, keeping cold-load replay bounded.

This is the systemic fix behind a recent incident where a plot-grid runaway bloated one document's change log to ~24,000 entries; the server snapshot was stuck thousands of revisions back, so every fresh session had to rebuild the doc from the full log and timed out / OOM'd. A count-based snapshot would have kept that document servable throughout.

## What changed

- **`OTServerOptions.maxChangesPerVersion`** — new option, **default `1000`**, `0` disables. Threaded through to the `commitChanges` algorithm via `CommitChangesOptions`.
- In `commitChanges`, after the existing session-gap check: if a commit pushes the tip **across a `maxChangesPerVersion` boundary** *and* at least that many changes have accrued since the last version, snapshot forward by **at most N**.

### Design notes
- **Cheap gate:** the boundary check is `floor(initialRev/N) < floor(newTipRev/N)` — pure arithmetic, no store read. The extra `listVersions` lookup only happens when the gate fires (~once per N commits), so the hot path is unaffected on normal commits.
- **Bounded forward step:** versions advance by at most N (`min(tip, lastVersionEndRev + N)`), so even a document far behind its last version is caught up in bounded chunks — the in-commit change scan and the out-of-band state build never load a huge backlog at once. Cold-load replay stays `< 2N`.
- **Safe-by-default:** defaulting on (1000) means this failure mode is prevented even if a consumer doesn't configure it — which is exactly the oversight that produced the incident. Consumers can tune or disable via the option.
- Reuses the existing `createVersionAtRev` path (same metadata, same out-of-band state build) — no new versioning machinery.

## Tests
- `commitChanges`: boundary fires; sub-boundary no-op; boundary crossed but `< N` accrued → no version; **bounded forward step** for a backlogged doc (versions to rev 10, not 99); disabled at `0`.
- `OTServer`: a configured `maxChangesPerVersion` triggers count-based versioning; the default threshold doesn't fire on a tiny commit.
- Full suite: **1978 passing**; `type:check`, `lint`, `format:check` clean.

## Rollout
Bumped to **`0.12.3`** (additive, backward-compatible). pup depends on `^0.12.0`, so it picks this up on a lockfile update — no range change needed. After pup adopts it, new documents are protected automatically; already-backlogged docs compact on their next edit-burst (each boundary crossing snapshots forward a bounded step). pup can override `maxChangesPerVersion` if it wants a different bound.

### DIFF
--- a/docs/OTServer.md
+++ b/docs/OTServer.md
@@ -47,12 +47,14 @@ const server = new OTServer(store, {
 
 ### Options
 
-| Option                  | Type     | Default | Description                                                                                       |
-| ----------------------- | -------- | ------- | ------------------------------------------------------------------------------------------------- |
-| `sessionTimeoutMinutes` | `number` | `30`    | Minutes of inactivity before creating a new version snapshot                                      |
-| `maxChangesPerVersion`  | `number` | `1000`  | Snapshot at least every N changes regardless of timing, bounding cold-load replay cost; `0` = off |
+| Option                  | Type     | Default | Description                                                                                            |
+| ----------------------- | -------- | ------- | ------------------------------------------------------------------------------------------------------ |
+| `sessionTimeoutMinutes` | `number` | `30`    | Minutes of inactivity before creating a new version snapshot                                           |
+| `maxChangesPerVersion`  | `number` | `1000`  | Snapshot forward in bounded steps of at most N changes when the un-versioned tail reaches N; `0` = off |
 
-> **Why `maxChangesPerVersion`?** Session-gap versioning only fires when consecutive changes are far apart in time. A continuous high-rate stream of changes (seconds apart) never triggers it, so without a count-based trigger a single document can accrue tens of thousands of un-versioned changes — every cold load then replays the entire log, which can grow large enough that the document can no longer be loaded at all. The count trigger snapshots forward in bounded steps so both the snapshot build and cold-load replay stay bounded.
+> **Why `maxChangesPerVersion`?** Session-gap versioning only fires when consecutive changes are far apart in time. A continuous high-rate stream of changes (seconds apart) never triggers it, so without a count-based trigger a single document can accrue tens of thousands of un-versioned changes — every cold load then replays the entire log, which can grow large enough that the document can no longer be loaded at all. The count trigger snapshots forward in **bounded steps of at most N changes**, so each snapshot build stays cheap.
+>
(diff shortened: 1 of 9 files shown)


RyanHavoc:
## Review + fixes pushed (`ef2a544`)

**TL;DR:** the design is sound, but the bounded step was accounted **per-commit, not per-change**, so the headline `< 2N` cold-load guarantee didn't actually hold. Fixed in `ef2a544`; all checks green (1980 tests, type/lint/format clean).

### The defect
The count branch advanced the version watermark by at most `N` per `commitChanges` call, while the tip advances by the commit's batch size:

- **Backlogged doc never compacts** — a doc already `> N` behind tracked the tip at a *constant lag forever* (e.g. tip 24000, watermark 100, N=10 → gap stays ~23900). So the "already-backlogged docs compact on their next edit-burst" claim didn't hold; those still needed the server-side compaction we did for the plot-grid runaway.
- **`batchSize > N` diverges** — a sustained stream of commits larger than N grew the un-versioned floor without bound.

The realistic high-rate case (many small commits) *was* bounded, and offline/bulk uploads version in full via `handleOfflineSessionsAndBatches`, so the default-config Hank scenario was prevented — but the guarantee in the docs/option text was false as written.

### What `ef2a544` changes
- **Bounded catch-up loop** (`catchUpVersions`): the watermark now drains toward the prior tip in as many N-sized steps as the backlog needs, *within the one commit*. Each build still loads `≤ N` changes (flat memory); only the step count scales with the backlog — a one-time heal, then steady-state commits run 1–2 steps.
- **Session-gap path bounded too** — it previously built the whole backlog in one shot (the residual OOM path); now it drains in N-sized steps (and versions the final `< N` tail). Falls back to the original single version when count versioning is disabled.
- **Gate projects the server tip** `max(initialRev + len, claimedRev)` so a stale-`baseRev` commit can't slip the boundary past.
- **`createVersionAtRev` `startAfterRev`** so the loop reads `listVersions` once, not per step.
- **`store.createVersion` idempotency expectation documented** — concurrent instances can race to create overlapping versions during high-rate streaming (not a data-loss hazard; the store should dedupe by range).
- **Docs/comments corrected** — it's a *preventative* bound (`< ~2N` going forward from near-current), **not** remediation; already-backlogged/unservable docs still need server-side compaction. Noted it's on by default.
- **Tests** — backlog draining in bounded builds, bounded session-gap drain incl. tail, versioning-runs-once across a RevConflict retry.

### Left intentionally (not blocking)
- **Version `0.12.3` / default `1000`** kept as-is — the patch bump stays inside pup's `^0.12.0` range and safe-by-default is the right call. The docs now make the default-on behavior change explicit.
- **The PR description above still describes the old single-step behavior** ("versions to rev 10, not 99" / unqualified `< 2N`) — worth updating to the draining/bounded-step model before merge.

PR #64, opened 2026-06-26
fix(ot-client): serialize receive-rebase vs local mint (silent change loss)

jacwright:
**TL;DR:** While investigating the production "inconsistent baseRev" rejections, a multi-agent root-cause run ruled out the original theory and surfaced a *different*, real bug in the sync client: under a specific timing race, a user's keystroke (or an incoming edit) could be silently dropped — no error, the change just disappears. This fixes that race. It is **not** the fix for the baseRev 500s (that root cause is still being chased in a separate session); it's a genuine data-loss bug found along the way.

## The bug

On the hub sync path there's one shared `OTAlgorithm` with no open doc. Two methods each do an unguarded read-modify-write on the `[docId, rev]`-keyed pending store:

- `handleDocChange` (mint a local edit): reads `committedRev`/`pendingRev`, mints a change at `pendingRev + 1`, writes it.
- `applyServerChanges` (apply an incoming server change): reads the snapshot, rebases pending to new revs, writes them.

Nothing serializes them against each other (the hub's `_changeQueues` only serializes mint-vs-mint; `@blockable` is inert here because nothing calls `blockWhile`). So they interleave: the rebased local change and the freshly minted change **both compute the same rev**, and because both writers use `put()` on a `[docId, rev]` key, the second overwrites the first. One change vanishes — a typed edit, or a rebased committed change. Silent.

(This is also why it does *not* produce the server `consistent baseRev` error on this path: the collision collapses the two changes into one before they could ever be sent as a mixed-baseRev batch.)

## The fix

A per-doc FIFO mutex in `OTAlgorithm` (`_withDocLock`) wrapping `handleDocChange`, `applyServerChanges`, and `dropResolvedPending`. Same-doc operations run one at a time, in order, each to completion — so the next op always reads post-rebase state and allocates a fresh rev. No change is lost.

Fixing it at the algorithm layer covers every caller: both patches' own `PatchesSync` and the app's custom hub go through these methods on the one shared `OTAlgorithm` instance.

## Test

`tests/client/OTAlgorithm-interleave.spec.ts` uses the **real** `OTIndexedDBStore` over `fake-indexeddb` on purpose — the bug is a rev-key collision, which `OTInMemoryStore` (an array, no rev key) can't reproduce. Verified it **fails without the lock** (pending collapses to 1 change) and **passes with it** (both changes survive at distinct revs, one uniform baseRev).

Full suite green: 1981 tests, lint/type/format clean.

## Scope note

Found via the baseRev investigation but distinct from it. The actual `consistent baseRev` trigger (offline/batch accumulation, doc-reload re-seed, or reconnect re-send) is still open and being reproduced against the affected user's local DB in another session. A defensive send-boundary baseRev-split guard is still worth adding separately to neutralize that 500 regardless of cause.

### DIFF
--- a/src/client/OTAlgorithm.ts
+++ b/src/client/OTAlgorithm.ts
@@ -207,11 +207,12 @@ export class OTAlgorithm implements ClientAlgorithm {
   }
 
   async deleteDoc(docId: string): Promise<void> {
-    return this.store.deleteDoc(docId);
+    // Under the lock too: a delete must not race an in-flight mint/apply for the doc.
+    return this._withDocLock(docId, () => this.store.deleteDoc(docId));
   }
 
   async confirmDeleteDoc(docId: string): Promise<void> {
-    return this.store.confirmDeleteDoc(docId);
+    return this._withDocLock(docId, () => this.store.confirmDeleteDoc(docId));
   }
 
   async close(): Promise<void> {
@@ -221,10 +222,16 @@ export class OTAlgorithm implements ClientAlgorithm {
   // --- Private helpers ---
 
   /**
-   * Run `fn` exclusively per `docId`: calls for the same doc run one at a time, FIFO, each
-   * to completion (none collapsed or dropped). Makes a read-modify-write on a doc's pending
+   * Run `fn` exclusively per `docId`: same-doc calls run one at a time, FIFO, each to
+   * completion (none collapsed or dropped). Makes a read-modify-write on a doc's pending
    * changes atomic against another, so a receive-rebase and a local mint can't interleave
    * between each other's read and write and clobber one change at the shared [docId, rev] key.
+   *
+   * The lock is per-instance, so it relies on a single OTAlgorithm serving both mint and
+   * receive for a given doc (true today); two instances over one database would still race.
+   * handleDocChange is additionally serialized upstream by Patches._changeQueues (mint-vs-mint,
+   * which also owns onChange/optimistic-rollback); this adds the mint-vs-receive exclusion that
+   * queue lacks. The overlap is harmless — independent locks, no contention.
    */
   private _withDocLock<R>(docId: string, fn: () => Promise<R>): Promise<R> {
     const prior = this._docLocks.get(docId) ?? Promise.resolve();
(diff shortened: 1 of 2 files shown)


matalina (approved):
Approving — solid fix. The root-cause analysis checks out: `Patches._changeQueues` only serializes mint-vs-mint, while incoming server changes flow through `PatchesSync._applyServerChangesToDoc` → `algorithm.applyServerChanges` and bypass that queue, so the read-modify-write on the `[docId, rev]` store genuinely interleaves. Putting the lock around the three methods is the right chokepoint — `OTAlgorithm` is the only caller of `store.savePendingChanges` / `applyServerChanges` / `dropPendingChanges`, so every path is covered. The mutex itself is correct (`prior.then(fn, fn)` survives a failed prior op, the stored `.catch()` tail keeps the chain alive while `run` still carries the real outcome, GC only deletes when still the tail), and there's no reentrancy/deadlock. Test is well-designed: real `OTIndexedDBStore` (the collision can't reproduce on the in-memory array store) with assertions that hold under both lock orderings.

A few non-blocking notes for follow-up — none need to hold this PR:

1. **`handleDocChange` is now serialized twice** — once by `Patches._changeQueues`, again by `_withDocLock`. Harmless (separate lock maps, no contention/deadlock), and the outer queue still owns `onChange.emit` + optimistic rollback so it can't just be deleted. Worth a comment noting the partial redundancy, or a later pass to consolidate.

2. **The lock is per-`OTAlgorithm`-instance** (in-memory `Map`), so correctness relies on a single shared instance serving both mint and receive for a given doc — true today, but if a consumer ever stood up two instances over the same IndexedDB database they'd race again with no protection. A sentence in the `_withDocLock` doc comment making that assumption explicit would be good.

3. **Out of scope / pre-existing:** `deleteDoc` / `confirmDeleteDoc` aren't under the lock, so a delete could still race an in-flight mint/apply. Just flagging it's untouched.

jacwright:
Thanks @matalina — addressed all three in 061b853 (kept them in this PR since they were small):

1. Documented the intentional overlap with `Patches._changeQueues` (mint-vs-mint) vs this lock's mint-vs-receive exclusion, in the `_withDocLock` doc comment.
2. Spelled out the per-instance assumption (single `OTAlgorithm` per doc serves both mint and receive) in the same doc comment.
3. Brought `deleteDoc`/`confirmDeleteDoc` under the lock so a delete can't race an in-flight mint/apply — same one-line wrap, no new races.

Full suite still green (1981), lint/type/format clean.

PR #65, opened 2026-06-26
fix(ot-client): heal inconsistent pending baseRev on send

jacwright:
## TL;DR

Some users hit a state where their edits to a project stopped saving to the server entirely. Every sync attempt failed and the project stayed stuck. This fixes the client so those users' pending edits flush successfully again, and so the same glitch can't strand anyone in the future. No data is lost.

## What was happening

Pup rejected the user's change batch with a 500: "Client changes must have consistent baseRev". The OT client sends a doc's pending changes as one batch, and the server requires every change in the batch to share the same `baseRev`, since they're a contiguous sequence on top of the same committed revision.

In the reported case, one change in a 123-change queue carried a stale `baseRev` (1791) while the rest were at 1794. That's a race: minting a local change reads `committedRev` and then does an async store write, and a server-echo rebase landing in that window advances `committedRev` and rewrites the other pending changes' baseRev, leaving the just-minted one behind on the old value. The mixed batch is then rejected on every flush, wedging sync permanently.

## The fix

`OTAlgorithm.getPendingToSend` now re-stamps the whole pending queue to the current `committedRev` before sending, enforcing the invariant at the send boundary. It is lossless (these changes were already applied on top of `committedRev` locally) and self-healing (the next server echo rebases the stored pending back into agreement), so it repairs an already-corrupted queue rather than only preventing new ones.

- Re-stamps via a small `_withConsistentBaseRev` helper; a no-op when the queue is already consistent (the common path).
- Regression test covering a stale-baseRev straggler, the already-consistent no-op, and the empty queue.
- Verified against the real failing payload: the lone stale baseRev normalizes and the batch passes the server's consistency check, with revs/ids/ops untouched.

## Relationship to the receive-vs-mint lock

This is the healing half of the same root cause. A separate branch adds a per-doc lock that prevents the race from creating new mixed-baseRev queues. They're complementary and independent: a lock can't retroactively repair data already persisted in a user's IndexedDB, which is what unblocks the reported user here. The two touch different methods and merge cleanly.

## Test plan

- `npm run test` (77 files, 1983 tests green; new spec added)
- `npm run type:check`, `npm run lint`, `npm run format:check` clean

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dabble/patches",
-  "version": "0.12.3",
+  "version": "0.12.4",
   "description": "Immutable JSON Patch implementation based on RFC 6902 supporting operational transformation and last-writer-wins",
   "author": "Jacob Wright <jacwright@gmail.com>",
   "bugs": {
--- a/src/client/OTAlgorithm.ts
+++ b/src/client/OTAlgorithm.ts
@@ -112,8 +112,12 @@ export class OTAlgorithm implements ClientAlgorithm {
   }
 
   async getPendingToSend(docId: string): Promise<Change[] | null> {
-    const pending = await this.store.getPendingChanges(docId);
-    return pending.length > 0 ? pending : null;
+    // getDoc reads pending and the rev it's based on in one store transaction, so a concurrent
+    // receive-rebase can't advance committedRev between the two reads. Heal only the outgoing
+    // batch; the next server echo rebases the stored copy into agreement.
+    const snapshot = await this.store.getDoc(docId);
+    if (!snapshot || snapshot.changes.length === 0) return null;
+    return this._withConsistentBaseRev(snapshot.changes, snapshot.rev);
   }
 
   async applyServerChanges<T extends object>(
@@ -244,6 +248,16 @@ export class OTAlgorithm implements ClientAlgorithm {
       if (this._docLocks.get(docId) === tail) this._docLocks.delete(docId);
     });
     return run;
+  }
+
+  /**
+   * Pending OT changes all sit on `committedRev`, so the queue must share `baseRev ===
+   * committedRev` — the consistency the server enforces. Re-stamp any straggler a receive-vs-mint
+   * race left on a stale baseRev; lossless, and the next echo heals the stored copy.
+   */
+  private _withConsistentBaseRev(pending: Change[], committedRev: number): Change[] {
+    if (pending.every(c => c.baseRev === committedRev)) return pending;
+    return pending.map(c => (c.baseRev === committedRev ? c : { ...c, baseRev: committedRev }));
   }
 
   /**
--- a/tests/client/OTAlgorithm-getPendingToSend.spec.ts
+++ b/tests/client/OTAlgorithm-getPendingToSend.spec.ts
@@ -1,0 +1,59 @@
+import { beforeEach, describe, expect, it } from 'vitest';
+import { OTAlgorithm } from '../../src/client/OTAlgorithm';
+import { OTInMemoryStore } from '../../src/client/OTInMemoryStore';
+import { createChange } from '../../src/data/change';
+
+/**
+ * The send choke point enforces the OT pending invariant: every pending change must share
+ * `baseRev === committedRev`, since pending is a contiguous sequence applied on top of the
+ * committed revision. A receive-rebase racing a local mint could persist one change with a
+ * stale baseRev among rebased siblings (mixed `baseRev`), and the server rejects that batch
+ * ("Client changes must have consistent baseRev"), wedging sync forever. getPendingToSend
+ * re-stamps the batch to the current committedRev so an already-corrupted queue still flushes.
+ */
+describe('OTAlgorithm.getPendingToSend baseRev normalization', () => {
+  let store: OTInMemoryStore;
+  let algorithm: OTAlgorithm;
+
+  beforeEach(async () => {
+    store = new OTInMemoryStore();
+    algorithm = new OTAlgorithm(store);
+    await store.trackDocs(['doc1']);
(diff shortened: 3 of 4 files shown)


src/client/OTAlgorithm.ts
```diff
@@ -106,7 +106,8 @@ export class OTAlgorithm implements ClientAlgorithm {
 
   async getPendingToSend(docId: string): Promise<Change[] | null> {
     const pending = await this.store.getPendingChanges(docId);
-    return pending.length > 0 ? pending : null;
+    if (pending.length === 0) return null;
+    return this._withConsistentBaseRev(pending, await this.store.getCommittedRev(docId));
```
matalina: **Anchor is read non-atomically (minor, same race class as the bug).** `getPendingChanges` and `getCommittedRev` are two separate store reads / IndexedDB transactions. A concurrent `applyServerChanges` landing *between* them advances `committedRev` and rebases the stored pending — but `pending` here is the pre-rebase snapshot. You'd then stamp pre-rebase ops with a `baseRev` higher than they actually sit on; the server sees `baseRev === currentRev`, skips transformation, and can diverge from the now-rebased local state. Narrow window, and the next echo heals the stored copy, but it reintroduces the read-skew this PR is fixing.

Also: this adds a store round-trip on *every* flush, including the common already-consistent path.

Two ways to close both, your call on the trade-off:
- Anchor to the queue itself: `Math.max(...pending.map(c => c.baseRev))`. Self-consistent with the ops actually being sent, no extra read, no skew. In the straggler case `max === committedRev` so behavior is identical; if the whole queue were genuinely behind, the server still transforms (`baseRev < currentRev`), so it stays correct.
- Or read once: take both `.rev` and `.changes` from a single `store.getDoc(docId)` snapshot.

I'd lean toward `max(baseRev)` — it also drops the extra round-trip.

src/client/OTAlgorithm.ts
```diff
@@ -207,6 +208,16 @@ export class OTAlgorithm implements ClientAlgorithm {
 
   // --- Private helpers ---
 
+  /**
+   * Pending OT changes all sit on `committedRev`, so the queue must share `baseRev ===
+   * committedRev` — the consistency the server enforces. Re-stamp any straggler a receive-vs-mint
+   * race left on a stale baseRev; lossless, and the next echo heals the stored copy.
+   */
+  private _withConsistentBaseRev(pending: Change[], committedRev: number): Change[] {
```
matalina, line 258: Heal-on-send, not heal-in-store: the stored straggler stays stale until the next server echo rebases it, and this re-heals on each read until then. That's fine (idempotent, survives restarts) — just worth being explicit. A test asserting the persisted queue is left untouched would document the intent.

tests/client/OTAlgorithm-getPendingToSend.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+  });
+
+  it('returns the queue untouched when every baseRev already matches', async () => {
+    const changes = [
+      createChange(1794, 1795, [{ op: 'add', path: '/a', value: 1 }]),
+      createChange(1794, 1796, [{ op: 'add', path: '/b', value: 2 }]),
+    ];
+    await store.savePendingChanges('doc1', changes);
+
+    const pending = await algorithm.getPendingToSend('doc1');
```
matalina, line 51: Solid coverage. Consider adding an "entire queue stale" case (every change at e.g. 1791, committed at 1794) — that's the one scenario where the anchor choice (`committedRev` vs `max(baseRev)`) actually changes behavior, so a test there would pin the decision.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dabble/patches",
-  "version": "0.12.3",
+  "version": "0.12.4",
   "description": "Immutable JSON Patch implementation based on RFC 6902 supporting operational transformation and last-writer-wins",
```
matalina, line 4: Version bump inside a feature PR — `package.json` 0.12.3→0.12.4, but `package-lock.json` jumps 0.12.2→0.12.4 (pre-existing drift between the two). If releases are tagged/published separately, is bumping the version here intended rather than a release-time step?

matalina (approved):
Approving. Traced the full flow (client mint → rebase → server `commitChanges`) and the core claim holds: the client's local state already applies pending **ops** sequentially ignoring `baseRev` ([OTDoc.ts:132](https://github.com/dabblewriter/patches/blob/fix/pending-baserev-heal/src/client/OTDoc.ts#L132)), and the server anchors on `changes[0].baseRev` and transforms against `listChanges(startAfter: baseRev)`, so re-stamping makes the server converge to exactly what the user already sees. Lossless in the convergence sense — confirmed.

Nothing here blocks merge. Comments below are refinements, not objections; the main one (non-atomic anchor read) is worth a look since it's the same read-skew class as the bug being fixed, but the next echo still heals the stored copy so the exposure is small.

jacwright:
Both items addressed in the rebased branch (now at `fdd3553`):

**1. Atomic read.** `getPendingToSend` now takes a single `getDoc` snapshot, so the pending changes and the `rev` they're based on come from one store transaction. That closes the two-await gap where a server echo could advance `committedRev` against a stale pending snapshot and make us stamp mis-based ops:
```ts
const snapshot = await this.store.getDoc(docId);
if (!snapshot || snapshot.changes.length === 0) return null;
return this._withConsistentBaseRev(snapshot.changes, snapshot.rev);
```

**2. Rebased onto `main`.** Kept both `_withDocLock` (#64) and `_withConsistentBaseRev`. No version conflict after all (main was still 0.12.3), so the 0.12.4 bump applied cleanly.

Non-blocking note folded in too: a comment on `getPendingToSend` documents that the heal touches only the outgoing batch and relies on the next echo to rebase the stored copy.

Full suite green (1984 tests), type/lint/format clean.

PR #68, opened 2026-07-02
fix(sync): recover latched docs, stop error resurrection, thread WS auth params

RyanHavoc:
## TL;DR

When a document's sync fails in a way the client can't immediately recover — a rejected write, a brief server outage, a permission error — it latches at per-doc `syncStatus: 'error'`. The consuming app renders that as a permanent amber "unsynced" indicator, and today **nothing ever re-attempts the doc** unless the user happens to edit it again. This adds a slow background re-probe so latched docs heal on their own, stops a stale-error entry being resurrected after untrack, quiets duplicate error reporting, and threads request params into `canAccess` so server-side role checks work over WebSocket (not just REST).

Part of a three-repo sync-stabilisation effort (pup `fix/role-write-parity`, dabble-writer-3.0 `fix/sync-amber-stabilise`). Builds on the OT frame fixes from #66/#67.

## The fixes

**1. Background re-probe for latched docs.** A doc left at `'error'` with pending changes was only ever retried by a local edit or a full reconnect. So a transient commit-path outage that outlived the retry ladder (~181s), or a terminal 403 that a later server-side policy fix makes valid, stayed amber indefinitely. Now one slow timer per doc re-enters `syncDoc` — 5 min after ladder exhaustion, 10 min after a terminal-code latch — starting a fresh ladder. Cleared on success / untrack / remote-delete / disconnect; a reconnect re-arms only through `syncAllKnownDocs`. `410` is excluded (it diverts to the remote-delete path).

**2. Surface a failure once, not on every probe.** `flushDoc` used to emit `onError` *and* rethrow to `syncDoc`, which emitted again — double-reporting every flush failure, and spamming `onError` even for transient blips `syncDoc` deliberately swallows. `syncDoc` is now the sole emitter, and a persistent terminal error logs/emits once per connection session rather than on every 10-minute probe.

**3. No docStates resurrection after untrack.** `_updateDocSyncState` merged over `EMPTY_DOC_STATE`, so a partial update for an *absent* doc created a full entry. An in-flight sync failing right after `untrackDocs` therefore revived an unreachable `'error'` entry that nothing could clean up until the next reconnect. Split into `_initDocSyncState` (explicit create, at the three sites synchronous with a `trackedDocs` check) vs. a strict update that no-ops when the doc is gone.

**4. WebSocket authorization params.** `JSONRPCServer.assertAccess` never passed request params to `auth.canAccess`, so a provider that validates the payload (e.g. pup's role check reading `params.changes`) saw `undefined` over WS and rejected every collaborator commit — REST was unaffected because it calls the provider with `{changes}` directly. `ApiDefinition` values widen to `Access | { access, params? }`; `commitChanges` declares its param names; the zip threads them through. The `AuthorizationProvider.canAccess` signature already had the optional `params`, so bundled/external providers stay source-compatible.

## Versioning

Minor bump to **0.13.0** — additive api-definition shape and new re-probe behavior, no breaking changes.

## Testing

- New specs (vitest fake timers): 5-min exhaustion probe + recovery, 10-min terminal-403 probe, persistent-403 re-schedule, surface-once-per-session, cleared-on-untrack/disconnect/success, no-probe-without-pending, untrack-race resurrection regression, `canAccess` receives params over the RPC path.
- `npm test` (full): **1997 pass**. `type:check`, `lint`, `format:check`: clean.
- Does not touch the OT frame code from #66/#67 (`rebaseChanges`/`transformIncomingChanges`/`commitChanges` bodies unchanged).

### DIFF
--- a/src/net/PatchesSync.ts
+++ b/src/net/PatchesSync.ts
@@ -68,6 +68,16 @@ const TERMINAL_SYNC_CODES = new Set([401, 402, 403, 404, 410]);
 // connection stays up (see `_scheduleSyncReprobe`).
 const SYNC_REPROBE_EXHAUSTED_MS = 5 * 60_000;
 const SYNC_REPROBE_TERMINAL_MS = 10 * 60_000;
+// Docs that fail together in the same syncAllKnownDocs pass (e.g. a bulk permission
+// revocation) would otherwise all schedule the exact same delay, then re-probe,
+// re-fail, and reschedule in lockstep every 5/10 minutes. Jitter downward only —
+// never later than the nominal delay, only ever equal or earlier — so callers can
+// still rely on the constants above as an upper bound.
+const SYNC_REPROBE_JITTER_RATIO = 0.2;
+
+function jitterReprobeDelay(delayMs: number): number {
+  return delayMs - Math.random() * delayMs * SYNC_REPROBE_JITTER_RATIO;
+}
 
 /**
  * Handles server connection, document subscriptions, and syncing logic between
@@ -606,6 +616,13 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
       throw new Error('Not connected to server');
     }
 
+    // Guarantee a docStates entry exists. flushDoc is protected/subclass-callable and
+    // only checks trackedDocs above, not that _initDocSyncState already ran for this
+    // doc — without this, a flush reached before that init is a silent no-op below
+    // (_updateDocSyncState no-ops for an absent entry), so the doc looks like it never
+    // synced even though the flush succeeded. Merges into any existing entry.
+    this._initDocSyncState(docId, {});
+
     const algorithm = this._getAlgorithm(docId);
 
     try {
@@ -726,6 +743,15 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
   protected async _applyServerChangesToDoc(docId: string, serverChanges: Change[]): Promise<void> {
     const doc = this.patches.getOpenDoc(docId);
     const algorithm = this._getAlgorithm(docId);
+
+    // Guarantee a docStates entry exists for a tracked doc. This is reachable from
+    // paths that don't run _initDocSyncState first — the public applyMergeChanges API
+    // and the raw onChangesCommitted push both call this directly — so without this,
+    // the committedRev update below silently no-ops for a doc whose entry hasn't been
+    // created yet. Gated on trackedDocs so a genuinely-untracked doc still gets no entry.
+    if (this.trackedDocs.has(docId)) {
+      this._initDocSyncState(docId, {});
+    }
 
     // Delegate to algorithm - it handles store updates and doc updates
     await algorithm.applyServerChanges(docId, serverChanges, doc);
@@ -1055,7 +1081,7 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
    */
   protected _scheduleSyncReprobe(docId: string, retryable: boolean): void {
     if (!this._isConnectedAndOnline() || !this.trackedDocs.has(docId)) return;
-    const delay = retryable ? SYNC_REPROBE_EXHAUSTED_MS : SYNC_REPROBE_TERMINAL_MS;
+    const delay = jitterReprobeDelay(retryable ? SYNC_REPROBE_EXHAUSTED_MS : SYNC_REPROBE_TERMINAL_MS);
     const existing = this._syncReprobeTimers.get(docId);
     if (existing !== undefined) globalThis.clearTimeout(existing);
     const timer = globalThis.setTimeout(() => {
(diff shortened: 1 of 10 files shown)


matalina (commented):
Reviewed the diff (8-angle pass: correctness, cross-file, reuse, simplification, efficiency, altitude, conventions). Overall this is solid and well-tested — the untrack-resurrection fix and error-surfacing dedupe are correctly implemented and match their specs. Left a few inline notes, most severe first. No CLAUDE.md violations found.

src/net/PatchesSync.ts
```diff
@@ -652,8 +677,10 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
       }
       const flushError = err instanceof Error ? err : new Error(String(err));
       this._updateDocSyncState(docId, { syncStatus: 'error', syncError: flushError });
```
matalina, line 696: **Correctness (medium confidence):** the new `_updateDocSyncState` no-ops when the doc has no existing docStates entry (see the guard at `_updateDocSyncState`, now requiring `docId in currentDocs`). `flushDoc` is `protected` and only guards on `trackedDocs.has(docId)`, not on a docStates entry existing — this codebase's own test suite calls `sync['flushDoc']('doc1')` directly with just `trackedDocs.add('doc1')` (no `_initDocSyncState` first) at `tests/net/PatchesSync.spec.ts:882-897`. In that shape the flush succeeds against the server but this status update (and the `hasPending`/`syncStatus: 'synced'` update a few lines up) becomes an inert no-op — `docStates.state[docId]` stays `undefined` even though sync happened. Same gap applies to `_applyServerChangesToDoc` (line ~735) when reached via the public `applyMergeChanges` API or the raw `onChangesCommitted` push, for a doc whose docStates entry hasn't been created yet. Worth either having `flushDoc`/`_applyServerChangesToDoc` call `_initDocSyncState` defensively, or documenting the precondition as loudly as `_initDocSyncState`'s docstring does.

src/net/PatchesSync.ts
```diff
@@ -991,18 +1040,53 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
...
+   * re-probe sooner than definitive rejections, which only heal via a server-side
+   * policy change. One timer per doc; cleared alongside the fast retry state on
+   * success, untrack, delete, and true disconnect (a reconnect re-arms only through
+   * `syncAllKnownDocs` re-attempting the doc — never from here). A transient
+   * `connecting` flap leaves the timer pending; it self-bails at fire time when it
+   * finds the connection down rather than being cleared up front.
+   */
+  protected _scheduleSyncReprobe(docId: string, retryable: boolean): void {
+    if (!this._isConnectedAndOnline() || !this.trackedDocs.has(docId)) return;
+    const delay = retryable ? SYNC_REPROBE_EXHAUSTED_MS : SYNC_REPROBE_TERMINAL_MS;
```
matalina: **Efficiency:** the reprobe delay is fixed (`SYNC_REPROBE_EXHAUSTED_MS`/`SYNC_REPROBE_TERMINAL_MS`), no jitter. A batch of docs that fail together in the same `syncAllKnownDocs` pass (e.g. a bulk permission revocation) will schedule identical delays, so every re-probe fires in the same tick, re-fails together, and reschedules together — a synchronized burst against the server every 5/10 minutes for as long as the condition persists. Might be worth adding +/-10-20% jitter to spread the load.

src/net/PatchesSync.ts
```diff
@@ -155,6 +159,16 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
...
+  /** Pending per-doc slow re-probe timers for docs latched at 'error' (see `_scheduleSyncReprobe`). */
+  private _syncReprobeTimers = new Map<string, ReturnType<typeof globalThis.setTimeout>>();
+  /**
+   * Docs whose current sync failure has already been surfaced (console + `onError`).
+   * A background re-probe re-enters `syncDoc` every few minutes; without this a
+   * permanently-latched doc (e.g. a 403) would re-log/re-emit the identical error on
+   * every probe. Cleared on recovery, untrack, remote delete, and disconnect — so a
+   * still-failing doc re-surfaces at most once per connection session.
+   */
+  private _surfacedSyncErrors = new Set<string>();
```
matalina, line 181: **Altitude:** this is now the third parallel per-doc state track (alongside `_syncRetryTimers`/`_syncRetryAttempts` and the new `_syncReprobeTimers`) that has to be hand-cleared at the same four exit points (success, untrack, remote-delete, disconnect). `_clearSyncRetry`/`_clearAllSyncRetries` already do this manually for all three today. Folding `errorSurfaced` into `DocSyncState` itself (riding along with the doc's existing single source of truth) would remove one of these side-channel collections and the corresponding manual-clear checklist.

src/net/protocol/JSONRPCServer.ts
```diff
@@ -98,7 +122,7 @@ export class JSONRPCServer {
           );
         }
         const ctx = getAuthContext();
-        await this.assertAccess(access, ctx, method, args);
+        await this.assertAccess(access, ctx, method, args, buildNamedParams(paramNames, args));
```
matalina: **Minor efficiency:** `buildNamedParams(paramNames, args)` is evaluated eagerly as an argument expression on every call to a method with declared param names, even when `this.auth` is undefined (in which case `assertAccess` immediately returns without using `params`). For a server with no `AuthorizationProvider` configured, every `commitChanges` call still builds and discards an object. Could gate the build on `this.auth` being set, or move it inside `assertAccess` after the `!this.auth` check.

matalina (approved):

PR #71, opened 2026-07-02
Deep review: fix 66 verified sync, transform, and transport bugs

jacwright:
## TL;DR

A full audit of the sync engine found and fixed 66 verified bugs. The headline fixes stop real user-facing data loss: concurrent edits could silently delete other people's committed changes, branch merges could drop every edit made on a branch, offline resends could duplicate content, undo could corrupt documents, and several reconnect races could leave a client stale forever while showing "synced." Every fix ships with a regression test; nothing about the architecture changed.

## How these were found

Every subsystem was reviewed by an independent auditor, every finding was adversarially verified against the code (refute-first, then reproduce — most confirmed findings have runnable repros), and every fix's regression test was checked to fail before the fix. Suite grew from 1,984 to 2,128 tests, all green, with type/lint/format clean.

## Fix areas

**JSON Patch core** (`fix(json-patch)` ×2)
- `transformPatch` now threads coordinate shifts through sequential ops (array indexes, `@txt` deltas, moves) instead of transforming each op against fixed coordinates
- `/arr/-` appends no longer corrupt to `/arr/NaN` under concurrent array edits
- `invertPatch` inverts against the evolving state (multi-op undo was wrong), handles root paths, `~0`/`~1` escapes, numeric object keys, and end-of-array appends
- `copy` no longer aliases source/destination; `composePatch` no longer reorders writes across a child edit

**OT** (`fix(ot)`, `fix(ot-client)`, part of merge with #66/#67)
- Server and client rebase walks advance concurrent ops between sequential changes (the same frame bug #66 fixed; this branch extends it to commit dedup seeing same-batch resends, echoing committed copies of deduped resends, fast-forward rev renumbering from the authoritative tip, and a whole-batch root-wipe guard)
- Flush-time change splits are persisted back before sending, so echoes clear what was actually sent instead of orphaning and duplicating the original

**Client core** (`fix(client)`)
- Optimistic ops are rebased when server changes land between `change()` and its queued mint — previously misplaced ops were committed verbatim
- Lifecycle races closed: closeDoc drains the change queue, deleteDoc awaits in-flight opens, optimistic rollback is scoped to the failed change

**Sync** (`fix(sync)`)
- Reconnect no longer evicts docs tracked during the resync window (their edits silently stopped sending until next reconnect)
- Broadcasts racing a doc's initial load are buffered and reconciled instead of dropped-while-reporting-synced
- Edits minted mid-load can no longer be re-stamped past the server's baseRev-0 root-replace guard

**LWW** (`fix(lww)` ×2)
- Server echoes merged delta results so a client whose `@inc` combined with an unseen write converges; catchup sorts by commit order instead of client timestamps; catchup floor off-by-one fixed
- Server commits are serialized per doc; in-flight sends are protected from older concurrent broadcasts; delta combining keeps the winning timestamp
- Client stores keep full committed op fidelity (fresh clients no longer lose uncompacted fields) and a per-doc lock closes a flush race that wiped concurrent edits

**Branching** (`fix(server)`)
- LWW branch merges no longer silently drop all branch edits (rev-space confusion); OT branch cold-loads no longer drop the earliest edits; repeat merges are idempotent and keep the source's version watermark sound; client-supplied merge watermarks are stripped

**Client stores** (`fix(client-stores)`)
- Re-delivered committed changes dedup instead of double-applying on rebuild; committedRev watermarks can't advance past the server head; IndexedDB upgrades fire when an algorithm's stores are missing

**Transport** (`fix(net)`)
- JSON-RPC errors rehydrate into `StatusError` over WebSocket, restoring 410/terminal-error handling in PatchesSync
- Stale socket/SSE handler races and FetchTransport error handling fixed

**WebRTC** (`fix(webrtc)`)
- Awareness broadcasts no longer silently dropped (id read lazily after connect); peer states replace instead of merge; offer glare gets a deterministic tie-break; disconnect→connect resubscribes signaling

## Known follow-ups (deliberately not in this PR)

- `src/micro/`: 12 verified findings including 3 critical — being specced as its own piece of work
- LWW change-id idempotency mechanism, server timestamp clamping, CompressedStoreBackend × versions composition — designs/fixes coming in a follow-up branch

### DIFF
--- a/src/client/LWWAlgorithm.ts
+++ b/src/client/LWWAlgorithm.ts
@@ -26,6 +26,9 @@ export class LWWAlgorithm implements ClientAlgorithm {
   readonly name = 'lww';
   readonly store: LWWClientStore;
 
+  /** Per-doc FIFO mutex (see {@link _withDocLock}). */
+  private readonly _docLocks = new Map<string, Promise<unknown>>();
+
   constructor(store: LWWClientStore) {
     this.store = store;
   }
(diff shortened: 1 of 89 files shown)


src/net/PatchesSync.ts
```diff
@@ -612,42 +641,90 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
...
+    flushPendingFirst = false,
+    resolvedChanges: Change[] = []
+  ): Promise<void> {
+    this._reloadBuffers.set(docId, []);
+    try {
+      // Read the rev the local committed state (and thus pending) sits on BEFORE overwriting it.
+      const baseRev = await algorithm.getCommittedRev(docId);
+      const snapshot = await this.connection.getDoc(docId);
+      if (flushPendingFirst && snapshot.rev > 0 && (await algorithm.hasPending(docId))) {
+        await this.flushDoc(docId);
```
RyanHavoc, line 673: The `flushPendingFirst` early return skips the buffer-reconciliation loop below, but the `finally` still deletes `_reloadBuffers` — so a change committed by another client during this `flushDoc` gets parked into the buffer by `_receiveCommittedChanges` and then destroyed unapplied. The doc silently misses that committed change until a later resync.

Repro shape: reload with pending (new-doc hydration path) → foreign commit lands mid-flush → parked at the buffer check in `_receiveCommittedChanges` → early return here → `finally` drops it.

Suggest running the reconciliation loop (or re-dispatching the buffered changes) on this exit path too, and auditing the other early exits between the buffer being set and reconciled: https://github.com/dabblewriter/patches/blob/065d51a1102eb93f21f678c55b156f99613e60dd/src/net/PatchesSync.ts#L666-L675

src/client/OTAlgorithm.ts
```diff
@@ -178,6 +182,18 @@ export class OTAlgorithm implements ClientAlgorithm {
     // Pending changes remain until server commits them back.
   }
 
+  async replacePendingChanges(docId: string, oldChanges: Change[], newChanges: Change[]): Promise<void> {
+    return this._withDocLock(docId, async () => {
+      // Preserve any changes minted after oldChanges was read, renumbered after the new queue
+      const oldIds = new Set(oldChanges.map(c => c.id));
+      const current = await this.store.getPendingChanges(docId);
+      let rev = newChanges[newChanges.length - 1].rev;
```
RyanHavoc: `newChanges` can be empty here, making this throw `TypeError: Cannot read properties of undefined (reading 'rev')` mid-flush. Reachable path: a pending change whose oversized `@txt` ops contain no insert operations makes `breakTextOp` return `[]`, which collapses the change in `breakChanges`; `flushDoc` then flattens the batches to `[]` and calls this with it: https://github.com/dabblewriter/patches/blob/065d51a1102eb93f21f678c55b156f99613e60dd/src/net/PatchesSync.ts#L796-L800

Suggest an early return when `newChanges` is empty (deciding explicitly whether the old pending should be dropped or kept in that case).

src/net/PatchesSync.ts
```diff
@@ -1007,9 +1101,12 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
     // Snapshot current subscriptions before removing docs
     const subscribedBefore = this._getActiveSubscriptions();
 
-    existingIds.forEach(id => this.trackedDocs.delete(id));
-    existingIds.forEach(id => this._clearSyncRetry(id));
-    existingIds.forEach(id => this._surfacedSyncErrors.delete(id));
+    existingIds.forEach(id => {
+      this.trackedDocs.delete(id);
+      this._untrackedDuringResync?.add(id);
```
RyanHavoc, line 1144: Nothing ever removes an id from `_untrackedDuringResync` — `_handleDocsTracked` doesn't delete from it. If a doc is untracked and then re-tracked while the same `syncAllKnownDocs` pass is still in flight, the merge at the end still excludes it (https://github.com/dabblewriter/patches/blob/065d51a1102eb93f21f678c55b156f99613e60dd/src/net/PatchesSync.ts#L450-L460), so the re-tracked doc silently drops out of `trackedDocs`/`docStates` and stops syncing until the next reconnect.

Suggest `this._untrackedDuringResync?.delete(id)` in the tracked handler.

src/client/LWWAlgorithm.ts
```diff
@@ -50,51 +53,55 @@ export class LWWAlgorithm implements ClientAlgorithm {
...
-      (doc as LWWDoc<T>).applyChanges(changes, true, {
-        inFlightOpsOverride: opsToSave,
-        retiredInFlightOps,
-      });
-    }
-
-    return changes;
+    // Serialize per-doc so getPendingToSend's read-build-clear can't interleave with this
+    // read-consolidate-save and wipe an op saved between its read and its clear.
+    return this._withDocLock(docId, async () => {
```
RyanHavoc, line 58: Test-coverage note (minor): the doc lock added here fixes the mint-vs-flush op-loss race (concurrent `handleDocChange` + `getPendingToSend` both reading stale pending), but no test actually races the two — e.g. firing both via `Promise.all` and asserting no op is lost. It's the one fix in this PR whose regression test appears to be missing; without it a future refactor could drop the lock without anything going red.

RyanHavoc:
Pushed fixes for the four review findings (065d51a..8753915), one commit each:

- 7cc7aa9 — broadcasts parked during a flush-pending-first reload are now drained instead of destroyed (reconciliation loop extracted into _drainReloadBuffer and run on both exit paths; the flushDoc-throws path deliberately skips it since syncDoc's retry/catch-up recovers those — documented inline).
- abc054a — replacePendingChanges no longer crashes when the queue re-splits to nothing: old pending is cleared, mintedSince ops survive (renumbered off getCommittedRev), and flushDoc short-circuits the empty-flattened case reporting the store's real hasPending.
- 7b9c481 — _handleDocsTracked clears a doc's untracked-during-resync mark, so untrack-then-retrack within one resync no longer drops the doc from sync.
- 8753915 — deterministic race test for the LWW per-doc lock (handleDocChange vs getPendingToSend, controlled deferreds).

Full suite 2174 passed (7 new tests); the five regression tests for the three code fixes verified red against the unfixed code. type/lint/format clean. Fast-forward push — no history rewritten.

PR #72, opened 2026-07-03
fix(sync): don't latch docs as errored when a request/transaction aborts

RyanHavoc:
## TL;DR

When Dabble is quit (or reloaded for an update) mid-sync, the browser kills the in-flight requests with an `AbortError`. The sync engine treated that like a real per-document failure — marking the doc "errored", which lights the amber unsynced indicator and fires `sync_doc_error_latched` telemetry in the app. Nothing was actually wrong: the user's changes are safe on disk and sync fine on the next launch. Aborts are now treated as interruptions to recover from, never as document failures.

## Diagnosis (Sentry DABBLE-WRITER-3-7W / 7X / 7Z, beta)

Three `sync doc error latched: … (20)` issues, all from the Electron desktop app. The "(20)" is **not an HTTP status** — it is `DOMException.ABORT_ERR` (legacy `code` 20) from an **aborted fetch** (`error_message: "The user aborted a request."`), the classic signature of page/worker teardown killing in-flight requests. Each affected session latched several docs in the same second — a mass-abort burst at app quit/reload.

The latch was not the terminal-code classification (`TERMINAL_SYNC_CODES` only matches `StatusError` 401/402/403/404/410): `syncDoc`'s catch set `syncStatus: 'error'` for **every** failure before deciding retryability, so even a scheduled-for-retry abort painted the doc — and at teardown the retry never runs, so the error state is the doc's last word.

## Changes

- `isAbortError()` helper in `net/error.ts` — name-based (`AbortError`), so it classifies DOMExceptions from both aborted fetches and aborted IndexedDB transactions, including copies that crossed a structured-clone/worker boundary.
- `syncDoc`: an abort no longer latches `'error'`. The doc returns to its stable status (same rule as the disconnect reset — `'synced'` when local data/pending exists, `'unsynced'` for a brand-new empty doc) and recovery rides the existing machinery: backed-off retry ladder while connected, slow reprobe once the ladder is exhausted with pending changes, otherwise the reconnect's `syncAllKnownDocs`.
- `flushDoc`: skips the momentary `'error'` docStates transition on abort (consumers broadcast every transition, so even a transient one fired telemetry).
- Aborts that persist past the ladder while connected+online (real environment trouble: storage pressure, wedged network stack) surface **once** via `onError`, mirroring the exhausted-transient contract — still without latching the doc.

Pending changes are never dropped on an aborted flush: pending only clears via `confirmSent` after a successful commit, and the tests pin that.

## Tests

- 6 new `PatchesSync` specs: aborted pull recovers via ladder; aborted flush never paints `'error'` into the docStates stream and preserves pending; aborted IndexedDB read handled the same; brand-new doc stays `'unsynced'`; teardown-coincident abort stays quiet with nothing armed; persistent aborts are bounded, surface once, keep reprobing, never latch.
- 4 new `isAbortError` specs.
- Full suite: 2184 passed (94 files). Type check, lint, format check all clean.

### DIFF
--- a/src/net/PatchesSync.ts
+++ b/src/net/PatchesSync.ts
@@ -11,7 +11,7 @@ import type { AlgorithmName, TrackedDoc } from '../client/PatchesStore.js';
 import { isDocLoaded } from '../shared/utils.js';
 import type { Change, DocSyncState, DocSyncStatus, PatchesSnapshot } from '../types.js';
 import { blockable, serialGate } from '../utils/concurrency.js';
-import { ErrorCodes, StatusError } from './error.js';
+import { ErrorCodes, isAbortError, StatusError } from './error.js';
 import type { PatchesConnection } from './PatchesConnection.js';
 import type { JSONRPCClient } from './protocol/JSONRPCClient.js';
 import type { BranchAPI, ConnectionState } from './protocol/types.js';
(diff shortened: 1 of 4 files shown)


RyanHavoc:
Superseded by #73, which incorporates this PR's abort classification (isAbortError, the pending-aware stable-status rule, surface-once-on-persistent-abort) and its full test suite verbatim, and extends the same treatment to all status-less network failures at the transport layer. Closing to avoid a conflicting double-merge — nothing here is lost.

PR #96, opened 2026-07-06
feat(net): add RelayTransport for server-relayed awareness

matalina:
**TL;DR:** Adds `RelayTransport` — a server-relayed peer transport — so `WebRTCAwareness` can run over the existing signaling protocol with no WebRTC, no simple-peer, and no NAT/ICE failure class. This is the patches-side groundwork for presence in Dabble (who's online in a project/branch), which will relay awareness states through pup over the doc-sync SSE stream. `WebRTCAwareness` itself is unchanged apart from accepting any `AwarenessTransport` instead of being hard-typed to `WebRTCTransport`.

## What's here

- **`AwarenessTransport`** (`net/protocol/types.ts`): the transport surface `WebRTCAwareness` actually consumes (`id`, `connect`, `disconnect`, `onPeerConnect`, `onPeerDisconnect`, `onMessage`, `send`). `WebRTCTransport` satisfies it as-is; the awareness constructor is widened to the interface with no behavior change.
- **`RelayTransport`** (`net/signaling/RelayTransport.ts`): satisfies `AwarenessTransport` over any `SignalingTransport` (e.g. `PatchesRESTSignalingTransport`, `WebSocketTransport`) by mapping the existing wire protocol:
  - `peer-welcome` rosters are **authoritative** — repeat welcomes (signaling reconnects) retire missing peers and re-announce present ones, so consumers re-push their full state and reconnects self-heal.
  - A `signal` from an unknown sender is announced via `onPeerConnect` before delivery — late joiners introduce themselves with their first frame, and the announcement prompts the receiving side to push its state back, completing the two-way exchange the WebRTC handshake would otherwise provide.
  - Optional **room tag**: constructed with a `roomId`, it filters inbound frames to `params.room` and stamps outbound `peer-signal` frames with the room as a third param — several rooms can share one signaling connection (needed because one SSE clientId serves all tabs). Fully additive: stock `SignalingService` ignores the extra param, and an untagged `RelayTransport` behaves exactly like the stock protocol.
- **Exports:** `RelayTransport`, `AwarenessTransport`, and `WebRTCAwareness` are available from `@dabble/patches/net`, so relay consumers never import the simple-peer-backed `/webrtc` barrel. `/webrtc` exports are unchanged.

## Why relay instead of P2P

For presence-sized payloads, relaying through the server removes the symmetric-NAT failure class entirely (no TURN to deploy), and `WebRTCTransport` currently exposes no ICE/TURN configuration anyway. The signaling protocol is a superset of what P2P needs, so a future P2P mode can layer on without wire changes.

## Tests

18 new tests in `tests/net/signaling/RelayTransport.spec.ts`, including an end-to-end that runs two `WebRTCAwareness` instances against the real `SignalingService` and verifies both-way state exchange, full-state replacement on update, and state removal on disconnect. `type:check`, `lint`, and `format:check` are clean. The three pre-existing `tests/solid/*` suite-load failures (`file:///@solid-refresh` TypeError) fail identically on untouched `main` and are unrelated.

## Consumers

- **pup** needs nothing from this PR — its installed patches already ships `SignalingService`, whose overridable surface is enough for the room layer (rooms, Redis fan-out, uid stamping live in pup).
- **dw3** consumes `RelayTransport` + the widened awareness from the release this cuts.

### DIFF
--- a/src/net/signaling/RelayTransport.ts
+++ b/src/net/signaling/RelayTransport.ts
@@ -118,7 +118,9 @@ export class RelayTransport implements AwarenessTransport {
       this.rpc.on('peer-welcome', (params: { id: string; peers: string[]; room?: string }) => {
         if (!this._inRoom(params)) return;
         this._id = params.id;
-        const roster = new Set(params.peers);
+        // Stock servers already exclude self from the roster; filter defensively so a server
+        // that doesn't can never induce a phantom self-peer (and a wasted self-send)
+        const roster = new Set(params.peers.filter(peerId => peerId !== params.id));
         // The roster is authoritative: peers we know that the server no longer lists are gone
         // (their disconnect broadcast may have been lost while our stream was down)
         for (const peerId of Array.from(this.peers)) {
--- a/src/net/signaling/RelayTransport.ts
+++ b/src/net/signaling/RelayTransport.ts
@@ -1,0 +1,161 @@
+import { signal, type Unsubscriber } from 'easy-signal';
+import { JSONRPCClient } from '../protocol/JSONRPCClient.js';
+import type { AwarenessTransport, ConnectionState, SignalingTransport } from '../protocol/types.js';
+
+/**
+ * Server-relayed peer transport: satisfies {@link AwarenessTransport} using nothing but the
+ * {@link SignalingService} wire protocol, so awareness state flows through the signaling server
+ * instead of WebRTC data channels. No P2P, no NAT/ICE failure class, no simple-peer — the
+ * signaling channel (e.g. `PatchesRESTSignalingTransport` over the doc-sync SSE stream) is the
+ * whole transport.
+ *
+ * Protocol mapping (server messages → peer events):
+ * - `peer-welcome {id, peers}` — adopts `id`, treats `peers` as the authoritative roster:
+ *   every listed peer is (re-)announced via `onPeerConnect` (consumers re-push their full
+ *   state, which makes signaling reconnects self-healing), and known peers missing from the
+ *   roster are retired via `onPeerDisconnect`.
+ * - `signal {from, data}` — delivers `data` via `onMessage`. A `from` we have not seen yet is
+ *   announced via `onPeerConnect` first: peers that join after our welcome introduce themselves
+ *   with their first frame, and the announcement prompts consumers to push our state back —
+ *   completing the two-way exchange the WebRTC handshake would otherwise provide.
+ * - `peer-disconnected {id}` — retires the peer via `onPeerDisconnect`.
+ *
+ * Sending fans out as directed `peer-signal [to, data]` notifications — one per known peer when
+ * `peerId` is omitted (the protocol has no broadcast).
+ *
+ * **Rooms.** Servers that host multiple peer groups on one connection can tag the three
+ * messages above with a `room` param (and read the third `peer-signal` param, which stock
+ * {@link SignalingService} ignores). Construct with that `roomId` to make this transport filter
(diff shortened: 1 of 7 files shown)


src/net/signaling/RelayTransport.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  /**
+   * Sends `data` to one peer, or to every known peer when `peerId` is omitted, as directed
+   * `peer-signal` notifications through the server.
+   */
+  send(data: string, peerId?: string): void {
+    const targets = peerId ? [peerId] : Array.from(this.peers);
+    for (const to of targets) {
+      // notify() drops a trailing undefined room, keeping stock frames two-element
+      this.rpc.notify('peer-signal', to, data, this.roomId);
```
RyanHavoc, line 110: Non-blocking heads-up for the pup follow-up: a room-tagged `RelayTransport` can't round-trip against the stock `SignalingService`. The relay rebuilds the outbound frame as `signal {from, data}` and drops the third `room` param (https://github.com/dabblewriter/patches/blob/18774af140d178d1c407da1d54b9676e3a3bb259/src/net/signaling/SignalingService.ts#L129-L136), so a room-aware receiver filters it right back out. That's correct per the CLAUDE.md scope boundary (room routing is pup's job), and the untagged path is fully e2e-tested — just flagging that the room round-trip isn't proven in-repo until pup echoes `room` on `signal` / `peer-welcome` / `peer-disconnected`.

src/net/signaling/RelayTransport.ts
```diff
@@ -0,0 +1,161 @@
...
+          if (!roster.has(peerId)) {
+            this.peers.delete(peerId);
+            this.onPeerDisconnect.emit(peerId);
+          }
+        }
+        // (Re-)announce every listed peer — even already-known ones, so consumers re-push
+        // state to peers that dropped us while our stream was down
+        for (const peerId of roster) {
+          this.peers.add(peerId);
+          this.onPeerConnect.emit(peerId);
```
RyanHavoc, line 136: Minor robustness note: this adds every roster entry as a peer without excluding our own `params.id`. The stock server already filters self out (`onClientConnected`, `pid !== id`: https://github.com/dabblewriter/patches/blob/18774af140d178d1c407da1d54b9676e3a3bb259/src/net/signaling/SignalingService.ts#L66-L68), so it never fires today, and the `from === this._id` guard in the `signal` handler prevents any echo loop — worst case is a phantom self-peer plus one wasted self-frame. `WebRTCTransport` has the same non-guard, so this is consistent, not a regression. Worth preserving the self-filter when pup reimplements roster construction for rooms.

RyanHavoc (approved):

src/net/signaling/RelayTransport.ts
matalina: Good call — the tagged path is now proven in-repo (5fc02e7): added an end-to-end test against a room-echoing `SignalingService` subclass (the shape pup implements) — two room-tagged `RelayTransport`s exchange awareness states both ways, the C→S third param is asserted server-side, and a transport tagged for another room sharing the same channel stays silent. The production server half echoes `room` on all three messages in dabblewriter/pup#145.

src/net/signaling/RelayTransport.ts
matalina: Added the defensive self-filter plus a test in 5fc02e7 — a roster that includes our own id can no longer induce a phantom self-peer or a wasted self-send. And noted for the server side: pup's roster construction in dabblewriter/pup#145 builds `peers` excluding the joiner, preserving the stock filter.

PR #98, opened 2026-07-07
DAB-607: Client re-sends an already-acked change with advanced baseRev; server commits a duplicate

jacwright:
## TL;DR

A rare sync race could make the app save the same edit twice, which doubled the affected text's word-count stats (and for a plain insert, could double the text itself). This teaches the client to recognize an edit the server has already accepted and never send it again.

Prod evidence: project `SjBEvVIbMjF9i6L5` revs 28795/28796 (2026-07-07 02:18 UTC) — the same change id committed twice, ~130ms after the client received the ack for the first commit.

## Root cause

The same commit can be delivered to a client more than once: the SSE broadcast and the HTTP ack both carry the committed change, plus re-broadcasts and catchup overlaps. The first delivery advances `committedRev` and confirms the pending copy by id. But `applyCommittedChanges` filtered redundant deliveries out by rev **before** the id-based pending drop, so a pending copy stranded by a raced pending write (e.g. a stale write landing between the broadcast apply and the ack apply) was never confirmed away.

The next flush then re-sent the stranded change with `baseRev` advanced **past its own first commit**. Server-side idempotency in `commitChanges` only checks `listChanges({ startAfter: baseRev })`, so the original commit is invisible and it commits a duplicate. Replace-shaped ops happen to be idempotent under double-apply (this prod instance was text-neutral), but a duplicated plain insert doubles typed words.

## Fix (client-side, three layers)

A "strand" = a pending copy of an already-committed change. Detection has to survive `rebaseChanges` re-stamping every surviving pending change to a clean `baseRev` on each foreign receive, so id identity is the only reliable signal:

1. **`applyCommittedChanges`**: pending copies of *redundantly delivered* committed changes (rev <= snapshot rev) are confirmed/dropped by id. Echoes among the *new* server changes keep their existing in-walk handling in `rebaseChanges` (interleaved foreign frames need them).

2. **Per-doc committed-id record (`OTAlgorithm`)**: applied commit ids are remembered in memory (bounded, noted only **after** the store write resolves so a remembered id can never be a live echo still owed to the rebase walk). Consulted at receive — a remembered id in the pending merge is stripped before the rebase, so it can't mis-advance foreign ops — and at send, where it catches strands regardless of baseRev shape, including ones a later rebase re-stamped clean.

3. **Send choke point (`getPendingToSend`)**: a stale-`baseRev` queue (the wire signature of the prod duplicate, and the shape a fresh session inherits) additionally checks the recent committed tail in the store; committed copies are matched via server-stamped `committedAt` so a torn pending row can't self-confirm. Strands are dropped from the store and the outgoing batch, and their ids are remembered so an open doc's in-memory copy can't resurrect them.

Also: `PatchesSync.syncDoc` now refreshes `hasPending` when a sync finds nothing to flush (a strand drop could otherwise leave a stale unsaved-changes indicator), `OTInMemoryStore` gains `listChanges` (parity with `OTIndexedDBStore`), and `.gitignore` covers a symlinked `node_modules`.

## Reviewed

Multi-agent review (10 finder angles + adversarial verification + sweep) ran against the initial version and its confirmed findings are fixed in the second commit: the rebase re-stamp bypass of the send-side guard, foreign-only deliveries polluting the rebase frame, doc-side strand resurrection, the stuck `hasPending` flag, plus reuse/simplification cleanups.

Known limitations (accepted): the store-scan detector is bounded to the last 200 revs (a cross-session strand older than that can still slip through), and id matching carries the same theoretical collision exposure as the existing `rebaseChanges` id-drop.

## Not in this PR

The optional server-side guard from DAB-607 (an id-indexed lookup below `baseRev` in `commitChanges`) is deferred: it needs an `OTStoreBackend` interface addition across backends plus a collision review of the 4-random-char change ids. It remains the eventual backstop for the bounded-window limitation above.

## Testing

- New unit tests for all three layers, including the rebase-re-stamped strand and foreign-only-delivery scenarios; the defect tests fail without the src changes.
- Full suite: 2454 passed, 4 skipped (pre-existing). Type, lint, and format checks clean.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 .nyc_output
 .obsidian
 npm-debug.log
-node_modules/
+node_modules
 coverage/
 dist/
 .svelte-kit/
--- a/src/algorithms/ot/client/applyCommittedChanges.ts
+++ b/src/algorithms/ot/client/applyCommittedChanges.ts
@@ -53,20 +53,24 @@ export function applyCommittedChanges(
 ): PatchesSnapshot {
   let { state, rev, changes } = snapshot;
 
-  // Filter out any server changes that are already reflected in the current snapshot's revision.
-  // Server changes should always have a rev.
-  const newServerChanges = committedChangesFromServer.filter(change => change.rev > rev);
+  // Split delivered changes into new ones and ones already reflected in the snapshot's
+  // revision. Server changes should always have a rev.
+  const newServerChanges: Change[] = [];
+  const staleServerChanges: Change[] = [];
+  for (const change of committedChangesFromServer) {
+    (change.rev > rev ? newServerChanges : staleServerChanges).push(change);
+  }
 
   // A commit can be delivered more than once (SSE broadcast + HTTP ack, re-broadcast, catchup
   // overlapping a broadcast). The first delivery advances `rev` and confirms the pending copy
-  // by id; a redundant delivery is filtered out above and used to skip that confirmation, so a
-  // pending copy stranded by a raced pending write was never cleared — the next flush re-sent
-  // it with a baseRev past its own commit, outside the server's `startAfter: baseRev` id dedup,
-  // committing a duplicate (DAB-607). Confirm those strands here. Echoes in `newServerChanges`
-  // are NOT pre-dropped: rebaseChanges removes them in-walk, where foreign changes interleaved
-  // before an echo must advance through it to meet the queue's tail in the right frame.
-  if (changes && changes.length > 0 && newServerChanges.length < committedChangesFromServer.length) {
-    const staleIds = new Set(committedChangesFromServer.filter(change => change.rev <= rev).map(change => change.id));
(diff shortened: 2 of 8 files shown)


RyanHavoc:
Pushed a small follow-up (4d96cbc): the no-flush branch in `syncDoc` now re-reads `algorithm.hasPending(docId)` instead of writing `hasPending: false` unconditionally.

The `false` write correctly clears a stale-true left by a dropped strand, but it can also clobber a `hasPending: true` set by a mint that lands concurrently during that branch — briefly showing "no unsaved changes" until the queued follow-up sync corrects it. Re-reading the store's real state keeps the strand-drop fix while respecting a concurrent mint, and matches the `stillHasPending = await algorithm.hasPending(docId)` idiom already used in `flushDoc`/`_reloadDocFromServer`.

Type-check, format, and the full `tests/net/PatchesSync*` suite (188 tests) pass. Rest of the review was clean — no correctness bugs found; the bounded-window and id-collision considerations are already documented as accepted limitations.

PR #100, opened 2026-07-08
fix(net): add dispose to JSONRPCClient and RelayTransport

matalina:
## What

Adds `dispose()` to `JSONRPCClient` and `RelayTransport`.

`JSONRPCClient`'s constructor subscribes to the transport's `onMessage` and `onStateChange` but discarded both unsubscribers and offered no way to release them. Any consumer creating short-lived clients over a **shared, long-lived transport** leaked two handlers per client: every inbound frame was re-parsed by each dead client, and each dead client's whole object graph stayed pinned in memory.

`RelayTransport` hits this exactly — it builds one `JSONRPCClient` per instance over a shared signaling channel, so a host creating one transport per room (e.g. dabble-writer's presence rooms multiplexed over one SSE connection) leaks without bound in its long-lived hub worker. dabble-writer currently works around this by pooling and reusing transports per family; once this ships, that pool becomes optional.

## How

- `JSONRPCClient` captures its transport subscriptions and exposes `dispose()`: releases the handlers, clears notification signals, and rejects pending calls with a `NetworkError`. The client must not be used afterward.
- `RelayTransport.dispose()` = `disconnect()` + `rpc.dispose()` for **final** teardown. `disconnect()` deliberately keeps the transport-level subscription so `connect()` can reuse the instance — the docs on both methods now spell out the difference.

## Testing

- `dispose()` releases the message + state-change subscriptions and rejects pending calls with `NetworkError`; notification subscribers stop firing after dispose.
- A `handlerCount` probe on the mock signaling transport proves N create/dispose cycles leave the shared channel clean, and that `disconnect()` alone (the reuse path) keeps its subscription as intended.
- `type:check`, lint, format clean; full suite green apart from the three pre-existing `tests/solid/*` suite-load failures that fail identically on untouched main.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@dabble/patches",
-  "version": "0.17.1",
+  "version": "0.19.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@dabble/patches",
-      "version": "0.17.1",
+      "version": "0.19.0",
       "license": "MIT",
       "dependencies": {
         "@dabble/delta": "^1.2.5",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dabble/patches",
-  "version": "0.18.0",
+  "version": "0.19.0",
   "description": "Immutable JSON Patch implementation based on RFC 6902 supporting operational transformation and last-writer-wins",
   "author": "Jacob Wright <jacwright@gmail.com>",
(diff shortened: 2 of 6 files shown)


RyanHavoc (approved):

matalina:
Added the version bump (0.18.0 → 0.19.0, minor for the new `dispose()` API — same precedent as 0.18.0 for the additive RelayTransport). Merging this enables an immediate `npm publish`.

PR #102, opened 2026-07-11
fix(client): reopen IndexedDB connection when the browser closes it mid-transaction

RyanHavoc:
## TL;DR

Fixes the `InvalidStateError: The database connection is closing` errors from `patches-sync` (Sentry **DABBLE-WRITER-3-CA**). `IndexedDBStore.transaction()` now transparently reopens the DB and retries once when the browser closes the connection out from under us, so an in-flight write survives instead of surfacing as a failed change round-trip. Ships as **0.19.2**.

## What was happening

The browser can move a still-referenced `IDBDatabase` connection into its "closing" state without us calling `close()`:
- **Safari suspending a backgrounded tab/worker** (this issue was Safari 26.5, ~48 min into the session), or
- a **`versionchange`** from another connection upgrading the DB (e.g. the missing-stores self-bump, another tab, or a `DB_VERSION` deploy).

The next `db.transaction()` throws `InvalidStateError` (DOMException code 11). It rode in on the `change` telemetry path — which is deliberately never downgraded because a failed local round-trip *can* mean rolled-back keystrokes — so it paged as an `error` even though the data was fine.

## The fix

- `transaction()` routes through a new `beginTransaction()` that catches the `InvalidStateError` (the **only** condition under which `transaction()` raises it — its other failures are `NotFoundError`/`InvalidAccessError`/`TypeError`, so matching the name is precise and engine-portable), drops the dead handle, reopens, and **retries once**.
  - Managed mode → the write completes on the fresh connection.
  - External databases (host owns the lifecycle) and genuine teardown/page-unload → the error still propagates.
- Added an `onversionchange` handler that closes + reopens proactively, so we also stop *blocking* another connection's upgrade.

## Tests

Two new cases in `IndexedDBStore-lifecycle.spec.ts` over `fake-indexeddb`:
- browser closes the live connection out from under us → next write reopens + succeeds on a new connection;
- a second connection upgrading past our version fires `versionchange` → we don't block it and stay usable on the new version.

Full suite green (2467 passed), plus type/lint/format.

## Rollout

Consumers (dabble-writer-3.0, dabble-rest) re-point their in-flight `@dabble/patches` bump from `^0.19.1` → `^0.19.2` after this publishes. The dw3 bump commit carries `Fixes DABBLE-WRITER-3-CA` so the Sentry issue auto-resolves on deploy.

### DIFF
--- a/src/client/IndexedDBStore.ts
+++ b/src/client/IndexedDBStore.ts
@@ -57,6 +57,14 @@ export class IndexedDBStore implements PatchesStore, BranchClientStore {
   private static readonly DB_VERSION = 2;
 
   protected db: IDBDatabase | null = null;
+  /**
+   * Terminal flag set by `close()`/`deleteDB()`. Any reopen path (suspend recovery,
+   * versionchange) nulls `this.db` before the async open completes, so a `close()` that
+   * lands in that window would early-return on `if (!this.db)` and then get silently
+   * resurrected by the open's `onsuccess`. Every reopen path checks this flag so a closed
+   * store stays closed.
+   */
+  protected closed = false;
   protected dbName?: string;
   protected dbPromise: Deferred<IDBDatabase>;
   protected external: boolean;
@@ -154,10 +162,24 @@ export class IndexedDBStore implements PatchesStore, BranchClientStore {
         this.initDB(null);
       } else {
         this.dbPromise.reject(request.error);
+        // A reopen (versionchange/suspend recovery) may have no awaiter, so a reopen that
+        // fails on anything but the VersionError handled above (quota, corruption) would
+        // surface as an unhandled rejection on an idle tab. Swallow it here; real getDB()
+        // callers still receive the rejected promise. Mirrors close()'s guard.
+        this.dbPromise.promise.catch(() => undefined);
       }
     };
     request.onsuccess = () => {
       const db = request.result;
+      // close()/deleteDB() ran while this open was in flight. A reopen nulls this.db, so
+      // that close() early-returned without settling this promise — honor the terminal close
+      // instead of resurrecting the connection, and reject so any waiting reopen unblocks.
+      if (this.closed) {
+        db.close();
+        this.dbPromise.reject(new Error('Store has been closed'));
+        this.dbPromise.promise.catch(() => undefined);
+        return;
+      }
       // Missing stores mean the database was created without this store's upgrade
       // subscribers — bump the version so onupgradeneeded fires and creates them
       const missing = [...this.requiredStores].some(name => !db.objectStoreNames.contains(name));
@@ -173,7 +195,7 @@ export class IndexedDBStore implements PatchesStore, BranchClientStore {
       // on this now-closing one (DABBLE-WRITER-3-CA).
       db.onversionchange = () => {
         db.close();
-        if (this.db === db) {
+        if (this.db === db && !this.closed) {
           this.db = null;
           this.dbPromise = deferred<IDBDatabase>();
           this.initDB();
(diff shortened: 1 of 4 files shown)


src/client/IndexedDBStore.ts
```diff
@@ -232,12 +258,44 @@ export class IndexedDBStore implements PatchesStore, BranchClientStore {
...
+    reopened = false
+  ): Promise<IDBTransaction> {
+    const db = await this.getDB();
+    try {
+      return db.transaction(storeNames, mode);
+    } catch (err) {
+      if (reopened || this.external || !this.dbName || !isConnectionClosingError(err)) throw err;
+      // Only reopen the connection we just found dead; a concurrent caller may have
+      // already swapped in a fresh one, in which case just retry against that.
+      if (this.db === db) {
```
jacwright: Both reopen paths can race `close()` and resurrect a closed store: this swap nulls `this.db`, a concurrent `close()` early-returns on `if (!this.db)`, then `initDB()`'s onsuccess re-populates `this.db` and the store stays open, contradicting the close() contract. Worst case, `deleteDB()` racing a versionchange reopen recreates the just-deleted DB as empty. A terminal `closed` flag set in `close()` and checked here + in the onversionchange handler would close the window.

src/client/IndexedDBStore.ts
```diff
@@ -153,6 +166,19 @@ export class IndexedDBStore implements PatchesStore, BranchClientStore {
...
+      // deploy that raised DB_VERSION) needs to upgrade the database. Close proactively
+      // so we don't block that upgrade, then drop our handle and reopen so the next
+      // transaction() runs on the fresh connection instead of throwing InvalidStateError
+      // on this now-closing one (DABBLE-WRITER-3-CA).
+      db.onversionchange = () => {
+        db.close();
+        if (this.db === db) {
+          this.db = null;
+          this.dbPromise = deferred<IDBDatabase>();
+          this.initDB();
```
jacwright, line 201: If this reopen fails with anything but the handled `VersionError` (quota, corruption), the fresh `dbPromise` rejects with no awaiter until the next operation, so an idle tab logs an unhandled rejection. `close()` guards this same pattern with `.catch(() => undefined)`.

jacwright (approved):
Fix is correct and precisely scoped: `InvalidStateError` is the right discriminator, the `this.db === db` guard handles concurrent callers, external mode and retry-once behave as described, and the versionchange test's `onblocked` rejection is a real regression tripwire. Bonus: the versionchange handler also fixes same-page sibling stores blocking each other's `requireStores` version bump. Two notes inline, neither blocking.

src/client/IndexedDBStore.ts
RyanHavoc: Fixed in c2c1974. Added a terminal `closed` flag: set at the top of `close()` before any await, checked in this reopen guard and the `onversionchange` handler, and — the authoritative backstop — enforced in `initDB`’s `onsuccess`, which closes the freshly-opened connection and rejects any waiting reopen if the store closed while the open was in flight. Regression test `close() during an in-flight reopen does not resurrect the store` spins microtasks to land in the exact null-`this.db` window, then closes; verified it fails without the guard.

src/client/IndexedDBStore.ts
RyanHavoc: Fixed in c2c1974. The reopen’s `onerror` reject now chains `.catch(() => undefined)`, same guard `close()` uses — real `getDB()` callers still get the rejection, but an awaiter-less reopen that fails on quota/corruption no longer logs an unhandled rejection. Regression test `a failed reopen does not surface an unhandled rejection` drives the versionchange reopen into a forced non-VersionError failure and asserts no `unhandledRejection`; verified it fails without the guard.

PR #103, opened 2026-07-13
fix(sync): classify raw status-less fetch failures instead of latching the doc

RyanHavoc:
## TL;DR

The largest client-side slice of the GA-cutover "stuck syncing / blank project" latches: a status-less fetch `TypeError` that escaped transport wrapping slipped past `isNetworkError` and latched the content doc at a terminal `error` instead of deferring to connection recovery. This makes `isNetworkError` recognize the browsers' fixed fetch-failure strings so those docs park + retry instead.

_(Scoped down from the original two-fix PR per review — the 413 half is split out; see below.)_

## Change

`isNetworkError` (`net/error.ts`) now also recognizes the browsers' fixed, non-localized status-less fetch-failure strings — Chromium "Failed to fetch", WebKit "Load failed", Gecko "NetworkError when attempting to fetch resource" — by name + an **exact** (normalized, trailing-period-stripped) message match. `PatchesREST._fetch` wraps fetch failures into a `NetworkError`, but any unwrapped path (or a bare `TypeError` re-thrown across the worker boundary) previously slipped past — `isNetworkError` only matched `name === 'NetworkError'` — and latched the doc. Now it defers to connection recovery like every other status-less failure. Exact rather than substring so near-misses like "Upload failed"/"Download failed" aren't swept in; a genuine programming `TypeError` stays unmatched.

## Tests

- `error.spec.ts`: raw fetch `TypeError`s (Chromium/WebKit/Gecko + worker-boundary copy) classify as network errors; "Upload failed"/"Download failed" and an unrelated `TypeError` do not.
- Full suite green (2470 passed, 4 skipped). type-check / lint / format clean.

## Split out: the 413 half

The original Fix 2 (413 → snapshot reload) is **removed** from this PR per @jacwright's review. The observed 413 cohort (`DABBLE-WRITER-3-DA`) is all `has_pending: true` → arrives via `flushDoc`, which the `!pending?.length` gate skipped entirely; the read path it *did* cover is already rescued server-side by pup, and `_reloadDocFromServer` would retry the same oversized read. The correct client gate is `data.scope === 'doc'` — but even then a version-less doc's `getDoc` is the same oversized 0→head read. The durable fix is server-side (**DAB-626**: rescue `commitChanges`, bound `getDoc`/capture a version), deploy pup-first — the client 413 recovery follows that.

## Follow-ups (not this PR)

- **403 Forbidden cohort** (largest overall, 21 users) — server-side (pup roleWrites/roles-config on the GA env), not a client latch bug.
- **dw3 re-point** — dw3's `^0.19.2` already accepts `0.19.3`; bump its lockfile + deploy after this publishes.

### DIFF
--- a/src/net/error.ts
+++ b/src/net/error.ts
@@ -74,17 +74,40 @@ export class NetworkError extends Error {
 }
 
 /**
+ * The fixed, non-localized messages browser fetch implementations put on the
+ * status-less `TypeError` they throw when a request dies before producing a
+ * response — offline, a DNS/TCP/TLS failure, a dropped connection, or a CORS-opaque
+ * rejection. Chromium: "Failed to fetch"; WebKit: "Load failed"; Gecko:
+ * "NetworkError when attempting to fetch resource". Compared lower-cased and with
+ * any trailing period stripped (Gecko's carries one), matched EXACTLY — a substring
+ * match would sweep in near-misses like "Upload failed" / "Download failed".
+ */
+const RAW_FETCH_FAILURE_MESSAGES = ['failed to fetch', 'load failed', 'networkerror when attempting to fetch resource'];
+
+/**
  * True for failures that never carried an HTTP status because the request died at
  * the network level — see {@link NetworkError}. Matches by name as well as
  * instance so errors from a duplicated module copy or a structured-clone boundary
  * (which preserves only name/message/stack) classify the same, and recognizes the
  * platform's raw timeout shape (`AbortSignal.timeout` firing mid body-read
- * surfaces a `TimeoutError` DOMException past any transport wrapping). Cancelled
- * requests/transactions are a sibling class — see {@link isAbortError}.
+ * surfaces a `TimeoutError` DOMException past any transport wrapping).
+ *
+ * Also recognizes a raw fetch rejection that escaped transport wrapping: a
+ * status-less `TypeError` whose message is one of the browsers' fixed fetch-failure
+ * strings ({@link RAW_FETCH_FAILURE_MESSAGES}). `PatchesREST` wraps these into a
+ * {@link NetworkError}, but a fetch on any unwrapped path (or one re-thrown across a
+ * worker boundary as a bare `TypeError`) would otherwise be misread as a doc-level
+ * failure and latch the doc at a terminal 'error' — so match it here by name+message.
+ * A `TypeError` with any other message (a genuine programming error) stays unmatched.
+ * Cancelled requests/transactions are a sibling class — see {@link isAbortError}.
  */
 export function isNetworkError(err: unknown): boolean {
   if (!(err instanceof Error)) return false;
-  return err.name === 'NetworkError' || err.name === 'TimeoutError';
+  if (err.name === 'NetworkError' || err.name === 'TimeoutError') return true;
+  if (err.name === 'TypeError') {
+    return RAW_FETCH_FAILURE_MESSAGES.includes(err.message.toLowerCase().replace(/\.$/, ''));
+  }
+  return false;
 }
 
 /**
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@dabble/patches",
-  "version": "0.19.2",
+  "version": "0.19.3",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@dabble/patches",
-      "version": "0.19.2",
+      "version": "0.19.3",
       "license": "MIT",
       "dependencies": {
         "@dabble/delta": "^1.2.5",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dabble/patches",
-  "version": "0.19.2",
+  "version": "0.19.3",
   "description": "Immutable JSON Patch implementation based on RFC 6902 supporting operational transformation and last-writer-wins",
   "author": "Jacob Wright <jacwright@gmail.com>",
   "bugs": {
--- a/tests/net/error.spec.ts
+++ b/tests/net/error.spec.ts
@@ -200,6 +200,22 @@ describe('StatusError', () => {
       expect(isNetworkError(new DOMException('The operation timed out.', 'TimeoutError'))).toBe(true);
     });
 
+    it('classifies raw status-less fetch TypeErrors that escaped transport wrapping', () => {
+      // A fetch on an unwrapped path (or one re-thrown across a worker boundary as a
+      // bare TypeError) surfaces the browser's fixed, non-localized failure string.
+      expect(isNetworkError(new TypeError('Failed to fetch'))).toBe(true); // Chromium
+      expect(isNetworkError(new TypeError('Load failed'))).toBe(true); // WebKit
+      expect(isNetworkError(new TypeError('NetworkError when attempting to fetch resource.'))).toBe(true); // Gecko
+      // Name survives structured clone even when the TypeError prototype does not.
+      const crossed = new Error('Failed to fetch');
+      crossed.name = 'TypeError';
+      expect(isNetworkError(crossed)).toBe(true);
+      // Exact (not substring) match: near-misses that merely contain "load failed"
+      // must not be swept into connection recovery.
+      expect(isNetworkError(new TypeError('Upload failed'))).toBe(false);
+      expect(isNetworkError(new TypeError('Download failed'))).toBe(false);
+    });
+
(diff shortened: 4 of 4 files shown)


src/net/PatchesSync.ts
```diff
@@ -607,6 +607,35 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
...
+      // doc's change history — the range from our committed rev to the head is too large
+      // to serve, so retrying the same range just 413s again and the doc would latch at
+      // 'error' forever. The authoritative snapshot is bounded, so recover the way a failed
+      // apply does: pull it and reset committed state, which also advances committedRev past
+      // the oversized range so the next catch-up is small. Gate on having been READING it
+      // (no pending): a 413 on a flush is an oversized UPLOAD, a different problem a snapshot
+      // reload can't fix — that stays on the retry/latch path below. Matched by numeric
+      // `code` (not `instanceof StatusError`) so a status preserved across a worker boundary
+      // classifies the same.
+      if (!pending?.length && (failure as { code?: unknown }).code === 413) {
```
jacwright: This gate inverts the observed cohort. Every live 413 event carries `has_pending: true` (Sentry `DABBLE-WRITER-3-DA`), so it arrives via `flushDoc` and `!pending?.length` skips this branch. And the read path it *does* cover is one pup already rescues server-side (`getChangesSince` → `serveSnapshotCatchup`) — when that rescue fails it's because `otReadServer.getDoc` failed, which is the exact read `_reloadDocFromServer` is about to retry, so it fails again and latches regardless.

pup's history-read 413 carries `data.scope === 'doc'`, and that survives to the client (`onError` forwards `data`; `PatchesREST._fetch` parses it). Gate on that instead of on `pending`:

```ts
const d = failure as { code?: unknown; data?: { scope?: string } };
if (d.code === 413 && d.data?.scope === 'doc') {
```

That catches the flush-path 413 (the one users are latched on) and still leaves a genuine oversized *upload* 413 — which carries no `data` — on the latch path.

src/net/error.ts
```diff
@@ -73,18 +73,40 @@ export class NetworkError extends Error {
...
 }
 
+/**
+ * The fixed, non-localized messages browser fetch implementations put on the
+ * status-less `TypeError` they throw when a request dies before producing a
+ * response — offline, a DNS/TCP/TLS failure, a dropped connection, or a CORS-opaque
+ * rejection. Chromium: "Failed to fetch"; WebKit: "Load failed"; Gecko:
+ * "NetworkError when attempting to fetch resource". Compared lower-cased.
+ */
+const RAW_FETCH_FAILURE_MESSAGES = ['failed to fetch', 'load failed', 'networkerror when attempting to fetch'];
```
jacwright: `includes` on `'load failed'` also matches "Upload failed" and "Download failed" — a `TypeError` carrying either would be silently parked for connection recovery instead of surfacing. These strings are fixed and non-localized, so an exact match on the normalized message is both safer and cheaper:

```ts
return RAW_FETCH_FAILURE_MESSAGES.includes(message.replace(/\.$/, ''));
```

(Gecko's has a trailing period, hence the strip.)

jacwright (changes requested):
Fix 1 (`isNetworkError`) is right and I'd take it as-is. Fix 2 (413) is gated on the wrong signal — as written it can't recover a single one of the 413 events it's aimed at. Details inline; summary here.

**The 413 gate excludes the only path that actually latches.**

pup already rescues the read path. `PupServer.getChangesSince` catches the store's oversized-history 413 and degrades to `serveSnapshotCatchup`, so a client's `getChangesSince` almost never surfaces a 413 at all. The one case where it does is when `serveSnapshotCatchup` itself fails — a version-less doc whose full history *is* the oversized read — and there its own comment says "the original 413 propagates." In exactly that case this PR's recovery calls `_reloadDocFromServer` → `connection.getDoc` → `PupServer.getDoc` → `readDoc` → the *same* `otReadServer.getDoc(path)` that just failed inside `serveSnapshotCatchup` (content docs are OT, so the Redis doc-cache branch doesn't apply). It fails identically, `failure = reloadErr`, and the doc latches anyway. Net effect: one extra doomed round-trip, same latch.

Meanwhile `PupServer.commitChanges` has **no** rescue — it does an unbounded `listChanges({startAfter: baseRev})` and rethrows the 413 raw. That's the flush path, and it's the one this PR explicitly excludes.

The live Sentry event ([DABBLE-WRITER-3-DA](https://dabblewriter.sentry.io/issues/DABBLE-WRITER-3-DA)) settles it:

```
has_pending: true
http_status: 413
error_message: "Payload Too Large: Change history read for doc projects/*/content exceeds
  safe serving limits (14957 changes, 6MB stored ops); reload the document instead of
  replaying changes"
```

`has_pending: true` on every event on that issue. So `getPendingToSend` returned changes, `syncDoc` went down `flushDoc`, and the new `!pending?.length` gate skips the recovery entirely.

**The premise is the bug.** "A 413 on a flush is an oversized upload" isn't true for pup: its flush-path 413 is a change-history *read* the server performs to answer the commit — its message literally says "reload the document instead of replaying changes." A genuine oversized upload is a *different* 413 (`rest.ts` "Request body too large"), and it carries no `data`.

**The signal to gate on already exists and already reaches the client.** pup tags the oversized-history 413 with `data.scope === 'doc'` (that's what its own `isOversizedHistoryError` matches), `onError` forwards `data` in the REST error body, and `PatchesREST._fetch` parses it into `StatusError.data`. So:

```ts
const d = (failure as { code?: unknown; data?: { scope?: string } });
if (d.code === 413 && d.data?.scope === 'doc') { /* reload */ }
```

That covers the flush path (the cohort that latches) and still leaves a real body-too-large 413 on the latch path — which is the distinction you were reaching for, just drawn on the right axis.

**One caveat even after that fix.** `_reloadDocFromServer` → `getDoc` reads from the doc's last sealed version to head. For a *version-less* doc that's 0→head — the same oversized read — so it 413s too and the doc stays stuck. The client can only dig out docs that have a sealed version. The durable fix is server-side (DAB-626: rescue `commitChanges`, bound `getDoc`/capture a version), and the deploy order there is pup-first precisely because this file is production sync core. Worth not shipping this as "fixes the 413 cohort" until we know whether `projects_content/{pid}/_versions` is empty for the affected projects.

Suggestion: split fix 1 out and merge it now (it stands on its own and rescues the largest client-side cohort), and let the 413 half follow the pup rescue.

RyanHavoc:
Thanks @jacwright — sharp review, addressed both:

- **Fix 1 (`isNetworkError`):** switched to an exact (normalized, trailing-period-stripped) message match instead of substring `includes`, so "Upload failed"/"Download failed" no longer match; added negative tests for both. Otherwise as-is.
- **Fix 2 (413):** removed from this PR per your suggestion. You're right the gate was on the wrong axis — `!pending?.length` skips the entire `has_pending: true` cohort that actually latches (via `flushDoc`), and the read path it covered is already pup-rescued (`_reloadDocFromServer` would retry the same oversized read). And even the corrected `data.scope === 'doc'` gate can't dig out a version-less doc, since its `getDoc` is the same 0→head read. Agreed the durable fix is **DAB-626** (rescue `commitChanges`, bound `getDoc`/capture a version) deployed pup-first, with the client 413 recovery following — and I won't reintroduce it as "fixes the 413 cohort" until we know whether `projects_content/{pid}/_versions` is empty for the affected projects.

Force-pushed; PR is now Fix 1 only (2470 tests green, checks clean).

src/net/error.ts
RyanHavoc: Good catch — switched to an exact match on the normalized message (`toLowerCase()` + strip a trailing period) against the full strings, so "Upload failed"/"Download failed" no longer match. Added negative tests for both.

src/net/PatchesSync.ts
RyanHavoc: Agreed — removed the 413 recovery from this PR. `!pending?.length` skipped the entire `has_pending: true` cohort that's actually latched (via `flushDoc`), the read path was already pup-rescued, and the reload would retry the same oversized read. Even the `data.scope === 'doc'` gate can't dig out a version-less doc (its `getDoc` is the same 0→head read), so the durable fix is server-side (DAB-626, pup-first). This PR is now Fix 1 only.

jacwright (approved):
Both addressed. Exact match on the normalized message is right, and the near-miss cases ("Upload failed" / "Download failed") are pinned in the test. Dropping the 413 recovery rather than re-gating it is the right call for now — the flush-path 413 needs the pup rescue first (DAB-626), and this half stands on its own and rescues the largest client-side cohort. Ship it.

PR #104, opened 2026-07-13
fix(server): chain captured versions to their parent

jacwright:
## TL;DR

Auto-saved versions were being rebuilt by replaying a project's **entire** edit history from the very beginning, every single time. On small projects nobody noticed. On big ones it got so slow it failed outright, which meant **712 projects silently stopped auto-saving versions today** and their owners had no idea. This makes version-saving read only the small slice of new edits since the last version, the way it was always supposed to.

Fixes the larger half of DAB-626.

## The bug

`captureCurrentVersion` never passed a `parentId`:

```ts
// src/server/OTServer.ts:222 (before)
const version = await createVersion(this.store, docId, changes, { metadata });
//                                                             ^ no parentId
```

`buildVersionState` -> `getBaseStateBeforeVersion` chains a new version onto its parent's stored state so it only replays the **gap** between them. With no `parentId`, `baseRev` stays `0` and it replays the whole history from rev 1, with no `limit`. The sibling `createVersionAtRev` chains correctly; `captureCurrentVersion` was the one version writer that didn't.

This has always been true, on every auto-version of every OT doc. Invisible on a median 30-change doc. On a 13k-change doc it is a 13,614-change decompress. It is very likely what OOM'd pup on Jul 6-8 (110MB stored, ~8GB decompressed). Pup's 10k/64MB cap turned that crash loop into a visible 413, which is how we found it.

## The fix

`getLatestMainVersion` already loaded the parent version and threw away `version.id`. Return it, pass it to `createVersion`, and warn when a version with `startRev > 1` has no parent.

Review also caught a **second unchained writer**: `mergeBranch` emitted unchained versions on every healthy branch merge, which would have fired the new warning constantly and destroyed its value as the detector for the ~700 legacy unchained versions already in prod. Fixed too.

And it caught that the new E2E test **could not fail**: `VersionBuildingBackend` discarded `buildVersionState`'s return value and `OTFuzzBackend` re-derived state by full replay, so the assertions read the true state no matter what the chain computed. Proven by sabotaging `buildVersionState` to return `{SABOTAGED: true}`: the spec stayed 4/4 green. Now it fails 2/4 under the same sabotage. There is a real regression guard here now.

## Verification

```
npm test        2481 passed | 4 skipped
npm run type:check   pass
npm run lint         pass
npm run format:check pass
```

## Deploy

patches must be released and pup's `@dabble/patches` bumped (currently `^0.19.1`) before this reaches prod. Safe to ship first and alone: version-path only, does not touch the sync/commit core.

Does **not** fix the ~700 already-unchained versions in prod; the new warning makes them findable. Their next auto-version chains correctly regardless.

---

EVIDENCE_PACK
## Evidence pack

Everything below was measured **read-only** against production on 2026-07-13 (launch day of the v2.5 to v3.0 cutover). No writes of any kind. Reproduce it yourself before trusting it.

**Sources**

- Sentry issue `DABBLE-WRITER-3-DA` — `sync doc error latched: projects/{id}/content (413)`. Client-side, emitted from dw3 `src/services/repairTelemetry.ts:425`.
- Sentry alert: https://dabblewriter.sentry.io/alerts/rules/details/441891/
- Slack thread: https://dabblewriter.slack.com/archives/C0BFPA35GC8/p1783953299051229
- pup Cloud Logging, GCP project `dabble-writer`, Cloud Run service `pup`, revision `pup-00039-4fr`
- Firestore database `dabble-next` (prod), collection `projects_content`
- Linear: DAB-626

### 1. There are two 413s on two routes, not one

pup's store logs a structured warn every time it refuses an unbounded read. Field is `jsonPayload.msg` (not `.message` — a wrong field name here silently returns zero rows):

```bash
gcloud logging read \
  'resource.type="cloud_run_revision"
   resource.labels.service_name="pup"
   jsonPayload.msg=~"Refusing unbounded change history read"
   timestamp>="2026-07-13T00:00:00Z"' \
  --project=dabble-writer --limit=5000 --format=json > warns.json
```

A record looks like:

```json
{
  "jsonPayload": {
    "count": 15000,
    "docId": "projects/MwIDQtxEEFQ8JiVf/content",
    "opsBytes": 4579426,
    "maxUnboundedReadChanges": 10000,
    "maxUnboundedReadBytes": 67108864,
    "msg": "Refusing unbounded change history read: exceeds safe serving limits"
  },
  "severity": "WARNING",
  "timestamp": "2026-07-13T18:08:47.878694Z"
}
```

Correlating the HTTP 413s by route over the same window:

```
  _versions     4983 HTTP 413s    690 distinct docs      <- Bug A
  _changes        17 HTTP 413s     16 distinct docs      <- Bug B (what Sentry showed)
```

Store-level, across 5,000 sampled refusals: **712 distinct documents**.

### 2. The two bugs have different, mechanically distinguishable signatures

Group each doc's refused read `count` across its retries:

```
  read size PINNED  : 447 docs   count identical on every retry -> read starts at rev 0 and never moves
  read size GROWING :  41 docs   count tracks the tip -> a FIXED startAfter:0 against a MOVING head
```

`MGZeSdToRcwOr2ds` shows both, and its log is the whole story in six lines:

```
13:48:54  POST _changes   413   count=13614   opsBytes=6,274,710
13:57:09  POST _versions  413   count=13614   opsBytes=6,274,710
14:05:11  POST _versions  413   count=13614   opsBytes=6,274,710
14:07:24  POST _versions  413   count=13614   opsBytes=6,274,710
14:07:32  POST _changes   413   count=13624   opsBytes=7,541,634
14:37-16:55  POST _versions  413 x12   count=13614 every time   (dabble-rest webhook retry backoff)
```

`_versions` is **pinned at 13,614** across all 16 retries. `_changes` **grows with the tip** (13,614 then 13,624). Two different reads.

### 3. The commit read provably starts at `startAfter: 0`

Summing the stored `ops` byte length of `MGZeSdToRcwOr2ds` revs 1..13,624 in Firestore gives **7,541,634 bytes**, matching the logged read **to the byte**. Revs 5..13,628 would give 8,312,230, which does not match. The client sent `baseRev: 0`: it believed the document was empty.

`OMm6EH1W1E1ErML5` independently corroborates: three retries seconds apart, read count growing by exactly one each time (14,957 → 14,958 → 14,959) while another client committed. That is a fixed `startAfter: 0` against a moving tip, not a lagging client.

### 4. Versioning is NOT the problem, and the "version-less deadlock" theory is dead

An earlier writeup (including my own first pass, and the first version of the artifact) claimed these docs had no sealed version and were deadlocked. **That was wrong.** Sampling `projects_content` and computing the real span metric (`tipRev - lastEndRev` where `origin == 'main'`):

```
zero versions                     : 0  (0.00%)
zero versions AND >10k changes    : 0  (0.00%)
span percentiles: p50=0 p90=0 p99=0 max=47
tip  percentiles: p50=36 p90=1277 p99=20523 max=130734
```

The three incident docs are all healthy on this metric:

| Doc                | Changes (tip) | Versions | Last main endRev | **Span** |
| ------------------ | ------------: | -------: | ---------------: | -------: |
| `MGZeSdToRcwOr2ds` |        13,628 |       52 |           13,614 |   **14** |
| `GhQEuv6mpiFovu4A` |        23,578 |       44 |           23,574 |    **4** |
| `Cb1xCJT28vZX0YZY` |        16,393 |      150 |           16,379 |   **14** |

**The nuance that makes this coherent:** `02hhVxgdOE9x8lt1` has **130,734 changes, 1,012 versions, span of 1**. Versioning demonstrably works at enormous scale. That is because `catchUpVersions` (called inside `commitChanges`) uses `createVersionAtRev`, which **does** chain `parentId` correctly, and it keeps the watermark within `maxChangesPerVersion` (1000) of head. Only `captureCurrentVersion` — the 30-minute-inactivity auto-version, reached via `POST _versions` — fails to chain a parent. So the span metric stays healthy _because a different code path is doing the work_, while `POST _versions` 413s forever. The span metric was measuring the read that works. There was never a contradiction.

### 5. Corrections to earlier claims (do not propagate these)

- ~~"About 7.5% of active projects are already over 10k changes"~~ — measured total change count, which is **not** the risk metric (the cap keys on read _span_). Actual: ~2.17% of sampled `projects_content` docs exceed 10k total changes.
- ~~"GhQEuv6mpiFovu4A has ~15,000 changes"~~ — it has **23,578**.
- ~~"The affected projects have no sealed version and are deadlocked"~~ — they have 44 to 150 versions each. Dead.

### 6. Why nobody saw the bigger bug

`pup/src/onError.ts` gates Sentry capture on `status === 503` (~line 99) and `status >= 500 || isCorruptHistory` (~line 111). The oversized-history 413 is a plain `StatusError` with `name === 'Error'`, so it matches neither. **pup threw roughly 1,250 of these an hour, all day, and reported exactly zero.** Every Sentry event the team has is client-side, from dw3.

Separately, `PupServer.ts:930-936` notes that the JSON-RPC/WS commit path never passes through `onError` at all. The incident traffic was REST (`POST` routes), so `onError` is the right place for this fix, but the WS path is a known second gap.

### 7. Reproduction scripts

Read-only, in the session scratchpad: `span.mjs` (version span across `projects_content`), `bytes.mts` (per-rev stored-ops byte sum, proves `startAfter: 0`), `evidence-query.sh` (the log split above). All use ADC (`gcloud auth application-default print-access-token`) and the Firestore REST API. None writes.

### DIFF
--- a/src/algorithms/ot/server/buildVersionState.ts
+++ b/src/algorithms/ot/server/buildVersionState.ts
@@ -2,6 +2,7 @@ import { jsonReadable, parseVersionState } from '../../../server/jsonReadable.js
 import type { OTStoreBackend } from '../../../server/types.js';
 import type { Change, PatchesState, VersionMetadata } from '../../../types.js';
 import { applyChanges, applyChangesForReconstruction, type ReplayOptions } from '../shared/applyChanges.js';
+import { findLatestMainVersion } from './getSnapshotAtRevision.js';
 
 /**
  * Applies committed changes per the caller's replay options: strict by default,
@@ -14,13 +15,50 @@ function replayChanges<T>(state: T, changes: Change[], options?: ReplayOptions):
 }
 
 /**
+ * Loads the state of the version `version` chains from, and the rev it ends at.
+ *
+ * A version that starts past rev 1 with no recorded `parentId` is a bug in whatever wrote it:
+ * every base state below is bridged from `parent.endRev`, so without a parent the bridge runs
+ * from rev 0 and replays the document's *entire* history — unbounded, and on a large document
+ * expensive enough to fail the read outright. Rather than do that silently, resolve the parent
+ * the writer should have recorded. The state it yields is identical; the read is bounded.
+ *
+ * Returns undefined only when there genuinely is no earlier version to chain from (the first
+ * version of a document), where replaying from rev 1 is the only way to build the base.
+ */
+async function loadParentState(
+  store: OTStoreBackend,
+  docId: string,
+  version: VersionMetadata
+): Promise<{ rawState: string | ReadableStream<string>; endRev: number } | undefined> {
+  let parentId = version.parentId;
+
+  if (!parentId && version.startRev > 1) {
+    parentId = (await findLatestMainVersion(store, docId, version.startRev - 1))?.id;
+    if (parentId) {
+      console.warn(
+        `Version ${version.id} of doc ${docId} starts at rev ${version.startRev} with no parentId; chaining to ${parentId} rather than replaying from rev 1.`
+      );
+    }
+  }
+  if (!parentId) return;
+
+  const [rawState, parentMeta] = await Promise.all([
+    store.loadVersionState(docId, parentId),
+    store.loadVersion(docId, parentId),
+  ]);
+  if (!parentMeta || rawState === undefined) return;
+
+  return { rawState, endRev: parentMeta.endRev };
+}
+
+/**
  * Computes the document state at `version.startRev - 1`: the state just before
  * this version's changes begin.
  *
- * Uses `version.parentId` (via `store.loadVersion`) when available for
- * an efficient cached lookup. After loading the parent state, bridges any gap
- * between the parent's `endRev` and `version.startRev` by applying intermediate
- * changes. When no parentId exists (first version ever), builds from scratch.
+ * Loads the parent version's state (see {@link loadParentState}), then bridges any gap
+ * between the parent's `endRev` and `version.startRev` by applying intermediate changes.
+ * With no earlier version to chain from (first version ever), builds from scratch.
  *
  * @param store - The store backend.
  * @param docId - The document ID.
@@ -35,19 +73,9 @@ export async function getBaseStateBeforeVersion(
   version: VersionMetadata,
   options?: ReplayOptions
 ): Promise<PatchesState> {
-  let baseState: any = null;
-  let baseRev = 0;
-
-  if (version.parentId) {
-    const [rawState, parentMeta] = await Promise.all([
-      store.loadVersionState(docId, version.parentId),
-      store.loadVersion(docId, version.parentId),
-    ]);
-    if (parentMeta && rawState !== undefined) {
-      baseState = await parseVersionState(rawState);
-      baseRev = parentMeta.endRev;
-    }
-  }
+  const parent = await loadParentState(store, docId, version);
+  let baseState: any = parent ? await parseVersionState(parent.rawState) : null;
+  const baseRev = parent?.endRev ?? 0;
 
   // Bridge any gap between baseRev and version.startRev.
   if (baseRev < version.startRev - 1) {
@@ -87,32 +115,27 @@ export async function getStateBeforeVersionAsStream(
   version: VersionMetadata,
   options?: ReplayOptions
 ): Promise<ReadableStream<string>> {
-  if (version.parentId) {
-    const [rawState, parentMeta] = await Promise.all([
-      store.loadVersionState(docId, version.parentId),
-      store.loadVersion(docId, version.parentId),
-    ]);
+  const parent = await loadParentState(store, docId, version);
 
-    if (parentMeta && rawState !== undefined) {
-      const baseRev = parentMeta.endRev;
+  if (parent) {
+    const { rawState, endRev: baseRev } = parent;
 
-      if (baseRev >= version.startRev - 1) {
-        // No gap — stream raw bytes directly without parsing.
-        return typeof rawState === 'string' ? jsonReadable(rawState) : rawState;
-      }
-
-      // Has gap — parse, apply changes, re-serialize.
-      const baseState = await parseVersionState(rawState);
-      const gapChanges = await store.listChanges(docId, {
-        startAfter: baseRev,
-        endBefore: version.startRev,
-      });
-      const state = gapChanges.length > 0 ? replayChanges(baseState, gapChanges, options) : baseState;
-      return jsonReadable(JSON.stringify(state));
+    if (baseRev >= version.startRev - 1) {
+      // No gap — stream raw bytes directly without parsing.
+      return typeof rawState === 'string' ? jsonReadable(rawState) : rawState;
     }
+
+    // Has gap — parse, apply changes, re-serialize.
+    const baseState = await parseVersionState(rawState);
+    const gapChanges = await store.listChanges(docId, {
+      startAfter: baseRev,
+      endBefore: version.startRev,
+    });
+    const state = gapChanges.length > 0 ? replayChanges(baseState, gapChanges, options) : baseState;
+    return jsonReadable(JSON.stringify(state));
   }
 
-  // No parentId (first version ever) — build from scratch via gap changes.
+  // No earlier version (first version ever) — build from scratch via gap changes.
   if (version.startRev > 1) {
     const gapChanges = await store.listChanges(docId, {
       startAfter: 0,
--- a/src/algorithms/ot/server/buildVersionState.ts
+++ b/src/algorithms/ot/server/buildVersionState.ts
@@ -23,33 +23,56 @@ function replayChanges<T>(state: T, changes: Change[], options?: ReplayOptions):
  * expensive enough to fail the read outright. Rather than do that silently, resolve the parent
  * the writer should have recorded. The state it yields is identical; the read is bounded.
  *
- * Returns undefined only when there genuinely is no earlier version to chain from (the first
- * version of a document), where replaying from rev 1 is the only way to build the base.
+ * Returns undefined when there is no usable parent to chain from — no earlier version exists,
+ * the parent (recorded or resolved) can't be loaded, or the parent's snapshot overlaps this
+ * version's range — leaving the caller to build the base by replaying from rev 1. Any of those
+ * on a version past rev 1 warns, since that replay is the unbounded read chaining exists to avoid.
  */
 async function loadParentState(
   store: OTStoreBackend,
   docId: string,
   version: VersionMetadata
 ): Promise<{ rawState: string | ReadableStream<string>; endRev: number } | undefined> {
-  let parentId = version.parentId;
+  const recordedParentId = version.parentId;
+  let parentMeta: VersionMetadata | undefined;
+  let rawState: string | ReadableStream<string> | undefined;
 
-  if (!parentId && version.startRev > 1) {
-    parentId = (await findLatestMainVersion(store, docId, version.startRev - 1))?.id;
-    if (parentId) {
+  if (recordedParentId) {
+    [rawState, parentMeta] = await Promise.all([
+      store.loadVersionState(docId, recordedParentId),
+      store.loadVersion(docId, recordedParentId),
+    ]);
+  } else if (version.startRev > 1) {
+    // Resolve the parent the writer should have recorded. The resolved metadata is already in
+    // hand, so only the state needs loading — no re-fetch that a transient miss could fail.
+    parentMeta = await findLatestMainVersion(store, docId, version.startRev - 1);
+    if (parentMeta) rawState = await store.loadVersionState(docId, parentMeta.id);
+  }
+
+  if (parentMeta && rawState !== undefined) {
+    // A parent whose snapshot extends into this version's own range can't be chained from: its
+    // state already contains revs at or past startRev, so bridging (or fast-path streaming)
+    // from it would serve too-new state. Resolved parents can't overlap — findLatestMainVersion
+    // caps endRev at startRev - 1 — so this only catches bad recorded parentIds.
+    if (parentMeta.endRev > version.startRev - 1) {
       console.warn(
-        `Version ${version.id} of doc ${docId} starts at rev ${version.startRev} with no parentId; chaining to ${parentId} rather than replaying from rev 1.`
+        `Version ${version.id} of doc ${docId} has parent ${parentMeta.id} whose endRev ${parentMeta.endRev} overlaps its startRev ${version.startRev}; ignoring the parent and replaying from rev 1.`
+      );
+      return;
+    }
+    if (!recordedParentId) {
+      console.warn(
+        `Version ${version.id} of doc ${docId} starts at rev ${version.startRev} with no parentId; chaining to ${parentMeta.id} rather than replaying from rev 1.`
       );
     }
+    return { rawState, endRev: parentMeta.endRev };
   }
-  if (!parentId) return;
 
-  const [rawState, parentMeta] = await Promise.all([
-    store.loadVersionState(docId, parentId),
-    store.loadVersion(docId, parentId),
-  ]);
-  if (!parentMeta || rawState === undefined) return;
-
-  return { rawState, endRev: parentMeta.endRev };
+  if (version.startRev > 1) {
+    console.warn(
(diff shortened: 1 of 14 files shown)


src/algorithms/ot/server/buildVersionState.ts
```diff
@@ -13,14 +14,51 @@ function replayChanges<T>(state: T, changes: Change[], options?: ReplayOptions):
...
+  store: OTStoreBackend,
+  docId: string,
+  version: VersionMetadata
+): Promise<{ rawState: string | ReadableStream<string>; endRev: number } | undefined> {
+  let parentId = version.parentId;
+
+  if (!parentId && version.startRev > 1) {
+    parentId = (await findLatestMainVersion(store, docId, version.startRev - 1))?.id;
+    if (parentId) {
+      console.warn(
```
RyanHavoc, line 58: This warn can be false: it fires before the resolved parent's metadata/state load, and if either is missing the function returns undefined at [line 50](https://github.com/dabblewriter/patches/blob/5affe96c8b7924fde4b0424ee5dc197ff308b18c/src/algorithms/ot/server/buildVersionState.ts#L50) and the caller replays from rev 1 anyway — after logging that it didn't. The detector also has two silent populations: a version with a recorded-but-dead parentId skips resolution entirely (the branch is gated on `!parentId`), and a doc with no main-origin versions resolves nothing and replays with no warn — so the fleet signal both over- and under-reports exactly where the unbounded replay still happens. Moving the warn after a successful load (and warning on the full-replay fallback itself) makes it trustworthy.

src/algorithms/ot/server/handleOfflineSessionsAndBatches.ts
```diff
@@ -25,10 +25,10 @@ export async function handleOfflineSessionsAndBatches(
   changes: Change[],
   origin: 'main' | 'offline-branch' = 'offline-branch'
 ) {
-  // For offline branches, find the last main version to use as the initial parent.
-  // This anchors the offline session chain to the main timeline.
+  // Find the last main version to use as the initial parent. This anchors the session chain to
+  // the main timeline; unanchored, the first session's state build replays the doc from rev 1.
   let parentId: string | undefined;
-  if (origin === 'offline-branch' && changes.length > 0) {
+  if (changes.length > 0) {
```
RyanHavoc, line 33: Extending parent-anchoring to main-origin sessions exposes a client-side gap that deserves its own issue: `PatchesHistoryClient.scrubTo` applies session changes directly onto the parent version's state with no bridge across `(parent.endRev, startRev)` ([PatchesHistoryClient.ts:105-111](https://github.com/dabblewriter/patches/blob/5affe96c8b7924fde4b0424ee5dc197ff308b18c/src/client/PatchesHistoryClient.ts#L105-L111)) — and the parent chosen here is routinely non-contiguous, since the live tail is un-versioned by design. Pre-PR these sessions were parentless and scrubbed to a near-blank baseline (the known scrub-blank symptom); post-PR they scrub to plausible-but-incomplete text, which reads as trustworthy. The server path bridges correctly via `getBaseStateBeforeVersion`; the fix is using the gap-bridging read from `scrubTo`. Two smaller notes on this block: it hand-rolls the exact query `findLatestMainVersion` was exported for (fourth copy of the reversed-cursor subtlety, minus the orphan clamp), and for `historicalImport` fast-forwards with preserved claimed revs the lookup can select an orphan main version stamped in the `(oldTip, firstRev)` hole — old code skipped the lookup entirely for `'main'`.

src/algorithms/ot/server/buildVersionState.ts
```diff
@@ -13,14 +14,51 @@ function replayChanges<T>(state: T, changes: Change[], options?: ReplayOptions):
...
+ */
+async function loadParentState(
+  store: OTStoreBackend,
+  docId: string,
+  version: VersionMetadata
+): Promise<{ rawState: string | ReadableStream<string>; endRev: number } | undefined> {
+  let parentId = version.parentId;
+
+  if (!parentId && version.startRev > 1) {
+    parentId = (await findLatestMainVersion(store, docId, version.startRev - 1))?.id;
```
RyanHavoc: The fallback can anchor to an orphan: a main version stamped past the log at mint time (the class this codebase documents as "never gets cleaned up") passes the `getCurrentRev` clamp once later commits grow the log past its claimed endRev, and its state contains ops the committed log never held. The old full replay was immune — it only ever read committed changes. Narrow window, but the damage is durable: a capture that chains to it records the orphan as parent permanently. A cheap validation (skip candidates whose range has no committed backing, or compare `parent.endRev` against the log) would close it.

src/algorithms/ot/server/buildVersionState.ts
```diff
@@ -35,19 +73,9 @@ export async function getBaseStateBeforeVersion(
...
-      store.loadVersion(docId, version.parentId),
-    ]);
-    if (parentMeta && rawState !== undefined) {
-      baseState = await parseVersionState(rawState);
-      baseRev = parentMeta.endRev;
-    }
-  }
+  const parent = await loadParentState(store, docId, version);
+  let baseState: any = parent ? await parseVersionState(parent.rawState) : null;
+  const baseRev = parent?.endRev ?? 0;
```
RyanHavoc, line 101: No guard that `parent.endRev <= version.startRev - 1`. Fallback-resolved parents are provably safe (the reversed-cursor arithmetic caps endRev at startRev-1), but a recorded overlapping parent — constructible from the pre-existing capture race plus the merge copy chain — skips the bridge at [line 81](https://github.com/dabblewriter/patches/blob/5affe96c8b7924fde4b0424ee5dc197ff308b18c/src/algorithms/ot/server/buildVersionState.ts#L81) and yields a base that already contains changes at/after startRev; the stream fast-path serves the too-new bytes directly as a scrub baseline. One assert (or clamp-with-warn) catches the whole class.

src/server/OTBranchManager.ts
```diff
@@ -217,13 +218,17 @@ export class OTBranchManager implements BranchManager {
         lastVersionId = v.id;
         continue;
       }
+      const startRev = toSourceRev(v.startRev);
+      // The first copy has no earlier copy to chain to, so it anchors to the source's own
+      // timeline. Unanchored, building its state replays the source document from rev 1.
+      const parentId = lastVersionId ?? (await findLatestMainVersion(this.store, sourceDocId, startRev - 1))?.id;
```
RyanHavoc, line 224: Reproduced empirically: on a source with no main-origin version below the branch point, the first copy of a second merge round writes with the parentId key deleted ([line 233](https://github.com/dabblewriter/patches/blob/5affe96c8b7924fde4b0424ee5dc197ff308b18c/src/server/OTBranchManager.ts#L233)) and the read-time fallback can't rescue it either (same `origin: 'main'` filter), so building its state replays the source's whole log from rev 1 — silently, since the warn only fires when a parent was resolved. Rare population, but it's the one place the unbounded replay survives this PR with zero telemetry.

src/server/OTServer.ts
```diff
@@ -212,14 +212,19 @@ export class OTServer implements PatchesServer {
...
    * @param docId The document ID.
    * @param metadata Optional metadata for the version.
    * @returns The ID of the created version, or null if no changes to capture.
    */
   async captureCurrentVersion(docId: string, metadata?: EditableVersionMetadata): Promise<string | null> {
     assertVersionMetadata(metadata);
-    const { changes } = await getSnapshotAtRevision(this.store, docId);
-    const version = await createVersion(this.store, docId, changes, { metadata });
+    const parent = await findLatestMainVersion(this.store, docId);
+    const changes = await this.store.listChanges(docId, { startAfter: parent?.endRev ?? 0 });
```
RyanHavoc, line 226: Two non-blocking notes for the record: (1) this tail read is unbounded (no limit), so a dormant doc with a >10k-change backlog since its last main version still 413s on capture under pup's caps — pre-existing (the old `getSnapshotAtRevision` read was identical), and it self-heals on the doc's next commit via `catchUpVersions`; bounded stepping here would additionally make capture a manual healing lever for latched docs. (2) Chaining removes an accidental safety property: parentless captures forced a full-log rebuild each time, which quietly scrubbed snapshot-only poison (state disagreeing with the log); now a bad parent snapshot propagates into every subsequent version and into `getDoc`, which streams the latest main version's persisted state verbatim. A sampled full-replay A/B or a repair hook would restore the coverage deliberately rather than accidentally.

src/algorithms/ot/server/buildVersionState.ts
```diff
@@ -13,14 +14,51 @@ function replayChanges<T>(state: T, changes: Change[], options?: ReplayOptions):
...
+    parentId = (await findLatestMainVersion(store, docId, version.startRev - 1))?.id;
+    if (parentId) {
+      console.warn(
+        `Version ${version.id} of doc ${docId} starts at rev ${version.startRev} with no parentId; chaining to ${parentId} rather than replaying from rev 1.`
+      );
+    }
+  }
+  if (!parentId) return;
+
+  const [rawState, parentMeta] = await Promise.all([
```
RyanHavoc: When the fallback resolved the parent, this `Promise.all` re-fetches metadata `findLatestMainVersion` already returned — one redundant read per legacy-version build, and a transient `loadVersion` miss sends the caller to the full-history replay despite having had `endRev` in hand two lines earlier. Carrying the returned `VersionMetadata` and loading only the state fixes both. Related author decision worth recording: the resolved parentId cannot be persisted today (`EditableVersionMetadata` disallows `parentId`, and `assertVersionMetadata` blocks it over RPC), so the ~700 legacy unchained versions re-pay the resolution reads and re-emit the warn on every state build, forever — a one-time backfill or an internal store-level write-back would let this fallback converge to a shim instead of permanent infrastructure. And since three writers each needed this fix individually, a `parentId ??= findLatestMainVersion(...)` default inside `createVersion` (verified safe for all current algorithm-path callers) would stop the next writer from reintroducing the bug — noting `mergeBranch` writes `store.createVersion` directly and needs its own guard either way.

RyanHavoc (commented):
Second-model review, as requested in Slack. What held up under verification: the chaining fix is correct and state-preserving — an A/B against the reverted anchor produced byte-identical version states for merge copies, so the second-merge-round anomaly is pre-existing `toSourceRev` behavior, not this PR; the "doc with >10k changes and zero main versions" population is structurally contradicted (fast-forward offline sessions and catch-up versioning both mint origin `'main'`), which matches the evidence pack's zero-doc sample; `parentId: undefined` never reaches a store write (`createVersion` spreads it conditionally); and the capture read/read/write race predates the PR — the old `getSnapshotAtRevision` path had the identical window, and the PR improves the artifact since overlapping versions now chain correctly.

Inline notes ordered by severity; nothing here blocks the core fix.

RyanHavoc:
Pushed bd93a89 addressing the review. What changed:

- **Truthful warn:** `loadParentState` only emits the "chaining to X" warn after the resolved parent's state and metadata actually loaded; every undefined exit for a version with `startRev > 1` (no parent found, recorded-or-resolved parent unloadable) now emits a "no usable parent; will replay the full history from rev 1" warn instead — so the previously silent populations (dead recorded parentId, branch-only sources in merges) are visible, and the detector can't claim a chaining that didn't happen.
- **Overlap guard:** a loaded parent whose `endRev > startRev - 1` is ignored with a warn and the build falls back to the full replay — this also stops the streaming fast path serving too-new bytes. Resolved parents can't overlap (the cursor arithmetic caps them), so this only catches bad recorded parentIds.
- **No redundant re-fetch:** the resolution path reuses the `VersionMetadata` that `findLatestMainVersion` returned and loads only the state — one fewer read per legacy-version build, and a transient `loadVersion` miss can no longer discard a resolution whose endRev was already in hand.
- **Query consolidation:** `handleOfflineSessionsAndBatches` now uses `findLatestMainVersion(store, docId, firstRev - 1)` instead of its hand-rolled copy of the same reversed-cursor query (and picks up the orphan clamp).

Five new tests cover the warn/overlap paths. Deliberately not changed, flagged for your call: the client `scrubTo` gap bridge (its own issue — main-origin sessions now have routinely non-contiguous parents and the client scrub trusts parent state directly), persisting resolved parentIds (blocked by `EditableVersionMetadata`; needs a backfill or an internal write-back decision), a `createVersion`-level parentId default (choke point for future writers; `mergeBranch` bypasses it either way), and the capture-path notes (unbounded tail read is pre-existing and self-heals on next commit; chaining does remove the accidental full-replay scrub of snapshot-only poison). 2485 tests, type:check, lint, and format:check green.

RyanHavoc (approved):

PR #105, opened 2026-07-13
fix(ot): refuse stale baseRev-0 continuations instead of replaying the log

jacwright:
## TL;DR

When a big upload got split into chunks, the server was mishandling every chunk after the first: it re-read the project's entire edit history (which failed on large projects and left people stuck on "syncing" forever), and it skipped the safety check that stops a stale device from overwriting newer work. This refuses those bad chunks cleanly and tells the device to reload instead. **It also closes a way a document could be silently wiped, which is possible on `main` today.**

Fixes the user-visible half of DAB-626. Read the "Three fixes I did not expect" section before merging.

## The bug

```ts
// src/algorithms/ot/server/commitChanges.ts:76 (before)
const batchedContinuation = batchId && changes[0].rev! > 1;
if (changes[0].baseRev === 0 && initialRev > 0 && !batchedContinuation) {
  // rebase + the root-op wipe guard both live in here
}
```

`breakChangesIntoBatches` mints a `batchId` **only when it splits** a flush, and all changes in the flush share one `baseRev`. So for a client at local rev 0 whose payload splits: batch 1 rebases and commits fine; batches 2..N read as continuations, skip the rebase, keep `baseRev: 0`, and `listChanges({startAfter: 0})` reads the whole log -> 413 -> retry the same precomputed batch forever.

Proven byte-exactly in prod: summing stored op bytes of `MGZeSdToRcwOr2ds` revs 1..13,624 gives 7,541,634, matching the logged read size exactly. The client sent `baseRev: 0`.

## Three fixes I did not expect

**1. Both proposed fixes were wrong, and I have receipts.**

- Keying on `initialRev === 0` (my first instinct) **breaks genuine first uploads**. Measured `initialRev` at batches 1/2/3 of a real new-doc upload: `[0, 2, 4]`. Batch 1 creates the head, so continuations never see `initialRev === 0`. It fails 3 tests including the pinned contract test `'should allow root creation when part of initial batch'`, and in prod it would return `docReloadRequired` between every batch of a growing doc.
- A continuity key (`initialRev === changes[0].rev - 1`) fixes the 413 but **does not close the data-loss hole**: a root `replace` at `rev = head + 1` with `baseRev 0` and a `batchId` still commits and wipes the doc, and that payload is arithmetically identical to the one the pinned test requires us to allow.

The only thing that separates a genuine continuation from a stale rev-0 flush is **the doc's own history**, so we ask it: `createdByBatch()` reads one indexed row, gated behind `batchId && baseRev === 0 && initialRev > 0`, so normal commits pay nothing.

**2. Rebasing these batches would have CORRUPTED documents.** The first attempt on this branch did exactly that, and A/B testing against `main` caught it. On `main`, batch 2's 413 **aborts** the flush loop, and that abort is load-bearing: batch 1's `docReloadRequired` already ran `_reloadDocFromServer` -> `reconcilePending`, the true client-side OT rebase, which turns a colliding `add /docs/d2` into a noop. Remove the 413 and the loop marches on, pushing stale precomputed batches that the server commits **verbatim** (the server "rebase" only re-stamps `baseRev`; `listChanges({startAfter: tip})` returns `[]` so the fast-forward branch saves the ops untransformed). Measured: `main` keeps `SERVER-CHAPTER-2/3`; the naive fix overwrote both with `CLIENT-2/3`.

So the server now **refuses** with a bounded `409 {scope:'doc'}` rather than rebasing. The refusal deliberately does **not** carry `docReloadRequired`, because the client treats that as "committed, drop from pending" (PatchesSync.ts:890-904, dw3 PatchesHubService.ts:1271-1280) and it would destroy unsynced work.

**3. There is a live document-wipe path on `main` today, and the 413 was not stopping it.** The root-op wipe guard was keyed on `baseRev === 0`. But after a reload, `OTAlgorithm._withConsistentBaseRev` re-stamps every pending change onto the new tip **without transforming its ops**. So a root `replace` rejected once at `baseRev 0` comes straight back at `baseRev = tip` and commits with the guard bypassed. 400 and 413 are both retryable (`TERMINAL_STATUS_CODES` = 401/402/403/404/410) and `OTAlgorithm` has no `ejectPendingChange`, so a 4xx on an OT doc never quarantines: it latches and retries, which is how the re-stamp gets its second chance.

Proven: `main` ends at `head = {WIPED: true}`. This branch keeps `{docs, title}`. The guard now runs on any commit to a doc that already has history, exempting only the upload that created it.

## Client half

`PatchesSync.flushDoc`: on `docReloadRequired`, stop sending the precomputed batches and re-enter sync (`void this.syncDoc(docId)`, the existing house idiom from `_tryEjectPoisonChange`; serial-gated, so exactly one follow-up queues rather than recursing). The follow-up flushes what the store actually holds, rebased by `reconcilePending` and re-stamped to head, so it carries `baseRev > 0` and can never answer `docReloadRequired` again.

## Verification

```
npm test        2476 passed | 4 skipped (109 files)
npm run type:check   pass
npm run lint         pass
npm run format:check pass
```

An A/B harness drove the **real** `commitChanges` + **real** `OTAlgorithm`/`OTInMemoryStore` + **real** `OTFuzzBackend` with a pup-like 413 cap, against `main`'s `commitChanges` extracted via `git show main:`. Four results:

1. FIXED preserves server content exactly as `main` does: `["CLIENT-1","SERVER-CHAPTER-2","SERVER-CHAPTER-3"]` on both.
2. FIXED issues no unbounded `baseRev`-0 read: zero errors, one clean flush (vs `main`'s 413).
3. A **stale, unpatched client** gets a bounded 409 instead of an unbounded 413 and reaches the same end state as `main`. Version-skew safe.
4. `main` wipes the doc on a root op; FIXED does not.

Tests were also inverted against the original buggy condition to confirm they actually fail on it.

## Deploy

Server half is safe with the clients already in users' browsers (result 3 above), so pup can ship first with no dw3 release. The client half only removes a wasted 409 round-trip.

## Relationship to #103

Complementary, no file conflict on the server side. #103 fixes client **recovery** (make a latched doc heal); this fixes the **cause** (stop pup emitting the 413). Both worth having. Note #103's 413 branch gates on `!pending?.length`, i.e. the read path pup **already** rescues via `serveSnapshotCatchup` — but every observed 413 event carries `has_pending: true` and therefore took the flush path, so as written that branch cannot fire on any of the real events. Suggested gate: `code === 413 && data?.scope === 'doc'`. Its `isNetworkError` half is a genuine independent bug and stands on its own.

**Heads up: this branch also touches `src/net/PatchesSync.ts`, which #103 touches. Whichever lands second will need a rebase.**

## Owner must know

1. **Audit the 22 affected docs.** Batch 1 of these rev-0 flushes **already commits today** (rebases onto head, returns `docReloadRequired`, client drops it from pending). The 413 never prevented that; it only stranded the tail. Someone should look at what batch 1 already wrote into those docs.
2. **Origin still UNKNOWN.** Neither the scout nor the implementer could establish what deposits a multi-megabyte rev-0 pending queue of nested-path ops. This fix is safe under that uncertainty because it makes committed content **identical to `main`'s** while removing the unbounded read and closing the wipe. It does not bet on what the ops are. Logging one live batch's op paths is still worth doing.
3. **Pre-existing bug, not fixed here:** `_reloadDocFromServer` calls `reconcilePending(docId, getChangesSince(docId, 0))`. On a rev-0 client that tail begins with the doc's root creation op, and rebasing pending against a root op **drops it**, so the client silently loses the tail of its own unsynced queue. `main` and this branch behave identically. This may be the real reason the 22 docs look torn. Deserves its own issue.

---

EVIDENCE_PACK
## Evidence pack

Everything below was measured **read-only** against production on 2026-07-13 (launch day of the v2.5 to v3.0 cutover). No writes of any kind. Reproduce it yourself before trusting it.

**Sources**

- Sentry issue `DABBLE-WRITER-3-DA` — `sync doc error latched: projects/{id}/content (413)`. Client-side, emitted from dw3 `src/services/repairTelemetry.ts:425`.
- Sentry alert: https://dabblewriter.sentry.io/alerts/rules/details/441891/
- Slack thread: https://dabblewriter.slack.com/archives/C0BFPA35GC8/p1783953299051229
- pup Cloud Logging, GCP project `dabble-writer`, Cloud Run service `pup`, revision `pup-00039-4fr`
- Firestore database `dabble-next` (prod), collection `projects_content`
- Linear: DAB-626

### 1. There are two 413s on two routes, not one

pup's store logs a structured warn every time it refuses an unbounded read. Field is `jsonPayload.msg` (not `.message` — a wrong field name here silently returns zero rows):

```bash
gcloud logging read \
  'resource.type="cloud_run_revision"
   resource.labels.service_name="pup"
   jsonPayload.msg=~"Refusing unbounded change history read"
   timestamp>="2026-07-13T00:00:00Z"' \
  --project=dabble-writer --limit=5000 --format=json > warns.json
```

A record looks like:

```json
{
  "jsonPayload": {
    "count": 15000,
    "docId": "projects/MwIDQtxEEFQ8JiVf/content",
    "opsBytes": 4579426,
    "maxUnboundedReadChanges": 10000,
    "maxUnboundedReadBytes": 67108864,
    "msg": "Refusing unbounded change history read: exceeds safe serving limits"
  },
  "severity": "WARNING",
  "timestamp": "2026-07-13T18:08:47.878694Z"
}
```

Correlating the HTTP 413s by route over the same window:

```
  _versions     4983 HTTP 413s    690 distinct docs      <- Bug A
  _changes        17 HTTP 413s     16 distinct docs      <- Bug B (what Sentry showed)
```

Store-level, across 5,000 sampled refusals: **712 distinct documents**.

### 2. The two bugs have different, mechanically distinguishable signatures

Group each doc's refused read `count` across its retries:

```
  read size PINNED  : 447 docs   count identical on every retry -> read starts at rev 0 and never moves
  read size GROWING :  41 docs   count tracks the tip -> a FIXED startAfter:0 against a MOVING head
```

`MGZeSdToRcwOr2ds` shows both, and its log is the whole story in six lines:

```
13:48:54  POST _changes   413   count=13614   opsBytes=6,274,710
13:57:09  POST _versions  413   count=13614   opsBytes=6,274,710
14:05:11  POST _versions  413   count=13614   opsBytes=6,274,710
14:07:24  POST _versions  413   count=13614   opsBytes=6,274,710
14:07:32  POST _changes   413   count=13624   opsBytes=7,541,634
14:37-16:55  POST _versions  413 x12   count=13614 every time   (dabble-rest webhook retry backoff)
```

`_versions` is **pinned at 13,614** across all 16 retries. `_changes` **grows with the tip** (13,614 then 13,624). Two different reads.

### 3. The commit read provably starts at `startAfter: 0`

Summing the stored `ops` byte length of `MGZeSdToRcwOr2ds` revs 1..13,624 in Firestore gives **7,541,634 bytes**, matching the logged read **to the byte**. Revs 5..13,628 would give 8,312,230, which does not match. The client sent `baseRev: 0`: it believed the document was empty.

`OMm6EH1W1E1ErML5` independently corroborates: three retries seconds apart, read count growing by exactly one each time (14,957 → 14,958 → 14,959) while another client committed. That is a fixed `startAfter: 0` against a moving tip, not a lagging client.

### 4. Versioning is NOT the problem, and the "version-less deadlock" theory is dead

An earlier writeup (including my own first pass, and the first version of the artifact) claimed these docs had no sealed version and were deadlocked. **That was wrong.** Sampling `projects_content` and computing the real span metric (`tipRev - lastEndRev` where `origin == 'main'`):

```
zero versions                     : 0  (0.00%)
zero versions AND >10k changes    : 0  (0.00%)
span percentiles: p50=0 p90=0 p99=0 max=47
tip  percentiles: p50=36 p90=1277 p99=20523 max=130734
```

The three incident docs are all healthy on this metric:

| Doc                | Changes (tip) | Versions | Last main endRev | **Span** |
| ------------------ | ------------: | -------: | ---------------: | -------: |
| `MGZeSdToRcwOr2ds` |        13,628 |       52 |           13,614 |   **14** |
| `GhQEuv6mpiFovu4A` |        23,578 |       44 |           23,574 |    **4** |
| `Cb1xCJT28vZX0YZY` |        16,393 |      150 |           16,379 |   **14** |

**The nuance that makes this coherent:** `02hhVxgdOE9x8lt1` has **130,734 changes, 1,012 versions, span of 1**. Versioning demonstrably works at enormous scale. That is because `catchUpVersions` (called inside `commitChanges`) uses `createVersionAtRev`, which **does** chain `parentId` correctly, and it keeps the watermark within `maxChangesPerVersion` (1000) of head. Only `captureCurrentVersion` — the 30-minute-inactivity auto-version, reached via `POST _versions` — fails to chain a parent. So the span metric stays healthy _because a different code path is doing the work_, while `POST _versions` 413s forever. The span metric was measuring the read that works. There was never a contradiction.

### 5. Corrections to earlier claims (do not propagate these)

- ~~"About 7.5% of active projects are already over 10k changes"~~ — measured total change count, which is **not** the risk metric (the cap keys on read _span_). Actual: ~2.17% of sampled `projects_content` docs exceed 10k total changes.
- ~~"GhQEuv6mpiFovu4A has ~15,000 changes"~~ — it has **23,578**.
- ~~"The affected projects have no sealed version and are deadlocked"~~ — they have 44 to 150 versions each. Dead.

### 6. Why nobody saw the bigger bug

`pup/src/onError.ts` gates Sentry capture on `status === 503` (~line 99) and `status >= 500 || isCorruptHistory` (~line 111). The oversized-history 413 is a plain `StatusError` with `name === 'Error'`, so it matches neither. **pup threw roughly 1,250 of these an hour, all day, and reported exactly zero.** Every Sentry event the team has is client-side, from dw3.

Separately, `PupServer.ts:930-936` notes that the JSON-RPC/WS commit path never passes through `onError` at all. The incident traffic was REST (`POST` routes), so `onError` is the right place for this fix, but the WS path is a known second gap.

### 7. Reproduction scripts

Read-only, in the session scratchpad: `span.mjs` (version span across `projects_content`), `bytes.mts` (per-rev stored-ops byte sum, proves `startAfter: 0`), `evidence-query.sh` (the log split above). All use ADC (`gcloud auth application-default print-access-token`) and the Firestore REST API. None writes.

### DIFF
--- a/src/algorithms/ot/server/commitChanges.ts
+++ b/src/algorithms/ot/server/commitChanges.ts
@@ -78,17 +78,22 @@ export async function commitChanges(
     // creation op its first batch carries, and the baseRev 0 its later batches keep. Once that
     // first batch lands, its later batches are indistinguishable on the wire from a never-synced
     // client flushing baseRev 0 onto a doc someone else wrote, so ask the doc's own history
-    // rather than the batch's revs. Resolved lazily: only these two rare shapes pay the read.
+    // rather than the batch's revs. Resolved lazily: normal commits (baseRev > 0, no root op)
+    // never pay the read; each continuation batch of a genuine multi-batch upload pays one.
     let ownUpload: Promise<boolean> | undefined;
     const isOwnUpload = () => (ownUpload ??= createdByBatch(store, docId, batchId));
-    const continuation = !!batchId && changes[0].rev! > 1;
+    // A flush starts at rev 1, so rev > 1 at baseRev 0 means the head of this queue was already
+    // resolved elsewhere — a continuation whether or not the flush split. Not keyed on batchId:
+    // an unsplit tail carries none, and it must not slip into the rebase heal below. Historical
+    // imports legitimately start chunks past rev 1 (they preserve their original revs).
+    const continuation = changes[0].rev! > 1 && !options?.historicalImport;
 
     // Prevent stale clients from wiping existing data with a root creation op anywhere in the
     // batch. Not keyed on baseRev: a reload re-stamps pending onto the new tip without
     // transforming its ops (OTAlgorithm._withConsistentBaseRev), so a root op re-sent after one
     // arrives at baseRev = tip and overwrites the doc just the same.
     const rootOpChange = changes.find(c => c.ops.some(op => op.path === ''));
-    if (rootOpChange && !(await isOwnUpload())) {
+    if (rootOpChange && !options?.allowRootReplace && !(await isOwnUpload())) {
       throw new StatusError(
         400,
         `Document ${docId} already exists (rev ${initialRev}). ` +
@@ -99,23 +104,6 @@ export async function commitChanges(
     }
 
     if (changes[0].baseRev === 0) {
-      // The first batch of this flush was answered with docReloadRequired, so these ops were
-      // minted against state the client has since replaced. Neither treatment is sound: rebasing
-      // moves baseRev to the tip, leaving the commit read nothing to transform against, so the
-      // ops stack verbatim onto a head they never saw and overwrite whatever paths they touch;
-      // transforming them honestly means reading the entire change log, which is unbounded and
-      // fails outright on a large doc. Refuse instead. The client's reload rebases what is left
-      // of the queue against the real history and re-sends it on the true head, so the work
-      // survives. Never docReloadRequired here: the client drops a batch from pending on that.
-      if (continuation && !(await isOwnUpload())) {
-        throw new StatusError(
-          409,
-          `Document ${docId} already exists (rev ${initialRev}) and was not created by batch ${batchId}. ` +
-            `Cannot continue a baseRev 0 upload onto it - these changes were made against state the client ` +
-            `has since reloaded. Reload the document and re-send them on the current revision.`,
-          { scope: 'doc' }
-        );
-      }
       if (!continuation) {
         // Rebase explicit baseRev: 0 on existing docs to current revision.
         // The caller signals docReloadRequired so the client knows to call getDoc.
@@ -124,6 +112,24 @@ export async function commitChanges(
         for (const c of changes) {
           c.baseRev = baseRev;
         }
+      } else if (!(await isOwnUpload())) {
+        // The head of this flush was already resolved (its first batch was answered with
+        // docReloadRequired), so these ops were minted against state the client has since
+        // replaced. Neither treatment is sound: rebasing moves baseRev to the tip, leaving the
+        // commit read nothing to transform against, so the ops stack verbatim onto a head they
+        // never saw and overwrite whatever paths they touch; transforming them honestly means
+        // reading the entire change log, which is unbounded and fails outright on a large doc.
+        // Refuse instead. The client reloads on this refusal (409 scope: 'doc'), which rebases
+        // what is left of the queue against the real history and re-sends it on the true head,
+        // so the work survives. Never docReloadRequired here: the client drops a batch from
+        // pending on that, and this batch was not committed.
+        throw new StatusError(
+          409,
+          `Document ${docId} already exists (rev ${initialRev}) and was not created by this upload. ` +
+            `Cannot continue a baseRev 0 upload onto it - these changes were made against state the client ` +
+            `has since reloaded. Reload the document and re-send them on the current revision.`,
+          { scope: 'doc' }
+        );
       }
     }
   }
--- a/src/algorithms/ot/server/commitChanges.ts
+++ b/src/algorithms/ot/server/commitChanges.ts
@@ -72,28 +72,59 @@ export async function commitChanges(
   const initialRev = await store.getCurrentRev(docId);
   let baseRev = changes[0].baseRev ?? initialRev;
 
-  // Check if this is a batched continuation (later part of an initial batch upload)
-  const batchedContinuation = batchId && changes[0].rev! > 1;
-
-  // Rebase explicit baseRev: 0 on existing docs to current revision.
-  // The caller signals docReloadRequired so the client knows to call getDoc.
   let docReloadRequired: true | undefined;
-  if (changes[0].baseRev === 0 && initialRev > 0 && !batchedContinuation) {
-    // Prevent stale clients from wiping existing data with a root creation op anywhere in the batch
+  if (initialRev > 0) {
+    // Both exemptions below belong to the multi-batch upload that CREATED this doc: the root
+    // creation op its first batch carries, and the baseRev 0 its later batches keep. Once that
+    // first batch lands, its later batches are indistinguishable on the wire from a never-synced
+    // client flushing baseRev 0 onto a doc someone else wrote, so ask the doc's own history
+    // rather than the batch's revs. Resolved lazily: only these two rare shapes pay the read.
+    let ownUpload: Promise<boolean> | undefined;
+    const isOwnUpload = () => (ownUpload ??= createdByBatch(store, docId, batchId));
+    const continuation = !!batchId && changes[0].rev! > 1;
+
+    // Prevent stale clients from wiping existing data with a root creation op anywhere in the
+    // batch. Not keyed on baseRev: a reload re-stamps pending onto the new tip without
+    // transforming its ops (OTAlgorithm._withConsistentBaseRev), so a root op re-sent after one
+    // arrives at baseRev = tip and overwrites the doc just the same.
     const rootOpChange = changes.find(c => c.ops.some(op => op.path === ''));
-    if (rootOpChange) {
+    if (rootOpChange && !(await isOwnUpload())) {
       throw new StatusError(
         400,
         `Document ${docId} already exists (rev ${initialRev}). ` +
-          `Cannot apply root-level replace (path: '') with baseRev 0 - this would overwrite the existing document. ` +
+          `Cannot apply root-level replace (path: '') - this would overwrite the existing document. ` +
           `Load the existing document first, or use nested paths instead of replacing at root.`,
         { changeId: rootOpChange.id, scope: 'change' }
       );
     }
-    docReloadRequired = true;
-    baseRev = initialRev;
-    for (const c of changes) {
-      c.baseRev = baseRev;
+
+    if (changes[0].baseRev === 0) {
+      // The first batch of this flush was answered with docReloadRequired, so these ops were
+      // minted against state the client has since replaced. Neither treatment is sound: rebasing
+      // moves baseRev to the tip, leaving the commit read nothing to transform against, so the
+      // ops stack verbatim onto a head they never saw and overwrite whatever paths they touch;
+      // transforming them honestly means reading the entire change log, which is unbounded and
+      // fails outright on a large doc. Refuse instead. The client's reload rebases what is left
+      // of the queue against the real history and re-sends it on the true head, so the work
+      // survives. Never docReloadRequired here: the client drops a batch from pending on that.
+      if (continuation && !(await isOwnUpload())) {
+        throw new StatusError(
+          409,
+          `Document ${docId} already exists (rev ${initialRev}) and was not created by batch ${batchId}. ` +
+            `Cannot continue a baseRev 0 upload onto it - these changes were made against state the client ` +
+            `has since reloaded. Reload the document and re-send them on the current revision.`,
+          { scope: 'doc' }
+        );
+      }
+      if (!continuation) {
+        // Rebase explicit baseRev: 0 on existing docs to current revision.
+        // The caller signals docReloadRequired so the client knows to call getDoc.
+        docReloadRequired = true;
+        baseRev = initialRev;
+        for (const c of changes) {
+          c.baseRev = baseRev;
+        }
+      }
     }
   }
 
@@ -316,6 +347,17 @@ export async function commitChanges(
 
   // Unreachable — the last iteration always re-throws
   throw new Error(`commitChanges: exhausted ${MAX_CONFLICT_RETRIES} retries for doc ${docId}`);
+}
+
+/**
+ * Whether `batchId` is the multi-batch upload that CREATED this doc, identified by the doc's
+ * oldest change. One row, and no `startAfter`, so a log that starts above rev 1 (pruned,
+ * migrated) reads as itself rather than as a gap.
+ */
+async function createdByBatch(store: OTStoreBackend, docId: string, batchId?: string): Promise<boolean> {
+  if (!batchId) return false;
+  const [oldest] = await store.listChanges(docId, { limit: 1 });
+  return oldest?.batchId === batchId;
 }
 
 /**
--- a/src/net/PatchesSync.ts
+++ b/src/net/PatchesSync.ts
@@ -880,6 +880,8 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
         }
       }
 
+      let reloadedMidFlush = false;
+
       for (const changeBatch of batches) {
         if (!this.state.connected || onlineState.isOffline) {
           throw new NetworkError('Disconnected during flush');
@@ -902,6 +904,14 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
           // Pass the batch as resolved so the pending-preserving import can't re-add it
           // from the open doc's stale in-memory queue.
           await this._reloadDocFromServer(docId, algorithm, false, changeBatch);
+          // `batches` predates the reload, which replaced the committed state they were minted
+          // against and rebased the rest of the queue onto the new head. Every remaining batch
+          // still holds the pre-rebase copy of its ops at baseRev 0, and the server cannot
+          // transform that onto a head it never saw: it stacks the ops verbatim over whatever
+          // paths they touch, or reads the whole change log trying to rebase them. Stop sending
+          // them and let the follow-up sync below flush what the store actually holds.
+          reloadedMidFlush = true;
+          break;
         } else {
           // Confirm sent first so server corrections (applied next) overwrite
           // any stale ops for fields the server won via LWW timestamp resolution.
@@ -939,6 +949,10 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
 
       const stillHasPending = await algorithm.hasPending(docId);
       this._updateDocSyncState(docId, { hasPending: stillHasPending, syncStatus: 'synced' });
+      // Send the remainder the reload rebased, from the store rather than from `batches`. Sync is
+      // serial-gated, so this queues exactly one follow-up pass instead of recursing, and that
+      // pass flushes on the reloaded committedRev, which no longer answers docReloadRequired.
+      if (reloadedMidFlush && stillHasPending) void this.syncDoc(docId);
     } catch (err) {
       if (this._isDocDeletedError(err)) {
         await this._handleRemoteDocDeleted(docId);
--- a/src/net/PatchesSync.ts
+++ b/src/net/PatchesSync.ts
@@ -776,11 +776,14 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
    * fetching/flushing: a contiguous tail past `committedRev` applies directly, a gap pulls
    * the authoritative tail. Loops because applying is async and more may buffer meanwhile;
    * the final empty check and the buffer removal (the caller's finally) share a microtask,
-   * so nothing slips between them. Must only run while the buffer entry still exists.
+   * so nothing slips between them. The entry can also vanish mid-drain: a docReloadRequired
+   * answered during a `flushPendingFirst` flush nests a second reload whose finally deletes
+   * it. Anything parked there is redelivered by the next catch-up, so a missing entry means
+   * nothing left to drain — never a crash.
    */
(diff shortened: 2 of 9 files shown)


src/algorithms/ot/server/commitChanges.ts
```diff
@@ -72,28 +72,59 @@ export async function commitChanges(
...
+    const isOwnUpload = () => (ownUpload ??= createdByBatch(store, docId, batchId));
+    const continuation = !!batchId && changes[0].rev! > 1;
+
+    // Prevent stale clients from wiping existing data with a root creation op anywhere in the
+    // batch. Not keyed on baseRev: a reload re-stamps pending onto the new tip without
+    // transforming its ops (OTAlgorithm._withConsistentBaseRev), so a root op re-sent after one
+    // arrives at baseRev = tip and overwrites the doc just the same.
     const rootOpChange = changes.find(c => c.ops.some(op => op.path === ''));
-    if (rootOpChange) {
+    if (rootOpChange && !(await isOwnUpload())) {
```
RyanHavoc: **Blocking: this breaks pup's document restore.** The guard now runs for any commit where `initialRev > 0`, and `createdByBatch` returns false immediately when the commit carries no `batchId` — but pup's restore flow commits exactly that shape: a single root replace at `baseRev = head` with no batchId ([restore.ts:357-359](https://github.com/dabblewriter/pup/blob/f05c983c27b99f104641376815c6a0b3fdb2b0d2/src/documents/restore.ts#L357-L359), driven by `scripts/restore-doc.ts`). On main this committed (the old guard was keyed on `baseRev === 0`; restore sends `baseRev = head > 0`); here every apply-mode restore of a doc with history 400s. Retried `historicalImport` chunks carrying the root creation have the same exposure — the guard runs before the resend dedup, so a redelivered already-committed chunk 400s instead of resolving as a resend. Needs a deliberate exemption: a server-originated bypass option, or restore passing something the guard recognizes.

src/algorithms/ot/server/commitChanges.ts
```diff
@@ -72,28 +72,59 @@ export async function commitChanges(
...
+      // The first batch of this flush was answered with docReloadRequired, so these ops were
+      // minted against state the client has since replaced. Neither treatment is sound: rebasing
+      // moves baseRev to the tip, leaving the commit read nothing to transform against, so the
+      // ops stack verbatim onto a head they never saw and overwrite whatever paths they touch;
+      // transforming them honestly means reading the entire change log, which is unbounded and
+      // fails outright on a large doc. Refuse instead. The client's reload rebases what is left
+      // of the queue against the real history and re-sends it on the true head, so the work
+      // survives. Never docReloadRequired here: the client drops a batch from pending on that.
+      if (continuation && !(await isOwnUpload())) {
+        throw new StatusError(
```
RyanHavoc: **The recovery this message prescribes is not implemented by any client.** 409 is not terminal ([error.ts:37-39](https://github.com/dabblewriter/patches/blob/4fa31179059c976db082e4a651c823c3b5763997/src/net/error.ts#L37-L39)), it is not `scope:'change'` so poison-eject can't fire, and nothing in `PatchesSync` reloads on it — reloads trigger only on `docReloadRequired`, 410, and `ApplyChangesError`. The new client survives only because its reload already ran in-session. The gap: `flushDoc` persists `dropResolvedPending` before `_reloadDocFromServer` ([PatchesSync.ts:902-906](https://github.com/dabblewriter/patches/blob/4fa31179059c976db082e4a651c823c3b5763997/src/net/PatchesSync.ts#L902-L906)), so a plain network failure in the reload's `getDoc` — no crash required — leaves committedRev 0 with a rev>1 queue. When that remainder still splits, every subsequent flush re-mints a batchId and lands here forever: retry ladder, error latch, reprobe, same 409; a collaborator's broadcast funnels back into the same flush too. main converged in this state (with the corruption documented in this PR), so the permanent wedge is new. A client hook for `409 + scope:'doc'` → reload-and-rebase (mirroring the eject idiom) closes it; until a client carrying that ships, this refusal strands that state.

src/algorithms/ot/server/commitChanges.ts
```diff
@@ -72,28 +72,59 @@ export async function commitChanges(
...
-    // Prevent stale clients from wiping existing data with a root creation op anywhere in the batch
+  if (initialRev > 0) {
+    // Both exemptions below belong to the multi-batch upload that CREATED this doc: the root
+    // creation op its first batch carries, and the baseRev 0 its later batches keep. Once that
+    // first batch lands, its later batches are indistinguishable on the wire from a never-synced
+    // client flushing baseRev 0 onto a doc someone else wrote, so ask the doc's own history
+    // rather than the batch's revs. Resolved lazily: only these two rare shapes pay the read.
+    let ownUpload: Promise<boolean> | undefined;
+    const isOwnUpload = () => (ownUpload ??= createdByBatch(store, docId, batchId));
+    const continuation = !!batchId && changes[0].rev! > 1;
```
RyanHavoc: Coverage note worth stating in the PR text: a stale rev-0 tail that fits a single wire batch carries no batchId at all (`breakChangesIntoBatches` only mints one when it splits — [changeBatching.ts:100-115](https://github.com/dabblewriter/patches/blob/4fa31179059c976db082e4a651c823c3b5763997/src/algorithms/ot/shared/changeBatching.ts#L100-L115)), so it reads as `continuation = false` here and takes the rebase heal, committing the stale nested-path ops verbatim — the outcome the 409 exists to refuse. Unchanged from main, and no legitimate in-repo flush produces `{baseRev: 0, rev > 1, no batchId}`, which also means keying `continuation` on `rev > 1` alone would close the gap. As written, "closes the data-loss hole" holds for root ops and split continuations only.

src/algorithms/ot/server/commitChanges.ts
```diff
@@ -72,28 +72,59 @@ export async function commitChanges(
...
+      if (continuation && !(await isOwnUpload())) {
+        throw new StatusError(
+          409,
+          `Document ${docId} already exists (rev ${initialRev}) and was not created by batch ${batchId}. ` +
+            `Cannot continue a baseRev 0 upload onto it - these changes were made against state the client ` +
+            `has since reloaded. Reload the document and re-send them on the current revision.`,
+          { scope: 'doc' }
+        );
+      }
+      if (!continuation) {
```
RyanHavoc, line 107: Two verified nits: (1) this `if (!continuation)` pairs with the `continuation && ...` throw above only by manual negation — `if (!continuation) { ...rebase... } else if (!(await isOwnUpload())) { throw 409 }` states the partition structurally and is behavior-identical in all input shapes. (2) Every continuation batch of a legitimate multi-batch first upload pays one `createdByBatch` read whose answer is immutable per doc (the memo at line 82 is per-request), so an n-batch upload pays n-1 reads for the same fact — the "only these two rare shapes pay the read" comment undersells that this is the feature's own happy path.

src/net/PatchesSync.ts
```diff
@@ -902,6 +904,14 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
           // Pass the batch as resolved so the pending-preserving import can't re-add it
           // from the open doc's stale in-memory queue.
           await this._reloadDocFromServer(docId, algorithm, false, changeBatch);
+          // `batches` predates the reload, which replaced the committed state they were minted
+          // against and rebased the rest of the queue onto the new head. Every remaining batch
+          // still holds the pre-rebase copy of its ops at baseRev 0, and the server cannot
+          // transform that onto a head it never saw: it stacks the ops verbatim over whatever
+          // paths they touch, or reads the whole change log trying to rebase them. Stop sending
+          // them and let the follow-up sync below flush what the store actually holds.
+          reloadedMidFlush = true;
```
RyanHavoc, line 933: Pre-existing bug on the recovery path this PR now leans on, worth its own issue: when this reload runs nested inside a hydration reload (syncDoc's flush-pending-first path — a change minted mid-fetch flips hasPending, flushDoc runs at baseRev 0, the server answers docReloadRequired), the nested `_reloadDocFromServer` clobbers the outer `_reloadBuffers` entry and deletes it in its finally; the outer drain then hits `this._reloadBuffers.get(docId)!` on undefined and throws TypeError ([PatchesSync.ts:782-783](https://github.com/dabblewriter/patches/blob/4fa31179059c976db082e4a651c823c3b5763997/src/net/PatchesSync.ts#L781-L784)), latching the doc and dropping the parked broadcasts. A re-entrancy guard on the buffer entry fixes it.

tests/algorithms/ot/server/commitChanges.spec.ts
```diff
@@ -198,6 +203,155 @@ describe('commitChanges', () => {
...
+
+      const err = await commitChanges(mockStore, 'doc1', [restamped], sessionTimeoutMillis).catch(e => e);
+
+      expect(err).toBeInstanceOf(StatusError);
+      expect(err.code).toBe(400);
+      expect(err.data).toEqual({ changeId: 'n', scope: 'change' });
+      expect(log.some(c => c.ops.some(op => op.path === ''))).toBe(false);
+    });
+
+    it('resolves an idempotent resend of a continuation batch as a resend', async () => {
```
RyanHavoc, line 339: This passes only because both commits hand-reuse `batchId: 'up'`. Real clients mint a fresh batchId on every flush and never persist it ([changeBatching.ts:108-115](https://github.com/dabblewriter/patches/blob/4fa31179059c976db082e4a651c823c3b5763997/src/algorithms/ot/shared/changeBatching.ts#L107-L116)), so a cross-flush resend arrives with a different batchId and `createdByBatch` classifies it as foreign — a lost-ack retry of a creating upload's first batch still 400s. That's identical on main (pre-existing, not a regression), but the exemption only actually covers batches of the same in-flight flush, and this test reads as broader coverage than exists.

tests/net/PatchesSyncStaleBatches.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+ */
+import { signal } from 'easy-signal';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { OTAlgorithm } from '../../src/client/OTAlgorithm.js';
+import { OTInMemoryStore } from '../../src/client/OTInMemoryStore.js';
+import { Patches } from '../../src/client/Patches.js';
+import { PatchesSync } from '../../src/net/PatchesSync.js';
+import type { Change } from '../../src/types.js';
+
+function makeConnection(overrides: Record<string, any> = {}) {
```
RyanHavoc: Fourth byte-identical copy of `makeConnection` in tests/net (races, poison-ejection, and reload-reconcile specs — md5-identical). Good moment to hoist it into a shared tests/net helper. Related prose nit: the "server can't transform baseRev-0 ops onto an unseen head" rationale is written out in full in five places (the commitChanges throw, the PatchesSync mid-flush comment, this file's header and final assertion, and the continuation test) — keeping the full story at the 409 throw with one-line pointers elsewhere will stop them drifting.

RyanHavoc (commented):
Second-model review, as requested in Slack. What held up under verification before the inline notes: the wipe-hole close is real (the re-stamped root-replace vector is blocked, and the A/B corruption claims could not be refuted); old deployed clients degrade to a single transient error flap after a mid-flush reload and converge on the next pass; and "Owner must know" item 3 is independently corroborated — the root-replace catch-up allowance exists only in `applyCommittedChanges`, not in the reload's `reconcilePending`/`rebaseChanges` path, so a rev-0 reload served a snapshot catch-up still drops pending.

Inline notes below, ordered by severity. The first two are the ones to resolve before merging.

RyanHavoc:
Pushed 7bcb725 addressing the review. What changed:

- **Restore breakage (blocking item 1):** added a `CommitChangesOptions.allowRootReplace` flag the root-op guard honors, for intentional server-originated whole-document replaces. Note for the pup side: `performRestore` needs to pass it once pup bumps to the release carrying this, and the REST/WS boundaries must strip it (and ideally `historicalImport`) from client-supplied options — it bypasses the wipe guard by design.
- **409 recovery (blocking item 2):** `PatchesSync.flushDoc` now catches a doc-scoped 409 from the commit, runs `_reloadDocFromServer` (nothing was committed or dropped, so the refused batch rebases with the rest of the queue), and queues the existing follow-up sync. This also covers the pre-existing `baseRev > server` 409, which prescribes the same reload. Test: `reloads and re-derives the queue when the server refuses a stale continuation (409)`.
- **Single-batch coverage gap:** `continuation` is now keyed on `rev > 1` instead of `batchId`, so a stale rev-0 tail that fits one wire batch (no batchId minted) is refused like a split tail instead of taking the rebase heal and committing verbatim. `historicalImport` is exempt since import chunks legitimately start past rev 1; tests pin both.
- **Nested-reload TypeError:** `_drainReloadBuffer` now tolerates the buffer entry vanishing (the docReloadRequired-during-hydration nesting), so that path degrades to redelivery-on-next-catch-up instead of latching the doc.
- **Cleanup:** the baseRev-0 branches are an if/else-if chain; the `createdByBatch` cost comment now states the real per-continuation-batch read; `makeConnection` is hoisted to `tests/net/connectionMock.ts` (fourth copy removed).

Deliberately not changed: the cross-flush batchId instability (the own-upload exemption still only covers batches of one in-flight flush — pre-existing, worth its own issue with the lost-ack creating-upload wedge), and the five-place transform-rationale prose (left to your taste). 2480 tests, type:check, lint, and format:check green.

RyanHavoc (approved):

PR #107, opened 2026-07-14
fix: type IndexedDB storage failures as StorageError

RyanHavoc:
**TL;DR:** Wrap WebKit's IndexedDB `UnknownError`/`QuotaExceededError` faults in a typed `StorageError` at the store boundary, with an `isStorageError` classifier — so "the browser can't persist locally" becomes a distinct, branchable failure instead of a generic transient that silently latches the doc. Enabling half of DAB-650; the UX lives in the dabble-writer-3.0 PR.

### Problem
WebKit throws `UnknownError` on IndexedDB put/delete/get under storage pressure/eviction (field reports: "Unable to store record in object store", "Failed to delete record from object store", "Attempt to get records… without an in-progress transaction"), and `QuotaExceededError` when the origin is full. None matched an existing guard — not `InvalidStateError` (the connection-closing reopen-retry), not `NetworkError`/`AbortError`/terminal `StatusError` — so `PatchesSync` treated them as generic retryable failures, ran the full ladder, then latched the doc with no way for the app to tell "browser can't save" apart from "server unreachable".

### Change
- `StorageError` class + `isStorageError()` classifier in `net/error.ts`, alongside `NetworkError`/`isAbortError`. Name-based so it survives the worker/structured-clone boundary and also matches the raw DOMException names.
- `toStorageError()` wraps the raw fault at every `IDBStoreWrapper` / `IDBTransactionWrapper` reject site (message + `cause` preserved); any non-storage error passes through untouched.
- Per the library scope boundary, patches only *types* the failure — the consuming app decides the UX.

### Tests
- `tests/net/error.spec.ts` — classifier + wrapper (connection-closing / abort / status correctly excluded).
- `tests/client/IndexedDBStore-storage-error.spec.ts` — put/get/transaction faults wrap to `StorageError`; `AbortError` passes through.
- Full client-store suite green (553); type/lint/format clean. Version bumped 0.19.4 → 0.19.5 (release still needs a manual `npm publish`).

Part of DAB-650.

### DIFF
--- a/src/client/IndexedDBStore.ts
+++ b/src/client/IndexedDBStore.ts
@@ -600,6 +600,12 @@ export class IDBTransactionWrapper {
     this.promise = new Promise((resolve, reject) => {
       tx.oncomplete = () => resolve();
       tx.onerror = () => reject(toStorageError(tx.error));
+      // A transaction that aborts at commit — notably under quota pressure — can fire ONLY
+      // `onabort` (with `tx.error` set) and no `onerror`, which would otherwise leave this
+      // promise pending forever. Reject on abort too, routing a storage-fault abort through the
+      // same typed StorageError; a plain user/teardown AbortError passes through untouched (and
+      // whichever of onerror/onabort fires first wins — the second reject is a no-op).
+      tx.onabort = () => reject(toStorageError(tx.error));
     });
   }
 
--- a/src/client/IndexedDBStore.ts
+++ b/src/client/IndexedDBStore.ts
@@ -8,6 +8,7 @@ import type {
   PatchesState,
 } from '../types.js';
 import { deferred, type Deferred } from '../utils/deferred.js';
+import { toStorageError } from '../net/error.js';
 import { signal } from 'easy-signal';
 import type { BranchClientStore } from './BranchClientStore.js';
 import type { PatchesStore, TrackedDoc } from './PatchesStore.js';
@@ -274,7 +275,7 @@ export class IndexedDBStore implements PatchesStore, BranchClientStore {
     await new Promise<void>((resolve, reject) => {
       const request = indexedDB.deleteDatabase(this.dbName!);
       request.onsuccess = () => resolve();
-      request.onerror = () => reject(request.error);
+      request.onerror = () => reject(toStorageError(request.error));
       request.onblocked = () => reject(request.error);
     });
   }
@@ -598,7 +599,7 @@ export class IDBTransactionWrapper {
     this.tx = tx;
     this.promise = new Promise((resolve, reject) => {
       tx.oncomplete = () => resolve();
-      tx.onerror = () => reject(tx.error);
+      tx.onerror = () => reject(toStorageError(tx.error));
     });
   }
 
@@ -627,7 +628,7 @@ export class IDBStoreWrapper {
     return new Promise((resolve, reject) => {
       const request = this.store.getAll(this.createRange(lower, upper), count);
       request.onsuccess = () => resolve(request.result);
-      request.onerror = () => reject(request.error);
+      request.onerror = () => reject(toStorageError(request.error));
     });
   }
 
@@ -636,7 +637,7 @@ export class IDBStoreWrapper {
       const index = this.store.index(indexName);
       const request = index.getAll(query);
       request.onsuccess = () => resolve(request.result);
-      request.onerror = () => reject(request.error);
+      request.onerror = () => reject(toStorageError(request.error));
     });
   }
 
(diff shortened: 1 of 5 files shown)


src/net/error.ts
```diff
@@ -125,6 +125,67 @@ export function isAbortError(err: unknown): boolean {
...
+}
+
+/**
+ * True for a storage-layer IndexedDB failure — see {@link StorageError}. Matches by name (not
+ * `instanceof`) so it classifies the same after crossing a structured-clone/worker boundary
+ * (which keeps only name/message/stack), AND recognizes the raw DOMException shape
+ * ({@link STORAGE_FAULT_NAMES}) on any path that didn't go through {@link toStorageError}.
+ */
+export function isStorageError(err: unknown): boolean {
+  if (!(err instanceof Error)) return false;
```
matalina: This `instanceof Error` guard (and the identical one in `toStorageError`) is the hinge of the whole change: `DOMException` only inherits from `Error` since ~Safari 12 / modern engines. On anything older, a raw WebKit `UnknownError` fails this check and passes through *unwrapped and unclassified* — silently defeating the fix on exactly the iPad-Safari target it's built for. The test env satisfies it (raw-DOMException tests pass) and any currently-supported Safari does too, so this is almost certainly fine — just worth a one-line confirmation against Dabble's actual minimum-supported Safari, since it's the non-obvious linchpin.

src/client/IndexedDBStore.ts
```diff
@@ -598,7 +599,7 @@ export class IDBTransactionWrapper {
     this.tx = tx;
     this.promise = new Promise((resolve, reject) => {
       tx.oncomplete = () => resolve();
-      tx.onerror = () => reject(tx.error);
+      tx.onerror = () => reject(toStorageError(tx.error));
```
matalina, line 602: The wrapper observes only `oncomplete`/`onerror`. A per-request storage fault bubbles here and gets typed correctly, but a transaction that aborts at *commit* under quota pressure can fire only `onabort` (with `tx.error` set) and no `onerror` — in which case this promise appears to neither resolve nor reject. It's a pre-existing gap and out of this PR's minimal typing scope, but it's directly adjacent to the `QuotaExceededError` faults you're now typing, so worth a follow-up: also wire `tx.onabort` and route `tx.error` through `toStorageError`.

matalina (approved):
Approving — tight, well-scoped, and correctly limited to "type the failure, let the app decide the UX." The exclusions (`InvalidStateError` reopen-retry, `AbortError` interruption, `StatusError`, `ConstraintError`/`DataError`) are all deliberate and documented, `toStorageError` is idempotent + total (safe at every reject site), and test coverage is excellent — classifier, wrapper, cross-boundary name copy, the specific WebKit fault messages, and null/already-wrapped pass-through.

No blockers. Three notes, inline where they anchor:

**1. Linchpin: `DOMException instanceof Error` (non-blocking, worth one confirmation).** Both `isStorageError` and `toStorageError` gate on `err instanceof Error`. `DOMException` only inherits from `Error` since ~Safari 12 / modern engines; on anything older, a raw WebKit `UnknownError` would fail the guard and pass through *unwrapped and unclassified* — silently defeating the fix on the exact iPad-Safari target. The vitest env clearly satisfies it (the raw-DOMException tests pass) and any currently-supported Safari does too, so almost certainly fine — just worth confirming against Dabble's actual browser floor since it's the non-obvious hinge.

**2. Transaction-level quota aborts may surface via `onabort`, not `onerror` (follow-up).** `IDBTransactionWrapper` wires only `oncomplete`/`onerror`. A per-request fault bubbles to `tx.onerror` (covered), but a transaction that aborts at commit under quota can fire only `onabort` with `tx.error` set — in which case `complete()` appears to neither resolve nor reject. Pre-existing gap, out of this PR's minimal scope, but directly adjacent to the faults you're now typing: some `QuotaExceededError` aborts could still slip through untyped.

**3. DB-`open` errors aren't wrapped.** Only request/transaction/`deleteDatabase` reject sites are covered. Likely intentional (open failures read as availability, not durability); a one-line note on why `open` is excluded would close the loop.

The 0.19.5 bump + manual publish coordination with dw3 #1041 is noted in the description. Nice work.

RyanHavoc:
Thanks — all three addressed in the follow-up commit:

1. **Linchpin (`DOMException instanceof Error`)** — good catch; the code contradicted its own "by name, not instanceof" doc. `isStorageError` and `toStorageError` are now purely name-based (no `instanceof Error` gate), so a pre-Safari-12 raw `UnknownError` (which doesn't inherit `Error`) still classifies + wraps. Added a test with a non-`Error`, DOMException-shaped object.
2. **`onabort` (commit-time quota abort)** — wired `tx.onabort → reject(toStorageError(tx.error))` so an abort no longer leaves `complete()` pending, and a quota abort routes through the typed error. Added a test.
3. **DB-`open` scope** — added a note on `StorageError` that it types per-request/transaction *durability* faults; `open` failures are *availability* (private mode / blocked / failed upgrade), owned by the open path + the app's startup persistence probe, and intentionally out of scope.

50 tests green; type/lint/format clean. (The `open`-error follow-up you flagged as pre-existing I've left out of this minimal PR.)

PR #108, opened 2026-07-14
fix: guard version state reads against missing blobs

jacwright:
**TL;DR:** If a version's saved snapshot ever goes missing from storage, Dabble now tells the app "try again shortly" instead of silently opening the document as if it were empty. This closes a path where a storage hiccup could look like a wiped project, and it lays the groundwork for pup building version snapshots in the background.

## What changed

**Reads of version state fail loud instead of serving an empty doc.** `getSnapshotAtRevision` previously fell back to `rawState ?? 'null'` when the latest main version's state blob was missing, which served the document as empty at that revision (doc-wipe-shaped, silent). It now throws a retryable `StatusError(503)` with `{ docId, versionId }`. A zero-byte blob (`''`) is treated the same as missing.

**`PatchesHistoryManager.getVersionState` distinguishes the failure modes:**

- state loads → stream it (unchanged)
- version metadata missing → 404
- metadata exists but state unavailable → retryable 503
- a `StatusError` thrown by the store (e.g. a backend's "build in progress, retry later") passes through unchanged instead of being wrapped into a generic non-retryable error

**`loadParentState` walks the `parentId` chain.** When a version's direct parent has no state (recorded-but-not-yet-built, or lost), the base-state build now walks up to `MAX_ANCESTOR_HOPS` (10) ancestors to the nearest one with state before falling back to replay-from-rev-1, and degrades gracefully (warn + replay) if the walk itself fails. Previously a stateless parent always meant a full replay inside an awaited commit.

503 is not in `TERMINAL_SYNC_CODES`, so clients treat it as transient and retry.

## Why now

This is half of the deferred version-build effort (see the companion pup PR, dabblewriter/pup#172): with builds running off the main thread, "metadata exists, state not yet written" becomes a normal transient state, and every read path must handle it as retryable rather than serving `null`. The lost-blob guard is independently valuable and ships first.

## Deliberate behavior change

A doc whose latest main version metadata exists but whose GCS state blob is truly lost now returns 503 on cold load instead of silently serving as empty. That converts invisible corruption into visible, retryable errors. Recommended before deploying consumers: audit prod for already-lost blobs so the flip from silent to loud is measured (any doc hitting this today was already serving wrong data).

## Testing

- 2522 tests pass (new coverage: 503 on missing/empty state, 404 on nonexistent version, StatusError passthrough, ancestor-walk bound, walk-failure degradation, single-warn consolidation)
- type:check, lint, format:check clean

### DIFF
--- a/src/algorithms/ot/server/buildVersionState.ts
+++ b/src/algorithms/ot/server/buildVersionState.ts
@@ -1,5 +1,7 @@
+import { isStatusError } from '../../../net/error.js';
 import { jsonReadable, parseVersionState } from '../../../server/jsonReadable.js';
 import type { OTStoreBackend } from '../../../server/types.js';
+import { isMissingVersionState } from '../../../server/utils.js';
 import type { Change, PatchesState, VersionMetadata } from '../../../types.js';
 import { applyChanges, applyChangesForReconstruction, type ReplayOptions } from '../shared/applyChanges.js';
 import { findLatestMainVersion } from './getSnapshotAtRevision.js';
@@ -56,20 +58,36 @@ async function loadParentState(
   }
 
   // Parent metadata without state: walk the ancestor chain to the nearest version with state.
-  // Missing is any falsy read (a zero-byte blob comes back '', never a valid state). A hop that
-  // fails transiently degrades to the replay fallback rather than failing the read; the replay
-  // runs off the change log, a separate subsystem from the state blobs.
-  const statelessParentId = parentMeta && !rawState ? parentMeta.id : undefined;
+  // Missing is any falsy read (a zero-byte blob comes back '', never a valid state — see
+  // isMissingVersionState). Each hop prefers the recorded parentId, but an ancestor that never
+  // recorded one (startRev > 1) is resolved the same way the entry branch resolves the version's
+  // own missing parent — otherwise a single stateless link with no parentId would drop straight
+  // to full-history replay even with an intact earlier version right below it. A hop that fails
+  // transiently degrades to the replay fallback (the replay runs off the change log, a separate
+  // subsystem from the state blobs); a StatusError from the store is an authoritative retry
+  // signal and propagates instead, so a deferred backend's "build in progress" reaches the client.
+  const statelessParentId = parentMeta && isMissingVersionState(rawState) ? parentMeta.id : undefined;
   try {
-    for (let hops = 0; parentMeta && !rawState && parentMeta.parentId && hops < MAX_ANCESTOR_HOPS; hops++) {
-      [rawState, parentMeta] = await loadPair(parentMeta.parentId);
+    for (let hops = 0; parentMeta && isMissingVersionState(rawState) && hops < MAX_ANCESTOR_HOPS; hops++) {
+      if (parentMeta.parentId) {
+        [rawState, parentMeta] = await loadPair(parentMeta.parentId);
+      } else if (parentMeta.startRev > 1) {
+        // Resolved metadata is already in hand, so only its state needs loading — no re-fetch.
+        const resolved = await findLatestMainVersion(store, docId, parentMeta.startRev - 1);
+        if (!resolved) break;
+        parentMeta = resolved;
+        rawState = await store.loadVersionState(docId, resolved.id);
+      } else {
+        break; // reached the rev-1 origin with no state; fall through to the replay fallback
+      }
     }
   } catch (error) {
+    if (isStatusError(error)) throw error;
     console.warn(`Ancestor walk failed for version ${version.id} of doc ${docId}; replaying from rev 1.`, error);
     return;
   }
 
-  if (parentMeta && rawState) {
+  if (parentMeta && !isMissingVersionState(rawState)) {
     // A parent whose snapshot extends into this version's own range can't be chained from: its
     // state already contains revs at or past startRev, so bridging (or fast-path streaming)
     // from it would serve too-new state. Resolved parents can't overlap — findLatestMainVersion
--- a/src/algorithms/ot/server/buildVersionState.ts
+++ b/src/algorithms/ot/server/buildVersionState.ts
@@ -15,33 +15,39 @@ function replayChanges<T>(state: T, changes: Change[], options?: ReplayOptions):
 }
 
 /**
+ * How far up the parent chain to look for an ancestor with state before giving up and
+ * replaying from rev 1: keeps a poisoned or unbuilt chain a bounded read.
+ */
+export const MAX_ANCESTOR_HOPS = 10;
+
+/**
  * Loads the state of the version `version` chains from, and the rev it ends at.
  *
  * A version that starts past rev 1 with no recorded `parentId` is a bug in whatever wrote it:
  * every base state below is bridged from `parent.endRev`, so without a parent the bridge runs
- * from rev 0 and replays the document's *entire* history — unbounded, and on a large document
- * expensive enough to fail the read outright. Rather than do that silently, resolve the parent
+ * from rev 0 and replays the document's *entire* history (unbounded, and on a large document
+ * expensive enough to fail the read outright). Rather than do that silently, resolve the parent
  * the writer should have recorded. The state it yields is identical; the read is bounded.
+ * A parent whose metadata exists but whose state is missing (a lost blob, or a deferred build
+ * that hasn't landed) gets the same treatment through its own `parentId` chain, bounded by
+ * {@link MAX_ANCESTOR_HOPS}.
  *
- * Returns undefined when there is no usable parent to chain from — no earlier version exists,
- * the parent (recorded or resolved) can't be loaded, or the parent's snapshot overlaps this
- * version's range — leaving the caller to build the base by replaying from rev 1. Any of those
- * on a version past rev 1 warns, since that replay is the unbounded read chaining exists to avoid.
+ * Returns undefined when there is no usable parent to chain from, leaving the caller to build
+ * the base by replaying from rev 1; on a version past rev 1 that warns, since the replay is
+ * the unbounded read chaining exists to avoid.
  */
 async function loadParentState(
   store: OTStoreBackend,
   docId: string,
   version: VersionMetadata
 ): Promise<{ rawState: string | ReadableStream<string>; endRev: number } | undefined> {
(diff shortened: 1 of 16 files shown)


src/algorithms/ot/server/buildVersionState.ts
```diff
@@ -14,42 +14,62 @@ function replayChanges<T>(state: T, changes: Change[], options?: ReplayOptions):
...
   }
 
-  if (parentMeta && rawState !== undefined) {
+  // Parent metadata without state: walk the ancestor chain to the nearest version with state.
+  // Missing is any falsy read (a zero-byte blob comes back '', never a valid state). A hop that
+  // fails transiently degrades to the replay fallback rather than failing the read; the replay
+  // runs off the change log, a separate subsystem from the state blobs.
+  const statelessParentId = parentMeta && !rawState ? parentMeta.id : undefined;
+  try {
+    for (let hops = 0; parentMeta && !rawState && parentMeta.parentId && hops < MAX_ANCESTOR_HOPS; hops++) {
```
RyanHavoc: The walk requires `parentMeta.parentId`, so a stateless parent that never recorded one (startRev > 1) exits the loop on entry and the read degrades to full-history replay — even when an earlier main version with intact state exists. The entry branch above (L54-L57) already proves the recovery for exactly this shape: when `parentId` is absent mid-walk, `findLatestMainVersion(store, docId, parentMeta.startRev - 1)` could resolve the link the writer should have recorded. Current writers always chain `parentId`, so this only bites legacy/imperfect records — but those are the shapes this PR exists to tolerate.

https://github.com/dabblewriter/patches/blob/5b0af8fe6cbb8745e011f5866df3ad184934eb4d/src/algorithms/ot/server/buildVersionState.ts#L63-L66

src/algorithms/ot/server/getSnapshotAtRevision.ts
```diff
@@ -36,13 +37,25 @@ export async function findLatestMainVersion(
...
+ */
 async function getLatestMainVersion(store: OTStoreBackend, docId: string, beforeRev?: number) {
   const version = await findLatestMainVersion(store, docId, beforeRev);
-  return {
-    rawState: version ? await store.loadVersionState(docId, version.id) : undefined,
-    versionRev: version?.endRev ?? 0,
-  };
+  if (!version) return { rawState: undefined, versionRev: 0 };
+  const rawState = await store.loadVersionState(docId, version.id);
+  if (!rawState) {
```
RyanHavoc: `!rawState` catches `undefined` and `''`, but `loadVersionState` may also return a `ReadableStream`, and a zero-byte blob surfaced as an empty stream is truthy — it sails past this guard, so `getSnapshotStream` emits `{"state":,...}` (invalid JSON) and `getSnapshotAtRevision` throws a raw `JSON.parse` SyntaxError instead of the retryable 503 this PR adds. The same gap exists at the `getVersionState` check (https://github.com/dabblewriter/patches/blob/5b0af8fe6cbb8745e011f5866df3ad184934eb4d/src/server/PatchesHistoryManager.ts#L107) and the walk trigger (https://github.com/dabblewriter/patches/blob/5b0af8fe6cbb8745e011f5866df3ad184934eb4d/src/algorithms/ot/server/buildVersionState.ts#L62), where failure paths also drop fetched streams without `cancel()`. Theoretical today — every known backend materializes strings — but the contract explicitly invites streaming implementations. Worth handling the empty-stream case, or narrowing the documented contract to require zero-byte blobs come back as `''`/`undefined`.

https://github.com/dabblewriter/patches/blob/5b0af8fe6cbb8745e011f5866df3ad184934eb4d/src/algorithms/ot/server/getSnapshotAtRevision.ts#L51-L58

src/algorithms/ot/server/buildVersionState.ts
```diff
@@ -14,42 +14,62 @@ function replayChanges<T>(state: T, changes: Change[], options?: ReplayOptions):
...
+      [rawState, parentMeta] = await loadPair(parentMeta.parentId);
+    }
+  } catch (error) {
```
RyanHavoc, line 84: This catch also swallows a store-thrown `StatusError` — `getVersionState` deliberately rethrows those (and the new test blesses stores failing reads with a 503), but a hop that throws one here degrades to the full replay instead of propagating the retry signal. `if (error instanceof StatusError) throw error;` before the warn would align the two paths.

https://github.com/dabblewriter/patches/blob/5b0af8fe6cbb8745e011f5866df3ad184934eb4d/src/algorithms/ot/server/buildVersionState.ts#L66-L71

src/server/PatchesHistoryManager.ts
```diff
@@ -89,21 +90,31 @@ export class PatchesHistoryManager {
...
+      if (error instanceof StatusError) throw error;
       console.error(`Failed to load state for version ${versionId} of doc ${docId}.`, error);
       throw new Error(`Could not load state for version ${versionId}.`, { cause: error });
     }
+    if (rawState) {
+      return typeof rawState === 'string' ? jsonReadable(rawState) : rawState;
+    }
+    // Missing state (undefined, or '' from a zero-byte blob) is not servable as a document:
+    // distinguish a version that doesn't exist from one whose state isn't available.
+    const version = await this.store.loadVersion(docId, versionId);
```
RyanHavoc: This call sits outside the try/catch above, so a transient store failure during the 404-vs-503 disambiguation escapes raw: no `console.error`, no wrap, and `JSONRPCServer.processMessage` serializes a code-less error as `-32000` with `err.stack` as data — the raw store message and stack reach the client, unlike every other failure mode of this method. Moving it into the try (or its own guarded block) keeps the logging/wrapping contract.

https://github.com/dabblewriter/patches/blob/5b0af8fe6cbb8745e011f5866df3ad184934eb4d/src/server/PatchesHistoryManager.ts#L111-L114

src/server/PatchesHistoryManager.ts
```diff
@@ -89,21 +90,31 @@ export class PatchesHistoryManager {
...
     try {
-      const rawState = await this.store.loadVersionState(docId, versionId);
-      if (rawState === undefined) {
-        return jsonReadable('null');
-      }
-      // rawState is already string or ReadableStream<string> — wrap string if needed
-      return typeof rawState === 'string' ? jsonReadable(rawState) : rawState;
+      rawState = await this.store.loadVersionState(docId, versionId);
     } catch (error) {
+      if (error instanceof StatusError) throw error;
```
RyanHavoc: `instanceof` only matches this package's `StatusError` class. pup's `FirestoreStore.loadVersionState` throws pup's own `StatusError` (404 from `assertVersioningEnabled`, 500 from `verifyStateChecksum`), which fails this check deterministically and gets wrapped generic. Not a regression — pre-PR everything was wrapped — but the passthrough would actually fire for the real deployment if it duck-typed on a numeric `code` the way `isRejectionError` in `net/error.ts` does, for exactly this cross-package hazard.

https://github.com/dabblewriter/patches/blob/5b0af8fe6cbb8745e011f5866df3ad184934eb4d/src/server/PatchesHistoryManager.ts#L102-L104

RyanHavoc (commented):
Findings from a review pass. Seven inline comments on the diff; three more that don't anchor to changed lines:

**docs/OTServer.md:298-299 states the superseded contract.** It still says version state is built "inline or queued — and must throw if state creation fails," but this PR's rewritten contract in `types.ts` says deferred implementations must NOT throw at `createVersion` (failures surface later as a 503 on read). A store author following this doc will implement the old contract. https://github.com/dabblewriter/patches/blob/5b0af8fe6cbb8745e011f5866df3ad184934eb4d/docs/OTServer.md#L297-L300

**`getStateBeforeVersion` still throws a bare `Error` for a missing version.** This PR establishes a `StatusError(404)` contract on `getVersionState` for exactly that condition, and both methods are RPC-exposed as `read` in the same ApiDefinition — one endpoint now 404s while its sibling surfaces a generic error. One-line polish for a consistent contract. https://github.com/dabblewriter/patches/blob/5b0af8fe6cbb8745e011f5866df3ad184934eb4d/src/server/PatchesHistoryManager.ts#L140-L144

**No version bump.** `package.json` is still 0.19.5, identical to main (already published), so consumers can't pick this up after merge without a bump. Possibly deliberate since the repo mixes in-commit and standalone bumps — flagging in case it isn't. https://github.com/dabblewriter/patches/blob/5b0af8fe6cbb8745e011f5866df3ad184934eb4d/package.json#L3

src/server/types.ts
```diff
@@ -47,7 +50,9 @@ export interface VersioningStoreBackend {
 
   /**
    * Loads the state snapshot for a specific version ID.
-   * Returns a JSON string, ReadableStream of JSON chunks, or undefined if not found.
+   * Returns a JSON string, ReadableStream of JSON chunks, or undefined if not found,
```
RyanHavoc: The contract still documents only `undefined` as the missing shape, but the three new guards each independently treat `''` as missing too (a zero-byte blob) — and the pre-PR `getVersionState` checked `rawState === undefined` per this doc, which is how the original bug slipped in. Documenting the zero-byte case here and sharing one `isMissingVersionState(raw)` predicate across the three call sites would keep a future fourth caller from reintroducing it.

https://github.com/dabblewriter/patches/blob/5b0af8fe6cbb8745e011f5866df3ad184934eb4d/src/server/types.ts#L52-L58

src/algorithms/ot/server/buildVersionState.ts
```diff
@@ -14,42 +14,62 @@ function replayChanges<T>(state: T, changes: Change[], options?: ReplayOptions):
     : applyChanges(state, changes);
 }
 
+/**
+ * How far up the parent chain to look for an ancestor with state before giving up and
+ * replaying from rev 1: keeps a poisoned or unbuilt chain a bounded read.
+ */
+export const MAX_ANCESTOR_HOPS = 10;
```
RyanHavoc, line 23: Minor: exceeding this bound abandons the walk into the full replay it exists to avoid — 11+ consecutive stateless versions (mass blob loss, or an out-of-order deferred builder) walks 10 cheap hops and then replays from rev 1 anyway. Fine as cycle protection for bad recorded parentIds; if the deferred-build backlog can ever exceed this, a visited-set (cycle-safe, unbounded hops) may serve better. Not blocking.

https://github.com/dabblewriter/patches/blob/5b0af8fe6cbb8745e011f5866df3ad184934eb4d/src/algorithms/ot/server/buildVersionState.ts#L17-L22

PR #111, opened 2026-07-16
fix: keep the ancestor walk reachable for overlapping stateless parents

jacwright:
**TL;DR:** No user-facing change. Fixes a regression in #110 (unreleased): version history for documents with a specific kind of damaged version bookkeeping would have been rebuilt the slow, expensive way instead of the fast way. Also restores a small read-speed optimization #110 accidentally dropped. Ship this in 0.19.8 before publishing.

Two findings from the round-5 review of the pup deferred-builds branch, both in #110's `resolveBuildParent`:

- **Overlap rejection killed the ancestor walk.** The resolver rejected an overlapping recorded parent on metadata alone, before its state was known — so an overlapping parent whose state is ALSO missing never entered `loadParentState`'s ancestor walk, and the build dropped to full-history replay from rev 0 even with a clean stateful ancestor one hop up (the pre-#110 walk handled exactly this; the fleet has produced this metadata class before, per the orphan-version heal). The resolver now returns the parent and leaves usability to the build: `loadParentState`'s final guard still rejects an overlapping snapshot after the walk, exactly as before #110. This also restores the single overlap warn.
- **Serialized parent loads.** Routing the recorded-parent entry through the metadata-only resolver made the Firestore metadata read complete before the GCS state read started — one extra round trip on every chained build and every `getStateBeforeVersion` read. The entry inlines the resolver's one-line first branch again so the loads stay parallel.

pup's chain walk doesn't need the overlap rejection either: as of pup#172's latest commits a poison or parked ancestor no longer cascades, so an overlapping pending parent is just a parent that gets built first.

Tests: new pin for the overlapping-stateless-parent walk; the direct-overlap tests are back to one warn; the resolver overlap test now asserts the parent is returned. 2534 passing, type/lint/format clean.

### DIFF
--- a/src/algorithms/ot/server/buildVersionState.ts
+++ b/src/algorithms/ot/server/buildVersionState.ts
@@ -26,9 +26,11 @@ export const MAX_ANCESTOR_HOPS = 10;
  * Resolves the version a build of `version` will chain from: the recorded `parentId`, else the
  * latest main version ending before `startRev` (bounded by the committed change log — see
  * {@link findLatestMainVersion}, so an orphan version stamped past the log is never selected).
- * A recorded parent whose snapshot overlaps the version's own range (`endRev` past
- * `startRev - 1`) already contains revs the build must exclude, so it is unusable and resolves
- * to `undefined` — the build replays from rev 1 instead.
+ *
+ * Resolution only — an overlapping recorded parent (`endRev` past `startRev - 1`) is still
+ * returned. Whether its snapshot is usable is the build's concern: `loadParentState` rejects it
+ * after the ancestor walk, so a stateless overlapping link can still be walked past to a clean
+ * ancestor instead of forcing a full-history replay.
  *
  * Metadata only, no state reads — cheap enough for dependency walks. Exported for
  * deferred-build backends: a builder draining pending versions must order builds with the same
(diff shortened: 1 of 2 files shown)


jacwright:
Verified against the real source, not just the diff.

**Both fixes confirmed correct and faithful to pre-#110 behavior:**

1. **Walk reachability.** Diffing `dab3676^` (pre-#110) against this branch, `loadParentState` is restored byte-for-byte apart from the `resolveBuildParent` extraction. #110 had routed the recorded-parent case through the resolver, whose overlap rejection returned `undefined` and skipped the ancestor walk entirely, dropping an overlapping-but-stateless parent to full-history replay from rev 0. The new inlined `loadPair` reaches the walk, and the post-walk guard in `loadParentState` still rejects an overlapping *snapshot* — so the direct-overlap-with-state case still correctly falls to replay (identical state output, now a single consolidated warn).

2. **Parallel loads.** #110's `await resolveBuildParent` + `await loadVersionState` was two serial store round trips; the restored `loadPair` (`Promise.all`) puts the metadata and state reads back in parallel on every chained build and `getStateBeforeVersion` read.

Tests: the new `walks past a stateless parent whose metadata overlaps` pins the exact regression (traced against `cleanLog` → chains from `v-gp`, `readsFrom []`, no replay); the direct-overlap tests correctly assert one warn. 2534 pass locally, typecheck / eslint / prettier all clean.

**Non-blocking observation:** the fix covers the stateless-overlapping parent. An overlapping parent that *has* a (too-new) snapshot still drops to full replay even when a clean stateful ancestor sits one hop up, because the walk only triggers on missing state. That's pre-existing, not a regression here — noting it since the fleet has produced overlapping metadata before and the stateful variant of the same class would take the slow path.

The `resolveBuildParent` contract change (returns overlapping parents) is safe in-repo: the only internal caller no longer routes the recorded-parent case through it, and the resolver is otherwise a metadata-only public export for pup, where an overlapping pending parent should be built first. pup#172's no-cascade guarantee isn't checked out locally so I took that cross-repo claim as-is.

Approving. Good to ship in 0.19.8.

PR #112, opened 2026-07-16
feat(ot): poison-pill ejection for the OT algorithm

RyanHavoc:
## TL;DR

Poison-pill ejection — move a rejected pending change into quarantine, sync the rest past it, keep its content recoverable — was **LWW-only**. OT docs latched forever on a change-scoped rejection. This implements it for OT.

The deferral was premised on there being no live emitter of a change-scoped OT rejection. That premise no longer holds: a consuming app server rejects on policy — Dabble's Pup rejects an OT content write from a role that may not make it, attaching `{ changeId, scope: 'change' }` — which strands the rejected change at the head of the queue so nothing behind it ever syncs. (Downstream: this is what silently swallowed beta readers' review-copy comments.)

Reviewed adversarially before opening (see **Correctness** below); no data-corruption or non-convergence bug found. All findings were low-severity and are addressed in the diff.

## The problem ejection has to solve

An OT pending queue is a *sequential program*: each change's ops are expressed in the frame the changes before it produce. So removing one change from the middle can't be a splice — its successors were built on top of it and must be rebased into the frame that skips it.

`computePendingEjection` (`src/algorithms/ot/shared/ejectPendingChange.ts`) inverts the ejected change against the state it applied to (committed + predecessors) and walks that inverse through the successors using the **same one-sided diamond `rebaseChanges` already runs** for an incoming server change — the ejected change genuinely preceded them, so its inverse is the "already-happened" side their position ties yield to. Predecessors are untouched; survivors renumber contiguously off `committedRev`, preserving the OT pending invariant (`baseRev === committedRev`, sequential revs). It reuses the exact transform primitives (`rebaseChanges` → `transformPatch`, `invertPatch`) normal OT sync already depends on, so it adds no new transform-correctness surface.

## What's in the diff

- **`computePendingEjection`** — the pure rebase core, standalone and unit-tested.
- **`OTAlgorithm`**: `verifyPendingChange` / `ejectPendingChange` / `listQuarantinedChanges` / `discardQuarantinedChange`, mirroring the LWW contract. `verify` probes in the change's own frame (committed + predecessors, not committed-only); a cleanly-applying **policy** rejection returns true, so PatchesSync latches for app consent rather than auto-ejecting.
- **OT stores**: `quarantinePendingChange` (atomic quarantine-write + pending swap in one transaction) plus list/discard, over the shared `quarantinedChanges` store. Preserved across untrack, cleared on delete; external-mode absence fails safe to the latch.
- `Patches.ejectPendingChange` and `PatchesSync` already forward generically — the public API and auto-eject path light up for OT with **no change there**.
- `docs/quarantine.md` updated (no longer "LWW only").

## Correctness

Proven against the **real OT server** (`commitChanges` + `OTFuzzBackend`), not mocks:

- **200 randomized pending programs**, each committed cleanly and converged after a random ejection (server head == client belief; contiguous revs; ejected id absent).
- A **disjoint-preservation** oracle: ejecting a change on a disjoint subtree leaves every survivor byte-identical (the generalized Doug shape).
- Deterministic unit tests for every shape: disjoint / overlapping `@txt` / head / middle / only-change / invert-failure / corrupt-predecessor.
- Algorithm + store lifecycle: idempotent double-eject, unknown id, untrack-preserves / delete-clears, invert-failure leaves the queue untouched (never half-rebased).

Full suite (2946) + OT fuzz convergence (73) green; type-check, lint, format, build clean.

### Review findings addressed
- `verifyPendingChange` now returns a clean boolean when a *predecessor* is un-appliable (fail toward `true` = can't corroborate = don't auto-eject), instead of throwing.
- Doc softened: `newPending` is *a* valid poison-free queue the server converges on, not provably the unique "never minted" one (the same tie-resolution caveat every OT rebase carries).
- Doc caveat added: a successor scoped to structure the ejected change *created* transforms away and is dropped — only the ejected change itself is preserved.

## Follow-ups (not in this PR)
- Release (release-please-managed) so consumers can bump.
- dw3 wiring: on a surfaced `403` with `scope: 'change'` + `changeId` on a branch content doc, call `patches.ejectPendingChange`. The shared `quarantinedChanges` store already exists in dw3's DB (LWW shipped it), so no client migration is needed.

### DIFF
--- a/src/algorithms/ot/shared/ejectPendingChange.ts
+++ b/src/algorithms/ot/shared/ejectPendingChange.ts
@@ -1,4 +1,5 @@
 import { createChange } from '../../../data/change.js';
+import { applyPatch } from '../../../json-patch/applyPatch.js';
 import { invertPatch } from '../../../json-patch/invertPatch.js';
 import type { Change } from '../../../types.js';
 import { applyChanges } from './applyChanges.js';
@@ -46,9 +47,12 @@ export interface PendingEjection {
  * @param pending        The full ordered pending queue.
  * @param changeId       Id of the change to eject.
  * @returns The ejected change plus the rebased queue, or null when `changeId` isn't pending.
- * @throws If the ejected change can't be inverted (a patch/state mismatch). The caller must
- *   treat a throw as "cannot safely eject" and leave the queue untouched — this function
- *   itself mutates nothing, so a throw leaves no partial state.
+ * @throws If the ejected change can't be inverted — including when it no longer applies
+ *   cleanly to its own frame (a patch/state mismatch), where an inverse would be computed
+ *   from values the change never actually saw. The caller must treat a throw as "cannot
+ *   safely eject" and leave the queue untouched — this function itself mutates nothing, so
+ *   a throw leaves no partial state. Only a poison with successors needs the invert; a
+ *   tail-of-queue poison ejects without it.
  */
 export function computePendingEjection(
   committedState: unknown,
@@ -73,6 +77,20 @@ export function computePendingEjection(
     // invertPatch reads each op's prior value (e.g. the base Delta of a `@txt` op) from this
     // state, so it must be the exact frame the poison was minted against.
     const preState = applyChanges(committedState, before);
+    // invertPatch's contract requires ops that apply to the state it reads from. On a
+    // mismatched poison it doesn't reliably throw: a one-level miss (e.g. `replace /arr/5`
+    // over a 3-element array) reads `undefined` silently and fabricates an inverse for an
+    // effect that never happened, which the rebase below would then walk through every
+    // survivor. Probe first so a mismatch always lands on the documented throw-and-latch
+    // path instead of persisting a queue derived from a phantom inverse.
+    try {
+      applyPatch(preState, poison.ops, { strict: true, silent: true });
+    } catch (cause) {
+      throw new Error(
+        `Cannot eject change ${changeId}: it does not apply cleanly to its own frame, so its inverse cannot be trusted`,
+        { cause }
+      );
+    }
     const invertedOps = invertPatch(preState, poison.ops);
     // A synthetic one-change carrier for the inverse. Its rev/baseRev are irrelevant — the
     // survivors are renumbered below — only the diamond walk matters. A fresh id keeps it
--- a/src/algorithms/ot/shared/ejectPendingChange.ts
+++ b/src/algorithms/ot/shared/ejectPendingChange.ts
@@ -1,0 +1,96 @@
+import { createChange } from '../../../data/change.js';
+import { invertPatch } from '../../../json-patch/invertPatch.js';
+import type { Change } from '../../../types.js';
+import { applyChanges } from './applyChanges.js';
+import { rebaseChanges } from './rebaseChanges.js';
+
+export interface PendingEjection {
+  /** The change removed from the queue, unchanged, for quarantine. */
+  poison: Change;
+  /** The queue with the poison gone and every survivor renumbered off `committedRev`. */
+  newPending: Change[];
+}
+
+/**
+ * Remove one change from a sequential OT pending program, rebasing its successors as
+ * though it had never been minted.
+ *
+ * The pending queue is a *sequential program*: each change's ops are expressed in the
+ * frame the changes before it produce (the same model {@link rebaseChanges} documents).
+ * Dropping a change from the middle therefore can't be a plain array splice — every
+ * successor was built on top of the ejected change and must be transformed back into the
+ * frame that skips it.
+ *
+ * The transform is exactly the diamond walk `rebaseChanges` already runs for an incoming
+ * server change: the ejected change genuinely *preceded* its successors, so its inverse is
+ * the "already-happened" side those successors' position ties yield to. We invert the
+ * ejected change against the state it applied to (committed + predecessors) and walk that
+ * inverse through the successors. Predecessors are untouched. Every survivor is then
+ * renumbered contiguously off `committedRev`, preserving the OT pending invariant that all
+ * pending share `baseRev === committedRev` with sequential revs (see
+ * `OTAlgorithm._withConsistentBaseRev`).
+ *
+ * The server accepts `newPending` as a valid poison-free queue and both sides converge on it
+ * deterministically. Exact tie-resolution follows the same one-sided transform as a normal
+ * rebase (see the tie-break note in {@link rebaseChanges}), so at concurrent same-offset ties
+ * this is *a* queue as if the change were never minted, not provably the unique one — the same
+ * caveat every OT rebase carries, not a convergence gap.
+ *
+ * A successor whose edits were scoped to structure the ejected change created (e.g. it added
+ * `/a`, the successor set `/a/b`) transforms away to nothing under the inverse and is dropped —
+ * its content is lost, since it edited something the ejection removes. Only the ejected change
+ * itself is preserved (in quarantine); its dependents are not.
+ *
+ * @param committedState Committed-only doc state — the frame every pending `baseRev` points at.
+ * @param committedRev   The committed revision the queue sits on.
+ * @param pending        The full ordered pending queue.
+ * @param changeId       Id of the change to eject.
+ * @returns The ejected change plus the rebased queue, or null when `changeId` isn't pending.
+ * @throws If the ejected change can't be inverted (a patch/state mismatch). The caller must
+ *   treat a throw as "cannot safely eject" and leave the queue untouched — this function
+ *   itself mutates nothing, so a throw leaves no partial state.
+ */
+export function computePendingEjection(
+  committedState: unknown,
+  committedRev: number,
+  pending: Change[],
+  changeId: string
+): PendingEjection | null {
+  const index = pending.findIndex(change => change.id === changeId);
+  if (index === -1) return null;
+
+  const before = pending.slice(0, index);
+  const poison = pending[index];
+  const after = pending.slice(index + 1);
+
+  let rebasedAfter: Change[];
+  if (after.length === 0) {
+    // Nothing depends on the ejected change, so no rebase (and no invert) is needed — the
+    // common case, since the poison is usually the oldest change blocking the queue head.
+    rebasedAfter = [];
+  } else {
+    // The state the poison applied to = committed state advanced through its predecessors.
+    // invertPatch reads each op's prior value (e.g. the base Delta of a `@txt` op) from this
+    // state, so it must be the exact frame the poison was minted against.
+    const preState = applyChanges(committedState, before);
+    const invertedOps = invertPatch(preState, poison.ops);
+    // A synthetic one-change carrier for the inverse. Its rev/baseRev are irrelevant — the
+    // survivors are renumbered below — only the diamond walk matters. A fresh id keeps it
+    // out of the successors' id set so rebaseChanges walks it rather than dropping it.
+    const inverseCarrier = createChange(poison.baseRev, poison.rev, invertedOps, {});
+    rebasedAfter = rebaseChanges([inverseCarrier], after);
+  }
+
+  // Renumber survivors contiguously off committedRev. Predecessors already sit on
+  // committedRev with these very revs, so this is a no-op for them and only re-seats the
+  // rebased successors into the gap the poison left. rebaseChanges has already dropped any
+  // successor whose ops transformed away to nothing.
+  let rev = committedRev;
+  const newPending = [...before, ...rebasedAfter].map(change => ({
+    ...change,
+    baseRev: committedRev,
+    rev: ++rev,
+  }));
+
+  return { poison, newPending };
+}
--- a/src/client/OTAlgorithm.ts
+++ b/src/client/OTAlgorithm.ts
@@ -1,9 +1,12 @@
 import { applyCommittedChanges } from '../algorithms/ot/client/applyCommittedChanges.js';
+import { applyChanges } from '../algorithms/ot/shared/applyChanges.js';
 import { breakChanges } from '../algorithms/ot/shared/changeBatching.js';
+import { computePendingEjection } from '../algorithms/ot/shared/ejectPendingChange.js';
 import { rebaseChanges } from '../algorithms/ot/shared/rebaseChanges.js';
 import { createChange } from '../data/change.js';
+import { applyPatch } from '../json-patch/applyPatch.js';
 import type { JSONPatchOp } from '../json-patch/types.js';
-import type { Change, PatchesSnapshot } from '../types.js';
+import type { Change, PatchesSnapshot, QuarantinedChange } from '../types.js';
 import type { ClientAlgorithm } from './ClientAlgorithm.js';
 import type { OTClientStore } from './OTClientStore.js';
 import { OTDoc } from './OTDoc.js';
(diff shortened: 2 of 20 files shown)


RyanHavoc:
Pushed 9349aac addressing the findings from a review pass on this PR:

1. **Phantom inverse guard** — `computePendingEjection` now strict-applies the poison against its reconstructed frame before inverting. `invertPatch` silently fabricates an inverse for one-level mismatches (empirically: poison `replace /arr/5` over a 3-element array turned a successor's `replace /arr/6` into `replace /arr/5`), and the rebase would have persisted a queue derived from an effect that never happened. Those shapes now throw and the doc latches. Consequence worth knowing: an un-appliable poison with successors can no longer be auto-ejected at all — the inverse genuinely can't be trusted in that state — so that shape latches for snapshot-reload recovery. Tail poisons (the common case) and cleanly-applying policy ejections are unaffected.
2. **Doc rebuild moved inside the doc lock** — `OTAlgorithm.ejectPendingChange` previously ran `loadDoc` + `import()` after releasing `_withDocLock`; a queued mint could read the doc's stale poison-inclusive queue and persist a mis-framed change into the rebased store queue, and an interleaved receive-rebase could re-inject the poison via the pending merge. The rebuild now imports the in-hand post-ejection snapshot inside the lock (also drops a redundant third `getDoc`), and a rebuild failure can no longer make a durable ejection report as failed.
3. **Torn-write hardening** — the eject snapshot read now merges doc-only in-memory pending id-aware, mirroring `applyServerChanges`, so a change present only in the open doc rides the rebase instead of being silently dropped by the import.
4. **Inversion failure now throws** instead of returning the benign no-match `null`. The app-consent path could not distinguish "already resolved" from "eject impossible, doc still wedged" — the exact call the dw3 wiring will make. `ClientAlgorithm`/`Patches` contracts updated.
5. **verify→eject race closed** — new `opts.onlyIfUnappliable` on `ejectPendingChange` (OT + LWW) re-runs the corroboration probe atomically with the ejection; `PatchesSync` passes it, so a broadcast landing between the probe and the eject can't get now-valid work quarantined (which would also have silently dropped its dependents).
6. **Duplicate listing fixed** — `Patches.listQuarantinedChanges()` dedups by `[docId, changeId]`: in shared-database multi-algorithm setups both algorithm stores read the same `quarantinedChanges` object store and every entry came back twice.
7. **Quarantine list/discard hoisted** to the shared `IndexedDBStore`, with OT/LWW delegating. This also fixes `LWWIndexedDBStore.discardQuarantinedChange` throwing on an external-mode database without the store — the `hasStore` guard existed only in the OT copy.
8. **Stale interface docs** — `ClientAlgorithm` no longer claims "only LWW implements it today" or a committed-only probe frame; `docs/quarantine.md` updated for the throw contract and atomic re-probe.

New tests: phantom-inverse shapes (array index and property-through-primitive) at both the pure-function and algorithm level, `onlyIfUnappliable` behavior for OT and LWW, and multi-algorithm listing dedup. Full suite green (2954 passed), type-check/lint/format clean.

src/algorithms/ot/shared/ejectPendingChange.ts
```diff
@@ -0,0 +1,114 @@
...
+  if (index === -1) return null;
+
+  const before = pending.slice(0, index);
+  const poison = pending[index];
+  const after = pending.slice(index + 1);
+
+  let rebasedAfter: Change[];
+  if (after.length === 0) {
+    // Nothing depends on the ejected change, so no rebase (and no invert) is needed — the
+    // common case, since the poison is usually the oldest change blocking the queue head.
```
matalina, line 73: This rationale describes the opposite case. `after.length === 0` means the poison is at the **tail**; a poison that is "the oldest change blocking the queue head" *has* successors, which is the invert path, not this one. The justification only holds when the queue has exactly one pending change (head and tail at once).

The fast path itself is right — just the reason given for it being common is inverted. Given how load-bearing comments are in this file, worth fixing.

src/algorithms/ot/shared/ejectPendingChange.ts
```diff
@@ -0,0 +1,114 @@
...
+    // out of the successors' id set so rebaseChanges walks it rather than dropping it.
+    const inverseCarrier = createChange(poison.baseRev, poison.rev, invertedOps, {});
+    rebasedAfter = rebaseChanges([inverseCarrier], after);
+  }
+
+  // Renumber survivors contiguously off committedRev. Predecessors already sit on
+  // committedRev with these very revs, so this is a no-op for them and only re-seats the
+  // rebased successors into the gap the poison left. rebaseChanges has already dropped any
+  // successor whose ops transformed away to nothing.
+  let rev = committedRev;
```
matalina, line 106: Minor: this renumber also stamps `baseRev: committedRev` onto `before`. Under the doc lock that's the documented no-op, but if a stale-baseRev straggler is ever present it gets **silently** re-stamped — whereas `_withConsistentBaseRev` does the identical re-stamp with a loud `console.warn` that it "can misplace ops if foreign changes landed in between".

Low severity (the strict probe catches genuinely bad frames), but it silently masks the exact condition the sibling path shouts about. Might be worth reusing that helper, or at least warning on the same trigger.

src/client/OTIndexedDBStore.ts
```diff
@@ -271,6 +277,62 @@ export class OTIndexedDBStore implements OTClientStore {
...
+   * absent on an external-mode database — ejection then fails safe to the caller's error latch.
+   */
+  @blockable
+  async quarantinePendingChange(
+    docId: string,
+    poison: Change,
+    reason: string,
+    rebasedPending: Change[]
+  ): Promise<QuarantinedChange | null> {
+    if (!(await this.db.hasStore('quarantinedChanges'))) return null;
```
matalina, line 298: This `return null` conflates "wedged" with "resolved", which is the one distinction the rest of this PR works hard to preserve.

It propagates through `OTAlgorithm.ejectPendingChange` → `Patches.ejectPendingChange`, whose contract says null means "already committed, already ejected". An app running the consent flow reads that as resolved while the doc is still wedged behind the change — exactly what the throw-vs-null design exists to prevent:

> The throw is deliberately distinct from the benign null ("nothing matched"): an app running the consent flow must be able to tell a resolved eject from a doc still wedged behind the change.

Note LWW **throws** in this situation — `quarantineSendingChange` has no `hasStore` guard, so `db.transaction` raises. So the two algorithms disagree on the same store condition.

The auto-eject path is fine either way (both end at the latch); it's only the public API path that misreports. Suggest throwing here to match LWW and the stated contract.

src/client/OTAlgorithm.ts
```diff
@@ -276,6 +279,144 @@ export class OTAlgorithm implements ClientAlgorithm {
...
+    changeId: string,
+    reason: string,
+    doc?: PatchesDoc<any>,
+    opts?: { onlyIfUnappliable?: boolean }
+  ): Promise<QuarantinedChange | null> {
+    return this._withDocLock(docId, async () => {
+      const snapshot = await this.store.getDoc(docId);
+      if (!snapshot) return null;
+
+      // Merge doc-only in-memory pending exactly like applyServerChanges (id-aware; see the
```
matalina, line 354: Non-blocking, but this merge invites a comparison that took me a while to resolve, so it may be worth a line of comment.

The merge is copied verbatim from `applyServerChanges`, but omits the `_recentCommittedIds` strand filter that immediately follows it there — and `computePendingEjection` then does `applyChanges(committedState, before)`, which looks like precisely the frame math that filter protects ("advances foreign ops through content the committed state already contains").

I built the DAB-607 shape to check: with a strand sitting ahead of the poison, `preState` does double-count (`"AAAAAAhello"` vs the true `"AAAhello"`), and the strict probe doesn't fire. **But it's not a bug** — `computePendingEjection` treats the pending queue as ground truth, so committed+strand genuinely *is* the poison's frame *within that queue*, and removing only the poison stays self-consistent (I verified a delete-shaped poison restores the right characters). The strand filter guards re-sending and foreign-op frame drift; eject does neither.

So the asymmetry is correct — a short "no strand filter here: we're not rebasing against server changes or re-sending, and the queue is the frame" would save the next reader the same detour.

src/client/OTAlgorithm.ts
```diff
@@ -276,6 +279,144 @@ export class OTAlgorithm implements ClientAlgorithm {
...
+        const storedIds = new Set(snapshot.changes.map(change => change.id));
+        const newChanges = inMemoryPending.filter(change => change.rev > latestRev && !storedIds.has(change.id));
+        snapshot.changes.push(...newChanges);
+      }
+
+      // The auto-eject path re-corroborates here, under the same lock the ejection runs in
+      // (its earlier verifyPendingChange probe released the lock, and a broadcast may have
+      // rebased the queue since). Same failure posture as the probe: a frame we can't
+      // reconstruct means we can't corroborate, so don't eject.
+      if (opts?.onlyIfUnappliable) {
```
matalina, line 371: Worth pinning with a test, because this makes OT auto-eject **tail-only** and nothing currently asserts it.

This probe and the one inside `computePendingEjection` are identical — same frame (`snapshot.state` + `snapshot.changes.slice(0, index)` vs `committedState` + `before`), same op, same strictness. So any change that reaches `computePendingEjection` via the auto-eject path has, by construction, already failed that probe — and will fail it again and throw whenever `after.length > 0`. Auto-eject can only ever *succeed* for a tail poison.

That matches what `quarantine.md` documents, and the motivating Pup case is unaffected (it applies cleanly, so it goes down the consent path, which handles mid-queue fine). But no test covers the combination `PatchesSync` actually exercises: `still ejects a change that fails its in-frame probe` uses a tail poison, and the throw tests don't pass `onlyIfUnappliable`. A test for "`onlyIfUnappliable` + un-appliable + successors → throws → doc latches" would lock in the real production behavior.

tests/algorithms/ot/shared/ejectPendingChange.spec.ts
```diff
@@ -0,0 +1,309 @@
...
+    const clientBelief = applyChanges(committed, result.newPending) as any;
+    expect(serverHead).toEqual(clientBelief);
+    expect(textOf(serverHead)).toBe('Yhello\n');
+    expect(serverHead.comments).toEqual({ a: { body: 'note' } });
+  });
+});
+
+describe('computePendingEjection — randomized property tests', () => {
+  const TIMEOUT = 30 * 60_000;
+  const SEEDS = Array.from({ length: 200 }, (_, i) => i + 1);
```
matalina, line 201: Coverage gap: none of these 200 seeds involve a concurrent foreign change. Each seeds a doc then commits the ejected queue with no other client, so the server transform is identity throughout (the assertions note this: "No foreign changes, so the server transform is identity").

That's a strong proof that the ejected queue is well-formed and commits without drops/dupes — but it doesn't exercise convergence *under concurrency*, which is the race `onlyIfUnappliable` was added to guard. A variant that commits a foreign change between the ejection and the flush would close the loop, and would also give the tie-resolution caveat in the docstring some teeth.

matalina (commented):
Reviewed the rebase core closely and tried hard to break it; I couldn't. Every hypothesis I formed got refuted by the implementation, so the notes below are all low-severity — one contract inconsistency, one wrong comment, and a couple of coverage gaps. Nothing here blocks.

**What I ran**

- Full suite on the branch tree: **2919 passed, 0 test failures**. (3 `tests/solid/*.spec.tsx` files failed to *collect* on `file:///@solid-refresh` — an artifact of my junctioned `node_modules`, not this PR.) The new files: 495 pass, and the 400-case property suite runs in ~1.5s, so no CI cost concern.
- **Soft container-init probe.** I expected `add /x {}` to be a trap: `isSoftOp` makes `applyPatch` *skip* it when the path already exists, but `add.invert` still emits a `replace` — which looked like it should kill a successor's writes into that container for a poison whose op was a no-op. It doesn't. `updateSoftWrites` in `add.transform` preserves the successor's own container-init op, so the real Dabble comment shape (every "add a comment" change emits both ops) survives ejection byte-intact. Worth knowing: the practical exposure of the dropped-dependent caveat in the docs is narrower than it reads, since real comment changes self-init. The caveat is accurate for a *bare* successor — I confirmed both soft and non-soft adds drop a lone `add /a/b`.
- **DAB-607 strand probe.** Details inline on `OTAlgorithm.ts`, but short version: I built the strand shape, confirmed `preState` double-counts, and satisfied myself it is **not** a bug.
- Confirmed `_tryEjectPoisonChange` wraps everything in try/catch, so the new throw correctly falls back to the error latch, and `OTInMemoryStore.getDoc` returns copied arrays, so the `snapshot.changes.push` merge doesn't alias store state.

**Things I want to call out as done well**

- The strict-apply probe before `invertPatch` is the most valuable thing in the diff. `invertPatch`'s internal apply is non-strict (`applyPatch(workingState, [op], { silent: true })`), so it silently fabricates inverses on a one-level miss. Catching that, and testing both the array (`replace /arr/5` over 3 elements) and primitive (`replace /s/a` over `'x'`) shapes, is a real find — that phantom `remove` walked through survivors would have been silent corruption.
- Pulling `listQuarantinedChanges`/`discardQuarantinedChange` up into `IndexedDBStore` fixes an actual latent bug (LWW's copy was missing the absent-store guard) and removes the drift vector.
- `computePendingEjection` being pure means a throw genuinely leaves no partial state, and the tests assert that rather than assume it.

**Housekeeping:** `mergeStateStatus` is BLOCKED and the branch is behind `main` by several commits including the `0.19.8` release — worth a rebase. That also explains the test-count gap (PR body says 2946; the branch's tree gives 2923).

src/algorithms/ot/shared/ejectPendingChange.ts
RyanHavoc: Good catch — the code was right and the rationale backwards. Fixed in #113 (a6ad320): the comment now describes it as the tail-of-queue case, including the single-change queue where head and tail coincide.

src/algorithms/ot/shared/ejectPendingChange.ts
RyanHavoc: Agreed. #113 (a6ad320) adds the same loud console.warn _withConsistentBaseRev uses, on the same trigger (any predecessor with a stale baseRev), rather than silently re-stamping.

src/client/OTIndexedDBStore.ts
RyanHavoc: Agreed on all points — this was the one real behavior gap in the batch. #113 (a6ad320) makes quarantinePendingChange throw when the store is absent, matching LWW and the wedged-vs-resolved contract; read paths (list/discard) stay inert. Pinned with a new fake-indexeddb external-mode test.

src/client/OTAlgorithm.ts
RyanHavoc: Appreciated the verification — you're right that it's correct-but-non-obvious. #113 (a6ad320) adds the comment: no strand filter because ejection neither rebases against server changes nor re-sends; the queue is the frame, and removing only the poison stays self-consistent with a strand ahead of it.

src/client/OTAlgorithm.ts
RyanHavoc: Added in #113 (a6ad320): a test for onlyIfUnappliable + un-appliable + successors asserting it throws and the queue latches untouched — pinning auto-eject as tail-only, which is the combination PatchesSync exercises in production.

tests/algorithms/ot/shared/ejectPendingChange.spec.ts
RyanHavoc: Closed in #113 (a6ad320): a deterministic case plus 200 seeded runs now commit a foreign change between the ejection and the flush, and convergence is asserted client-side too via applyCommittedChanges (pending clears, client state equals the server head).

PR #114, opened 2026-07-16
feat(net): detect and recover misdirected SSE subscribes

RyanHavoc:
**TL;DR:** Foundation for the prod live-sync fix (the SSE subscribe instance lottery). Server side: `SSEServer.hasClient()` + `SSEServer.addSubscriptions()` so a multi-instance deployment can route/forward subscribes to the stream-owning instance. Client side: `PatchesREST.subscribe` retries a subscribe that silently registered nothing, honors explicit `denied` accounting via a set-based completeness gate, rebuilds the stream on an `UNKNOWN_CLIENT` verdict, and surfaces retry exhaustion by rejecting with `SubscribeIncompleteError` (caught by `PatchesSync`, which emits telemetry exactly once).

## Why

pup runs multi-instance with no session affinity. `SSEServer` state is instance-local, so a `POST /subscriptions/:clientId` that lands on a different instance than the one holding the client's `/events` stream hits `subscribe()`'s `if (!client) return []` — nothing is registered, nothing errors, and live updates for that stream silently never arrive. With ~14 prod instances that's a ~1/14 lottery re-rolled on every stream reconnect: cross-client live sync has been refresh-only for most sessions since the GA cutover made pup multi-instance, invisible to telemetry because commits and reads succeed.

## Server (SSEServer)

- `hasClient(clientId)` — whether THIS instance holds state for the client (live stream, or disconnected within buffer TTL). Lets a routing layer tell "unknown client" apart from "all docIds denied", which `subscribe()`'s `[]` conflates (now documented on the method).
- `addSubscriptions(clientId, docIds)` — registers **pre-authorized** docIds without auth checks, for subscribes forwarded from the instance that ran authorization. Returns `null` (not `[]`) for an unknown client so a failed forward is detectable.

## Client (PatchesREST.subscribe)

- The completeness gate is **set-based**, per the `denied[]` contract: every requested id must be accounted for as a member of either the registered `docIds` or the new optional `denied[]` response field (authoritative authorization denials). Any unaccounted ids are retried up to 4 attempts with backoff — each POST re-rolls the load balancer — and each retry re-POSTs **only the still-missing ids**. `denied: []` accounts for nothing and does not disable retry. Exhaustion **rejects** with `SubscribeIncompleteError` — `PatchesSync` catches it at both call sites and emits `onError` exactly once (→ app telemetry) while sync continues; direct `PatchesREST` consumers now see a rejection instead of a silent `[]`.
- A `StatusError` carrying `data.code: 'UNKNOWN_CLIENT'` (new server verdict: no instance holds a stream for this clientId) retries the same way, then tears down the stale stream and schedules a reconnect — guarded on the EventSource captured at subscribe entry, so a stream that already reconnected mid-retry is never torn down — the fresh `connected` state drives the consumer's full re-subscribe. Other status errors propagate immediately, unchanged.
- Partial registrations without explicit `denied` accounting ARE retried (missing ids only). Only ids the server lists in `denied` are treated as authorization denials and dropped. Servers must omit transiently-errored ids from **both** arrays — an id in neither array means "unknown, safe to retry" — or the accounting is not sound.
- No behavior change for empty requests, and old servers/new clients + new servers/old clients interoperate (all changes are additive).

Contract documented in `docs/sse-rest.md` (new "Multi-instance deployments" section).

## Release / rollout

- Version bumped to **0.21.0** in this PR (feat); publish to npm after merge — dabblewriter/pup and dabblewriter/dabble-writer-3.0 have chained PRs bumping to `^0.21.0` (pup's subscribe-forwarding needs the new SSEServer APIs).
- Safe to release standalone, but **not telemetry-only**: in the client-ships-first window, a misdirected doc subscribe stalls that doc's live sync for the retry duration (~3.5s) before recovering — `PatchesSync` awaits `subscribe` before `syncDoc`. Today that same misdirection silently loses live updates for the whole session, so the trade is a bounded stall for eventual recovery.

## Testing

- `npm run test` — 3185 passed (19 new: hasClient/addSubscriptions, set-based retry/denied accounting, UNKNOWN_CLIENT/teardown-guard paths)
- `npm run type:check`, `npm run lint`, `npm run format:check` — clean

### DIFF
--- a/docs/sse-rest.md
+++ b/docs/sse-rest.md
@@ -112,13 +112,21 @@ All document operations use standard HTTP methods. Document IDs can contain slas
 | DELETE | `/subscriptions/:clientId` | Unsubscribe from docs `{ docIds: [...] }`                                     |
 | POST   | `/signal/:clientId`        | WebRTC signaling frame (raw JSON-RPC body). See [awareness.md](awareness.md). |
 
-The subscribe response is `{ docIds: [...] }` — the ids the server actually registered. Servers may
-additionally return `denied: [...]` (ids excluded by authorization) so the client can tell an
-authoritative empty result apart from a subscribe that silently registered nothing, and may answer
-with a status error carrying `data.code: 'UNKNOWN_CLIENT'` when no instance holds an event stream
-for the clientId. `PatchesREST` uses both: it retries a nothing-registered response (unless `denied`
-accounts for every id) and rebuilds the stream on `UNKNOWN_CLIENT`. See
-[Multi-instance deployments](#multi-instance-deployments).
+The subscribe response is `{ docIds: [...], denied?: [...] }`:
+
+- `docIds` — the ids actually registered on the instance holding the client's event stream.
+- `denied` — ids **authoritatively refused by access control** (a definitive authorization "no").
+- An id whose access check errored transiently (a thrown/rejected check, an unreachable registry)
+  belongs in **neither** array — absence means "unknown, safe to retry".
+
+`PatchesREST` gates completeness on set membership, never on counts: the response is complete when
+every requested id appears in `docIds` or `denied`. Unaccounted ids are retried — only those ids,
+with backoff, each POST re-rolling the load balancer. An older server that never sends `denied`
+simply leaves non-registered ids unaccounted (they retry naturally), and `denied: []` accounts for
+nothing — it does not disable retries. Servers may also answer with a status error carrying
+`data.code: 'UNKNOWN_CLIENT'` when no instance holds an event stream for the clientId;
+`PatchesREST` responds by rebuilding its stream. Adopting `denied` requires the unknown-client
+check — see [Multi-instance deployments](#multi-instance-deployments) for the contract.
 
 ### Documents
 
@@ -240,13 +248,58 @@ Route subscribes to the stream-owning instance yourself (e.g. record which insta
 Three APIs support this:
 
 - `hasClient(clientId)` — whether THIS instance holds state for the client (live stream or a
-  disconnected client within its buffer TTL).
+  disconnected client within its buffer TTL). Treat `true` as "may hold state", **not** "owns the
+  live stream": a disconnected client stays claimable for up to `bufferTTLMs` (default 5 minutes),
+  and without session affinity a reconnect routinely lands on a different instance — so during
+  that window two instances can both truthfully answer `true` for the same client. A router that
+  picks the stale claimant registers subscriptions on an instance the client will never return
+  to, and events buffer there silently. Prefer the instance holding the live stream (shared state
+  recording who served `/events/:clientId` last, or a live-only check à la `getConnectionIds()`)
+  whenever both claim the client.
 - `addSubscriptions(clientId, docIds)` — register **already-authorized** docIds without auth
   checks, for a forwarded subscribe arriving from the instance that ran authorization. Returns
-  `null` (instead of `[]`) when the client is unknown, so a failed forward is detectable.
+  `null` (instead of `[]`) when the client is unknown, so a failed forward is detectable. This
+  method is trusted: it must never be reachable with client-supplied docIds that haven't passed
+  authorization — wiring it to such a route hands any client read access to arbitrary documents
+  via `changesCommitted`.
 - When no instance owns the stream, answer the POST with a status error carrying
   `data.code: 'UNKNOWN_CLIENT'` — `PatchesREST` responds by rebuilding its stream and
   re-subscribing, instead of waiting on a stream that no longer exists.
+
+#### The `denied` contract
+
+`denied` must mean "authorization refused" — never "this instance didn't know the client" and
+never "the access check errored". The `denied` field and the `hasClient`/`UNKNOWN_CLIENT` check
+are a package deal, not independent options: a server that adopts `denied` alone will answer a
+misdirected subscribe with `{ docIds: [], denied: [everything] }`, which the client treats as
+authoritative — no retry, no telemetry, and live updates silently never arrive. That re-creates
+the exact failure this contract exists to fix, with the client-side mitigation disabled. Adopting
+`denied` without the unknown-client gate is strictly worse than not adopting it.
+
+The required ordering:
+
+```typescript
+app.post('/subscriptions/:clientId', async c => {
+  const clientId = c.req.param('clientId');
+  if (!sse.hasClient(clientId)) {
+    // Forward to the stream-owning instance — or, when no instance owns it:
+    return c.json({ error: 'Unknown client', data: { code: 'UNKNOWN_CLIENT' } }, 409);
+  }
+  const { docIds } = await c.req.json();
+  const ctx = { clientId, ...c.get('auth') };
+  const subscribed = await sse.subscribe(clientId, docIds, ctx);
+  return c.json({ docIds: subscribed, denied: docIds.filter((id: string) => !subscribed.includes(id)) });
+});
+```
+
+The subtraction on the last line is only safe because the `hasClient` gate above it ran — and only
+when every non-registered id really was an authorization "no". If an access check can fail
+transiently (a database blip, a registry timeout), the affected ids must be left out of BOTH
+arrays so the client retries them; sweeping them into `denied` converts a transient fault into a
+permanent, silent subscription loss. Note that `SSEServer.subscribe` folds errored checks into
+"not registered" — subtraction over its result is only contract-safe when `canAccess` cannot fail
+transiently. Otherwise derive `denied` from your own auth results, or omit the field entirely:
+without `denied`, the client simply retries any shortfall, which is always safe.
 
 ## Reconnection and Recovery
 
--- a/docs/sse-rest.md
+++ b/docs/sse-rest.md
@@ -112,6 +112,14 @@ All document operations use standard HTTP methods. Document IDs can contain slas
 | DELETE | `/subscriptions/:clientId` | Unsubscribe from docs `{ docIds: [...] }`                                     |
 | POST   | `/signal/:clientId`        | WebRTC signaling frame (raw JSON-RPC body). See [awareness.md](awareness.md). |
 
+The subscribe response is `{ docIds: [...] }` — the ids the server actually registered. Servers may
+additionally return `denied: [...]` (ids excluded by authorization) so the client can tell an
+authoritative empty result apart from a subscribe that silently registered nothing, and may answer
+with a status error carrying `data.code: 'UNKNOWN_CLIENT'` when no instance holds an event stream
+for the clientId. `PatchesREST` uses both: it retries a nothing-registered response (unless `denied`
+accounts for every id) and rebuilds the stream on `UNKNOWN_CLIENT`. See
+[Multi-instance deployments](#multi-instance-deployments).
+
 ### Documents
 
 | Method | Path                             | Description                            |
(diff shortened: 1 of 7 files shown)


src/net/rest/PatchesREST.ts
```diff
@@ -298,12 +309,82 @@ export class PatchesREST implements PatchesConnection {
...
+        }
+        throw err;
+      }
+      unknownClient = null;
+      const docIds = Array.isArray(result?.docIds) ? (result.docIds as string[]) : [];
+      if (docIds.length > 0) return docIds;
+      // Nothing registered. A server that reports denial accounting told us
+      // every id was denied — a real (authoritative) empty result, not a
+      // misdirected subscribe. Older servers leave the two indistinguishable.
+      if (Array.isArray(result?.denied)) return docIds;
```
matalina: This gate is weaker than the contract the JSDoc and PR description state ("accounted for **every** id as denied") — it only checks the field's *type*.

Probed it: `{ docIds: [], denied: [] }` for a 1-id request suppresses the retry **and** the `SubscribeIncompleteError` telemetry (1 fetch, 0 `onError`). An empty `denied` accounts for nothing, so by the documented contract that should retry.

The bigger worry is the partial-adoption path. A server that adds `denied` but *not* the `hasClient`/`UNKNOWN_CLIENT` check will, on a misdirected subscribe, return `{ docIds: [], denied: [everything] }` (see my note on `SSEServer.ts:259` — subtraction is the only derivation available to server authors). The client reads that as authoritative denial: no retry, no telemetry, original bug intact, mitigation disabled. Adopting `denied` alone is strictly worse than not adopting it.

Tightening to `docIds.length + denied.length === requested.length` would match the stated contract and fix the `denied: []` variant. The subtraction variant can only be prevented by making the `denied` + `UNKNOWN_CLIENT` coupling explicit in the docs.

src/net/rest/PatchesREST.ts
```diff
@@ -298,12 +309,82 @@ export class PatchesREST implements PatchesConnection {
...
+    }
+
+    if (unknownClient) {
+      // Authoritative: our stream is gone server-side (stale/half-open here).
+      // Rebuild it — the consumer re-subscribes everything on 'connected'.
+      if (this.shouldBeConnected) {
+        this._teardownStream();
```
matalina, line 400: This tears down whatever stream is current when the loop exhausts, not the one the subscribe was issued against.

With two overlapping subscribes — routine, since `PatchesSync.ts:483` (resync) and `:1222` (newly-tracked docs) can overlap — A exhausts at ~3.5s and triggers a reconnect, then B, started 2s later, exhausts at ~5.5s and tears down the freshly rebuilt stream. Confirmed with a probe:

```
F3 rebuilt stream open -> connected | after B exhausts -> error
F3 rebuilt stream closed by B? true
```

It self-heals via another reconnect, so churn rather than a hang: one extra teardown → error → reconnect → full re-subscribe cycle.

The idiom for this already exists a few lines up at `:184` — capture the `EventSource` at subscribe start and only tear down if `this.eventSource === capturedEs`.

src/net/rest/PatchesREST.ts
```diff
@@ -298,12 +309,82 @@ export class PatchesREST implements PatchesConnection {
...
+        this._setState('error');
+        this._scheduleReconnect();
+      }
+      this.onError.emit(unknownClient);
```
matalina: This path both emits and throws the same object (probe: `thrown === emitted[0]`), and `PatchesSync` consumes both channels — it forwards `connection.onError` into its own `onError` at `PatchesSync.ts:169`, *and* both subscribe call sites catch and re-emit (`:487`, `:1228`).

So every exhausted `UNKNOWN_CLIENT` lands in app telemetry twice, which distorts exactly the signal this PR exists to create.

Dropping the emit and keeping the throw restores symmetry with how a 403 already behaves: propagates, emitted once by `PatchesSync`. The `SubscribeIncompleteError` path below is already right — it resolves `[]`, so its single emit is the only channel.

src/net/rest/SSEServer.ts
```diff
@@ -206,13 +206,55 @@ export class SSEServer {
...
    * @param docIds - Document IDs to subscribe to.
    * @param ctx - Auth context for authorization checks.
-   * @returns The list of document IDs successfully subscribed.
+   * @returns The list of document IDs successfully subscribed. NOTE: an
+   *   unknown clientId (no stream state on this instance) also returns [] —
+   *   indistinguishable from every docId being denied. Callers that need to
+   *   tell the two apart (e.g. to route a misdirected subscribe to the right
+   *   instance) should check {@link hasClient} first.
    */
   async subscribe(clientId: string, docIds: string[], ctx?: AuthContext): Promise<string[]> {
```
matalina, line 275: Related to the `denied` gate on the client (`PatchesREST.ts:368`): this returns only `allowed` — the `Promise.allSettled` rejections are dropped — so there's **no API to obtain the true auth-denied list**.

A server author implementing the new `denied` wire field can therefore only derive it by subtraction, which the class JSDoc example steers them toward:

```ts
const subscribed = await sse.subscribe(clientId, docIds, ctx);
return c.json({ docIds: subscribed, denied: docIds.filter(id => !subscribed.includes(id)) });
```

For a misdirected subscribe that's `{ docIds: [], denied: [everything] }` — the client treats it as authoritative and stops retrying. The API pushes authors toward the one derivation that breaks the fix.

Worth considering returning `{ docIds, denied }` (or a variant) so the safe derivation is the available one. Failing that, the example above should show the `hasClient` gate first, since subtraction is only correct once the unknown-client case is ruled out.

matalina (commented):
Really nice diagnosis — the instance-lottery writeup made the failure mode obvious, and the change is well-scoped, additive, and genuinely backward-compatible. Comments explain *why* rather than *what*, which matches the surrounding file. I ran the two touched specs (109 passing) and verified the notes below with throwaway tests in a worktree rather than reasoning from reading alone.

Seven inline notes. The ones I'd weigh most:

1. **The `denied` gate can silently defeat the fix** (`PatchesREST.ts:368`, `SSEServer.ts:259`, `docs/sse-rest.md:116`) — the gate is weaker than the contract you documented, and `SSEServer.subscribe()` only offers server authors the *unsafe* way to derive `denied`. A well-intentioned partial adoption re-creates the original bug with the client mitigation disabled.
2. **Exhausted `UNKNOWN_CLIENT` double-reports to telemetry** (`PatchesREST.ts:379`) — distorts the signal this PR exists to create.
3. **A stale subscribe tears down a healthy stream** (`PatchesREST.ts:375`) — self-heals, but costs an extra teardown/reconnect/re-subscribe cycle.

Plus two notes for the chained pup PR (`hasClient` stale-claim window, `addSubscriptions` blast radius).

One rollout note that isn't inline-able: "the only behavior change is capped subscribe retries + telemetry" understates it slightly. `PatchesSync` awaits `subscribe` before `syncDoc` (`PatchesSync.ts:483`), so in the window where this client ships but pup's routing doesn't, a misdirected subscribe stalls doc sync by up to 3.5s. Still a clear net win — 4 attempts take recovery from ~1/14 to ~25% — but worth naming in the rollout notes.

Not gating this either way; flagging and leaving the calls to you.

src/net/rest/SSEServer.ts
```diff
@@ -206,13 +206,55 @@ export class SSEServer {
...
+   * SSE stream, or a disconnected client still within its buffer TTL.
+   *
+   * In a multi-instance deployment each SSEServer only knows its own clients:
+   * a subscribe request load-balanced to a different instance than the one
+   * holding the client's stream would silently register nothing (see
+   * {@link subscribe}). Callers routing between instances should use this to
+   * decide whether to handle a subscribe locally or forward it to the
+   * stream-owning instance ({@link addSubscriptions}).
+   */
+  hasClient(clientId: string): boolean {
```
matalina, line 236: Worth documenting the stale-claim window here, since this doc block actively recommends `hasClient` for the routing decision.

This returns true for a disconnected client still within `bufferTTLMs` (default **5 minutes**). Without session affinity a reconnect routinely lands on a different instance — so for up to 5 minutes *two* instances answer `hasClient` truthfully for the same client.

If pup's router picks the stale one, `addSubscriptions` succeeds (it checks `clients.get()`, not `writer`), returns a non-null list, and the client sees a healthy `docIds.length > 0` and never retries. Events then buffer on an instance the client will never return to — the same silent failure this PR is fixing, one layer up.

Lands in pup's PR rather than here, but a note on the window (or a live-only distinction — `getConnectionIds()` filters on `writer`, though it's O(n)) would keep the router honest.

src/net/rest/SSEServer.ts
```diff
@@ -206,13 +206,55 @@ export class SSEServer {
...
+   * the HTTP request to the instance that owns the client's event stream.
+   * Never call this with client-supplied docIds that haven't passed
+   * authorization.
+   *
+   * @returns The registered docIds, or null when the client is unknown on this
+   *   instance — distinguishable from success so the caller can surface the
+   *   failure (and the client can reconnect/retry) instead of the subscription
+   *   being silently dropped.
+   */
+  addSubscriptions(clientId: string, docIds: string[]): string[] | null {
```
matalina, line 254: No change requested — the JSDoc is appropriately loud and skipping auth for pre-authorized ids is the right design for a forwarded subscribe.

Just flagging the blast radius for whoever reviews pup's side: this is a public method on an exported class with no structural guard, and wiring it to a route with client-supplied docIds would hand any client read access to arbitrary documents via `changesCommitted`. That's the thing to scrutinize hardest in the chained PR.

docs/sse-rest.md
```diff
@@ -112,6 +112,14 @@ All document operations use standard HTTP methods. Document IDs can contain slas
 | DELETE | `/subscriptions/:clientId` | Unsubscribe from docs `{ docIds: [...] }`                                     |
 | POST   | `/signal/:clientId`        | WebRTC signaling frame (raw JSON-RPC body). See [awareness.md](awareness.md). |
 
+The subscribe response is `{ docIds: [...] }` — the ids the server actually registered. Servers may
+additionally return `denied: [...]` (ids excluded by authorization) so the client can tell an
```
matalina: These read as two independent options ("Servers may additionally return `denied`" … "and may answer with … `UNKNOWN_CLIENT`"), but they're a package deal: `denied` is only meaningful *after* a `hasClient()` check has ruled out the unknown-client case.

A server that adopts `denied` alone will return `{ docIds: [], denied: [everything] }` for a misdirected subscribe, which the client treats as authoritative — no retry, no telemetry, and the bug this PR fixes is back with the mitigation disabled (see `PatchesREST.ts:368`).

Suggest stating the coupling as mandatory and showing the required ordering:

```ts
if (!sse.hasClient(clientId)) { /* forward to the owning instance, or 409 UNKNOWN_CLIENT */ }
const subscribed = await sse.subscribe(clientId, docIds, ctx);
return c.json({ docIds: subscribed, denied: docIds.filter(id => !subscribed.includes(id)) });
```

The subtraction on the last line is only safe because the gate above it ran. Also worth saying explicitly that `denied` must mean "authorization denied" and never "this instance didn't know the client".

matalina (approved):

RyanHavoc:
Pushed 03f94c1 addressing the review:

- **Completeness gate (set-based accounting):** `subscribe` now dedupes the requested ids and accumulates each response's `docIds` and `denied` into sets; the request is complete only when every requested id is a member of the union. Retries re-POST only the still-missing ids, `denied: []` accounts for nothing (retry stays enabled), all-denied full accounting completes without retry or error, and an absent `denied` (old server) leaves unregistered ids retrying naturally. This flips the original "partial registrations are NOT retried" semantics — partials without explicit `denied` accounting now retry, per the agreed contract.
- **UNKNOWN_CLIENT teardown guard:** `subscribe()` captures `this.eventSource` at entry (same idiom as `connect()`), and the exhaustion branch tears down only if that captured stream is still current. A new test reproduces the probe from review: a stream that dies and reconnects mid-retry is never torn down by the stale subscribe.
- **Telemetry single-channel:** exhaustion now rejects with `SubscribeIncompleteError` instead of resolving `[]` + emitting `onError` locally, and the `UNKNOWN_CLIENT` rejection no longer double-emits. `PatchesREST.onError` stays scoped to transport-level errors per its JSDoc; `PatchesSync` catches the rejection at both call sites and emits exactly once (matching 403 symmetry). Note for any direct `PatchesREST` consumer: exhaustion is now a rejection.
- **Docs:** `docs/sse-rest.md` now specifies the response contract (`{docIds, denied?}`, `denied` = authoritative authorization no, transiently-errored ids in neither array = "unknown, safe to retry", set-membership completeness) and a new "denied contract" subsection covering the mandatory `denied` + `hasClient`/`UNKNOWN_CLIENT` coupling, the 5-minute `hasClient` dual-claim window, and the `addSubscriptions` trust boundary. `SSEServer.ts` JSDoc mirrors the stale-claim window on `hasClient` and shows the `hasClient` gate first in the class example. Callout included that `SSEServer.subscribe` folds errored checks into "not registered", so deriving `denied` by subtraction is only contract-safe when `canAccess` cannot fail transiently — routes should derive `denied` from their own auth results or omit it.
- **Version:** already at 0.21.0 (25befe8); no change needed.

Tests: 3185 passed / 4 pre-existing skips (+5 net new covering the above); `type:check`, `lint`, `format:check` clean. PR body updated to reflect the set-based gate, the retry-on-partial semantics, the rejection on exhaustion, and the rollout note (client-first window trades a bounded ~3.5s per-doc stall for recovery, not telemetry-only).

PR #115, opened 2026-07-18
fix(branches): stop editor-copy merges from doubling the manuscript (DAB-760)

matalina:
## Summary

Merging an **editor copy** could re-insert every scene's body onto `main`, doubling the manuscript — even when the reviewer had **rejected every tracked change** (a net-zero merge). Reported in [DAB-760](https://linear.app/dabblewriter/issue/DAB-760) (Leah Smith: "Chapter 1 is 16 pages instead of 8").

## Root cause

An editor copy is a client-seeded branch: the whole project is seeded as a root-replace split into changes, and `contentStartRev` marks where the seed ends so `mergeBranch` can exclude it (`startAfter = contentStartRev - 1`).

The floor was computed from a **compressed** storage split (`breakChanges` + `compressedSizeUint8`, `MAX_STORAGE_BYTES` = 900 KB) while `PatchesSync` commits the seed via `breakChangesIntoBatches` using the **uncompressed** 1 MB payload limit. Prose compresses ~3×, so a large manuscript's seed fits **one** compressed storage piece, yet the wire re-splits it into **more** (renumbered) changes than `contentStartRev` counted — and the floor was already sent to the server before the flush. The extra per-doc `@txt` seed pieces then land **above** the floor, so the merge replays them as retain-less whole-body inserts. It survives "reject all" because the doubled content is the **seed**, not the tracked edits.

Measured against a real 113 K-word manuscript:

| Scale | Words | Seed uncompressed | Compressed | Result |
|---|---|---|---|---|
| 1× | 113 K | 0.61 MB | 0.23 MB (2.7×) | no desync |
| 2× | 226 K | 1.22 MB | 0.43 MB | **doubles** |
| 3× | 339 K | 1.83 MB | 0.62 MB | **doubles** |

The bug fires once the seed's uncompressed JSON exceeds ~1 MB (≈ 185 K+ words) — the band where uncompressed is over the 1 MB wire limit but compressed is still under the 900 KB storage limit.

## The fix (two layers)

1. **Root cause** — seed the branch with `breakChangesIntoBatches` (both limits) instead of `breakChanges` (storage only), so the persisted seed is flush-stable and `contentStartRev` matches the committed span. `maxPayloadBytes` is now a `PatchesDocOptions` field that `PatchesSync` also resolves (consistent with how it already falls back `maxStorageBytes` / `sizeCalculator`), so seeding and sync agree.
2. **Backstop** — `OTBranchManager.mergeBranch` refuses (with telemetry, `MergeContentDuplicationError`) any merge whose `@txt` op re-inserts a field's existing head — the doubling signature — instead of corrupting the document. This protects branches already created with a bad floor. The check is cheap in the common case (it only reconstructs state when a change carries a substantial leading text insert) and does not block a legitimate leading insert of new text.

## Testing

- **Recreate** builds the desync from the **real** batching functions + `compressedSizeUint8` (no hand-set floor) and confirms the guard refuses the merge.
- **Guard**: an undercounting floor is refused; `main` left intact.
- **Seeding**: the wire-split seed makes `contentStartRev` equal the committed span.
- **No false positive**: a legitimate large leading insert of new text still merges.
- Full suite green (3155 passing; the 3 pre-existing `@solid-refresh` Solid.js suite failures are unrelated). `tsc` / eslint / prettier clean.

## Rollout

The merge runs server-side, so the fix takes effect when **pup redeploys** the new `@dabble/patches`. The seeding fix additionally ships to clients when dw3 rebundles (prevents *new* branches from getting a bad floor).

## Follow-up (not in this PR)

Editor copies **already created** with an undercounting `contentStartRev` will now *fail* to merge under the guard (no corruption, but blocked) until their floor is corrected. Tracked separately: recompute the true seed boundary `M` (where the branch's replayed content first equals `source@branchedAtRev`) and set `contentStartRev = M + 1`, delivered as a one-time migration or a self-heal in `mergeBranch`.

### DIFF
--- a/src/client/PatchesBranchClient.ts
+++ b/src/client/PatchesBranchClient.ts
@@ -169,22 +169,19 @@ export class PatchesBranchClient {
       committedAt: 0,
     });
 
-    // Split the seed exactly as PatchesSync will commit it on the wire — by BOTH the per-change
-    // storage limit (possibly a compressed measure) AND the uncompressed payload limit — so the
-    // pending seed we persist here matches the changes that actually get committed. Splitting by
-    // storage only let the wire's payload re-split add revisions the branch never counted, so
-    // `contentStartRev` undercounted the seed and the merge replayed the tail as a retain-less
-    // whole-body insert, doubling the document (DAB-760). `maxPayloadBytes` is resolved the same
-    // way PatchesSync resolves it (docOptions, else a 1MB default) so the two splits agree.
+    // Pre-split the seed the way PatchesSync will commit it on the wire — by BOTH the
+    // per-change storage limit (possibly a compressed measure) AND the uncompressed payload
+    // limit — so the pending changes persisted below are flush-stable: a flush that re-split
+    // the seed would renumber it under a frozen `contentStartRev`, and a later merge would
+    // replay the seed's tail into the source document, duplicating its content.
+    // `maxPayloadBytes` is resolved the same way PatchesSync resolves it (docOptions, else a
+    // 1MB default) so the two splits agree.
     const options = this.patches.docOptions;
     const initChanges = breakChangesIntoBatches([rootReplace], {
       maxPayloadBytes: options.maxPayloadBytes,
       maxStorageBytes: options.maxStorageBytes,
       sizeCalculator: options.sizeCalculator,
     }).flat();
-
-    // contentStartRev is the first revision after all init changes
-    const contentStartRev = initChanges[initChanges.length - 1].rev + 1;
 
     // Validate algorithm exists before persisting anything
     const algorithmName = this.options?.algorithm ?? this.patches.defaultAlgorithm;
@@ -193,17 +190,30 @@ export class PatchesBranchClient {
       throw new Error(`Algorithm '${algorithmName}' not found`);
     }
 
-    // Create branch via the API (store saves with pendingOp: 'create')
-    await this.api.createBranch(this.id, rev, { ...metadata, contentStartRev });
-
     try {
       // Track the branch document through the standard pipeline
       await this.patches.trackDocs([branchDocId], algorithmName);
 
-      // Save initial changes as pending through the algorithm's store
+      // Persist the seed as pending changes, then derive `contentStartRev` from the
+      // revisions the algorithm ACTUALLY assigned. The algorithm applies its own storage
+      // split with its own options, so predicting the final revision span from `docOptions`
+      // desyncs the floor whenever the two configs disagree — and an undercounted floor
+      // makes a merge replay the seed's tail, duplicating the document's content. The
+      // persisted revs are the truth.
+      let lastSeedRev = 0;
       for (const change of initChanges) {
-        await algorithm.handleDocChange(branchDocId, change.ops, undefined, {});
+        const persisted = await algorithm.handleDocChange(branchDocId, change.ops, undefined, {});
+        lastSeedRev = persisted[persisted.length - 1]?.rev ?? lastSeedRev;
       }
+      if (!lastSeedRev) {
+        throw new Error('Branch seeding persisted no changes');
+      }
+
+      // contentStartRev is the first revision after the persisted seed
+      const contentStartRev = lastSeedRev + 1;
+
+      // Create branch via the API (store saves with pendingOp: 'create')
+      await this.api.createBranch(this.id, rev, { ...metadata, contentStartRev });
     } catch (err) {
       // Rollback: remove saved branch meta and untrack doc so we don't leave inconsistent state
       if (this.isOffline) {
--- a/src/client/PatchesBranchClient.ts
+++ b/src/client/PatchesBranchClient.ts
@@ -1,5 +1,5 @@
 import { store, type Store } from 'easy-signal';
-import { breakChanges } from '../algorithms/ot/shared/changeBatching.js';
+import { breakChangesIntoBatches } from '../algorithms/ot/shared/changeBatching.js';
 import { createChange } from '../data/change.js';
 import type { BranchAPI } from '../net/protocol/types.js';
 import type { Branch, Change, CreateBranchMetadata, EditableBranchMetadata, ListBranchesOptions } from '../types.js';
@@ -169,12 +169,19 @@ export class PatchesBranchClient {
       committedAt: 0,
     });
 
-    // Break the change into multiple if it exceeds the storage size limit
-    let initChanges = [rootReplace];
+    // Split the seed exactly as PatchesSync will commit it on the wire — by BOTH the per-change
+    // storage limit (possibly a compressed measure) AND the uncompressed payload limit — so the
+    // pending seed we persist here matches the changes that actually get committed. Splitting by
+    // storage only let the wire's payload re-split add revisions the branch never counted, so
+    // `contentStartRev` undercounted the seed and the merge replayed the tail as a retain-less
+    // whole-body insert, doubling the document (DAB-760). `maxPayloadBytes` is resolved the same
+    // way PatchesSync resolves it (docOptions, else a 1MB default) so the two splits agree.
     const options = this.patches.docOptions;
-    if (options.maxStorageBytes) {
-      initChanges = breakChanges(initChanges, options.maxStorageBytes, options.sizeCalculator);
-    }
+    const initChanges = breakChangesIntoBatches([rootReplace], {
+      maxPayloadBytes: options.maxPayloadBytes,
+      maxStorageBytes: options.maxStorageBytes,
+      sizeCalculator: options.sizeCalculator,
+    }).flat();
 
     // contentStartRev is the first revision after all init changes
     const contentStartRev = initChanges[initChanges.length - 1].rev + 1;
--- a/src/net/PatchesSync.ts
+++ b/src/net/PatchesSync.ts
@@ -144,7 +144,7 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
     }
 
     // Use options if provided, otherwise fall back to patches.docOptions
-    this.maxPayloadBytes = options?.maxPayloadBytes;
+    this.maxPayloadBytes = options?.maxPayloadBytes ?? patches.docOptions?.maxPayloadBytes;
     this.maxStorageBytes = options?.maxStorageBytes ?? patches.docOptions?.maxStorageBytes;
     this.sizeCalculator = options?.sizeCalculator ?? patches.docOptions?.sizeCalculator;
 
--- a/src/server/OTBranchManager.ts
+++ b/src/server/OTBranchManager.ts
@@ -2,6 +2,7 @@ import { findLatestMainVersion } from '../algorithms/ot/server/getSnapshotAtRevi
 import { getStateAtRevision } from '../algorithms/ot/server/getStateAtRevision.js';
 import { breakChanges } from '../algorithms/ot/shared/changeBatching.js';
 import { createChange } from '../data/change.js';
+import { toOps } from '../json-patch/ops/text.js';
 import { createVersionMetadata } from '../data/version.js';
 import type { Branch, Change, CreateBranchMetadata, EditableBranchMetadata, ListBranchesOptions } from '../types.js';
 import type { BranchManager } from './BranchManager.js';
@@ -24,6 +25,74 @@ import type { BranchingStoreBackend, OTStoreBackend } from './types.js';
  * Requires both OT operations and branch metadata operations.
  */
 type OTBranchStore = OTStoreBackend & BranchingStoreBackend;
+
+/**
+ * Minimum length of a re-inserted text run we treat as content *duplication* rather than a
+ * legitimate leading edit. Bodies/descriptions that double are far longer than this; a short
+ * genuine prepend never reaches it.
+ */
+const MERGE_DUP_MIN_CHARS = 16;
+
+/** The run of text a `@txt` delta prepends at position 0 (empty when it begins with retain/delete). */
+function leadingInsertText(value: unknown): string {
+  const ops = toOps(value);
+  if (!ops) return '';
+  let text = '';
+  for (const op of ops) {
+    const insert = (op as { insert?: unknown }).insert;
+    if (typeof insert === 'string') text += insert;
+    else break; // a retain / delete / embed ends the leading prepend
+  }
+  return text;
+}
+
+/** Read the value at a JSON-Pointer `path` (e.g. `/docs/<id>/body/content`) in a plain doc state. */
+function valueAtPath(root: unknown, path: string): unknown {
+  let cur: unknown = root;
+  for (const token of path.split('/')) {
+    if (token === '') continue; // leading '' from the initial slash
+    if (cur == null || typeof cur !== 'object') return undefined;
+    const key = token.replace(/~1/g, '/').replace(/~0/g, '~');
+    cur = (cur as Record<string, unknown>)[key];
+  }
+  return cur;
+}
+
+/** Plain text of a `@txt` field's current value (`Delta` / `{ ops }` / `Op[]`), or null if not a delta. */
+function fieldPlainText(value: unknown): string | null {
+  const ops = toOps(value);
+  if (!ops) return null;
+  let text = '';
+  for (const op of ops) {
+    const insert = (op as { insert?: unknown }).insert;
+    if (typeof insert === 'string') text += insert;
+    else if (insert != null) text += '￼'; // embed placeholder — never matches real text
+  }
+  return text;
+}
+
+/**
+ * Thrown by {@link OTBranchManager.mergeBranch} when a merge would duplicate a field's existing
+ * content — the DAB-760 signature: a merge floor that undercounts the branch's seed replays
+ * seeded body text as a retain-less whole-field insert onto a non-empty target, doubling it.
+ * Refusing keeps the merge from silently doubling the document; merges are idempotent, so a
+ * corrected retry can proceed.
+ */
+export class MergeContentDuplicationError extends Error {
+  readonly code = 'MERGE_CONTENT_DUPLICATION';
+  constructor(
+    readonly sourceDocId: string,
+    readonly path: string,
+    readonly duplicatedChars: number
+  ) {
+    super(
+      `Merge aborted: would duplicate existing content at "${path}" of ${sourceDocId} — a ` +
+        `retain-less insert of ${duplicatedChars} chars matching the field's current head ` +
+        `(content-doubling signature; refusing to double the document).`
+    );
+    this.name = 'MergeContentDuplicationError';
+  }
+}
 
 /**
  * OT-specific branch manager implementation.
@@ -248,6 +317,11 @@ export class OTBranchManager implements BranchManager {
       batchId: branchId,
     }));
 
+    // Refuse a merge that would duplicate existing content (DAB-760): a floor that undercounts
+    // the branch's seed replays seeded body text as a retain-less whole-field insert, doubling
+    // the document. Checked against the branch's base revision before anything is committed.
+    await this.assertNoContentDuplication(sourceDocId, branchStartRevOnSource, changesToCommit);
+
     // Commit changes to source doc with error handling
     const committedMergeChanges = await wrapMergeCommit(branchId, sourceDocId, async () => {
       return (await this.patchesServer.commitChanges(sourceDocId, changesToCommit)).changes;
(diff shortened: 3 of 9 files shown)


src/server/OTBranchManager.ts
```diff
@@ -260,6 +334,62 @@ export class OTBranchManager implements BranchManager {
...
+        `[OTBranchManager] content-doubling guard could not reconstruct ${sourceDocId}@${baseRev}; ` +
+          `skipping check:`,
+        error
+      );
+      return;
+    }
+
+    for (const { path, inserted } of candidates) {
+      const current = fieldPlainText(valueAtPath(state, path));
+      if (current != null && current.startsWith(inserted)) {
```
RyanHavoc: This refuses legitimate merges, and the refusal is permanent.

`@dabble/delta` normalizes every replacement to insert-before-delete, so `leadingInsertText` returns the *entire* pasted body for a paste-over-selection at position 0:

```
new Delta().delete(10).insert('Chapter one. The grey cat')
  -> [{"insert":"Chapter one. The grey cat"},{"delete":10}]
```

`current.startsWith(inserted)` then fires whenever the pasted text is a prefix of the old. Probed through the real `mergeBranch` path on a branch with a **correct** floor (`manager.createBranch('doc1', 1)`, server-materialized seed):

```
trim-tail   select-all, paste back a trimmed version keeping the opening -> MergeContentDuplicationError | body unchanged
re-paste    select-all, paste identical content                          -> MergeContentDuplicationError | body unchanged
delete+undo delete the opening paragraph, Ctrl+Z, same merge batch       -> MergeContentDuplicationError
```

Delete-first shapes (`[{delete:n},{insert:...}]`) are safe — `leadingInsertText` returns `''` there. It is the insert-first shape that breaks, and that is the one Delta actually emits.

The check is unconditional with no override, and `lastMergedRev` never advances on a throw, so a retry throws identically — a writer who trims a scene in an editor copy can never merge that branch without a code change. It also isn't gated on the condition it backstops: all three probes above ran on branches whose floor was correct.

Root cause shared with the tail-chunk gap I noted below: the check asks "does this op's leading insert prefix the base field?" rather than "does applying this whole batch leave the field containing its prior content twice?". Composing the batch's per-path deltas and comparing the result against the current head would let undo, truncation and moves net out, and would catch the cases this misses. Whatever shape it takes, an explicit override looks worth having, since today a false positive is unrecoverable.

src/server/OTBranchManager.ts
```diff
@@ -260,6 +334,62 @@ export class OTBranchManager implements BranchManager {
...
+   */
+  private async assertNoContentDuplication(sourceDocId: string, baseRev: number, changes: Change[]): Promise<void> {
+    // Cheap first pass: a change can only produce the doubling signature via a `@txt` op that
+    // *prepends* a substantial run of text (no leading retain). Everything else is skipped.
+    const candidates: { path: string; inserted: string }[] = [];
+    for (const change of changes) {
+      for (const op of change.ops) {
+        if (op.op !== '@txt' || typeof op.path !== 'string') continue;
+        const inserted = leadingInsertText(op.value);
+        if (inserted.length >= MERGE_DUP_MIN_CHARS) candidates.push({ path: op.path, inserted });
```
RyanHavoc: The guard's catch rate is a function of *how badly* the floor undercounts, and it misses the near-misses.

`breakTextOp` splits an oversized body into per-chunk changes where chunks 2..N are bare `[{insert: chunkN}]` with no leading retain. Each still inserts at position 0 and corrupts the doc, but only chunk 1 is a prefix of the field head — so `startsWith` is false for every tail chunk.

Reproduced using this PR's own `breakChangesIntoBatches` recreate setup, with the floor off by **one** rev instead of many:

```
merge succeeded, no error raised
body 31,800 -> 34,401 chars
```

Test D only passes because its floor is 2, so chunk 1 happens to be included in the replay. A floor that undercounts by a little produces silent corruption; a floor that undercounts by a lot is caught. That's the opposite of the ordering you'd want from a backstop.

src/server/OTBranchManager.ts
```diff
@@ -260,6 +334,62 @@ export class OTBranchManager implements BranchManager {
...
+   * content that is already present rather than adding anything new. New docs/fields the branch
+   * introduced have no base value and are correctly left alone.
+   */
+  private async assertNoContentDuplication(sourceDocId: string, baseRev: number, changes: Change[]): Promise<void> {
+    // Cheap first pass: a change can only produce the doubling signature via a `@txt` op that
+    // *prepends* a substantial run of text (no leading retain). Everything else is skipped.
+    const candidates: { path: string; inserted: string }[] = [];
+    for (const change of changes) {
+      for (const op of change.ops) {
+        if (op.op !== '@txt' || typeof op.path !== 'string') continue;
```
RyanHavoc: Only `@txt` is inspected, so a whole-field `replace`/`add` carrying doubled content passes straight through. Probed:

```
{ op: 'replace', path: '/docs/d1/body/content', value: { ops: [{ insert: BODY + BODY }] } }
-> merge succeeded, source body verifiably doubled
```

Relevant rather than theoretical because `breakLargeValueOp` emits a structural `replace` alongside the `@txt` ops, so replace-shaped seed fragments are in the same family as the shape being guarded.

src/server/OTBranchManager.ts
```diff
@@ -260,6 +334,62 @@ export class OTBranchManager implements BranchManager {
...
+   * Cheap in the common case: only reconstructs the base state when a change actually carries a
+   * substantial leading text insert. An ordinary edit — a leading retain, a short insert, or a
+   * delete-then-insert replacement — never reaches the state read.
+   *
+   * `baseRev` is the source revision the branch diverged from. A healthy branch's base equals
+   * the source there, so a candidate op that re-inserts the field's existing head is duplicating
+   * content that is already present rather than adding anything new. New docs/fields the branch
+   * introduced have no base value and are correctly left alone.
+   */
+  private async assertNoContentDuplication(sourceDocId: string, baseRev: number, changes: Change[]): Promise<void> {
```
RyanHavoc: `baseRev` here is `branchStartRevOnSource` — the branch point, not the state the merge actually lands on. That's stale for any second merge of the same branch, and for any source that moved on independently.

Probed both directions:

```
false positive: merge 1 deletes d1's body from main; writer re-pastes it on the branch;
                merge 2 refused — the base state still shows the old head, though the
                source no longer contains it. Content is now unmergeable.
false negative: a genuine doubling relative to the *current* head goes unnoticed when
                the base differs from it.
```

Comparing against the source's current head rather than the branch point would fix both, and is also what the "already present" framing in the docblock actually describes.

src/server/OTBranchManager.ts
```diff
@@ -260,6 +334,62 @@ export class OTBranchManager implements BranchManager {
...
+    }
+    if (candidates.length === 0) return;
+
+    // Only now pay for a state reconstruction. A guard failure here must not block healthy
+    // merges, so a reconstruction error is logged and treated as "cannot check" rather than
+    // fatal — the primary defense is a correct merge floor, not this backstop.
+    let state: unknown;
+    try {
+      ({ state } = await getStateAtRevision(this.store, sourceDocId, baseRev, { reconstruction: {} }));
+    } catch (error) {
```
RyanHavoc: A transient failure here silently disables the guard and lets the doubling merge proceed.

`getLatestMainVersion` throws `StatusError(503)` when a version's state blob is missing or unservable — exactly the retryable case. This catches it, warns, and returns, so "cannot check" is treated as "checked, fine".

Given the refusal is otherwise permanent and unrecoverable, the asymmetry seems backwards: a reconstruction error is the one case where a retryable throw costs little (the caller retries) and passing costs a doubled manuscript.

src/server/OTBranchManager.ts
```diff
@@ -248,6 +317,11 @@ export class OTBranchManager implements BranchManager {
       batchId: branchId,
     }));
 
+    // Refuse a merge that would duplicate existing content (DAB-760): a floor that undercounts
+    // the branch's seed replays seeded body text as a retain-less whole-field insert, doubling
+    // the document. Checked against the branch's base revision before anything is committed.
+    await this.assertNoContentDuplication(sourceDocId, branchStartRevOnSource, changesToCommit);
```
RyanHavoc: "before anything is committed" isn't accurate — the version-copy loop above ([`OTBranchManager.ts:304`](https://github.com/dabblewriter/patches/blob/6fc5cb5c9958c56dbd4d4373c9ac93138dc887f3/src/server/OTBranchManager.ts#L303-L305)) has already run `this.store.createVersion(sourceDocId, copy, changes)` by the time the guard is consulted.

Probe on a refused merge:

```
source versions BEFORE: []
source versions AFTER : [ '0VPlP2ImLHO09Xns/BranchSession/branch' ]
lastMergedRev before/after: undefined / undefined
LEAKED VERSION COPIES: 1
```

Changes and the watermark are correctly untouched, but each refused attempt orphans branch version copies (`origin: 'branch'`, `groupId: branchId`) onto the source doc, where they surface in version history for a merge that never happened. Because refusal is permanent, so are the orphans, and they accumulate per retry.

The guard only needs `changesToCommit`'s ops and `baseRev`, both available right after `listChanges` — moving the call above the version loop makes the comment true and the refusal genuinely side-effect-free. Neither test B nor D asserts any side effect of refusal, only `docBody`.

src/server/OTBranchManager.ts
```diff
@@ -25,6 +26,74 @@ import type { BranchingStoreBackend, OTStoreBackend } from './types.js';
...
+}
+
+/**
+ * Thrown by {@link OTBranchManager.mergeBranch} when a merge would duplicate a field's existing
+ * content — the DAB-760 signature: a merge floor that undercounts the branch's seed replays
+ * seeded body text as a retain-less whole-field insert onto a non-empty target, doubling it.
+ * Refusing keeps the merge from silently doubling the document; merges are idempotent, so a
+ * corrected retry can proceed.
+ */
+export class MergeContentDuplicationError extends Error {
```
RyanHavoc, line 286: This isn't exported from the package's public server entry, so a consuming server can't `instanceof` it.

Its three siblings are ([`src/server/index.ts:39-41`](https://github.com/dabblewriter/patches/blob/6fc5cb5c9958c56dbd4d4373c9ac93138dc887f3/src/server/index.ts#L38-L42)):

```ts
export { RevConflictError } from './RevConflictError.js';
export { DuplicateChangeIdsError } from './DuplicateChangeIdsError.js';
export { ApplyChangesError } from '../algorithms/ot/shared/applyChanges.js';
```

`package.json` exposes only `./server` with no deeper path, so the only discrimination left to pup is string-matching `err.name` or reading the untyped `err.code`. Since the whole point of a dedicated class is to let the consumer tell "refused, repair the floor and retry" apart from a generic merge failure, it needs the export line alongside the others.

src/server/OTBranchManager.ts
```diff
@@ -25,6 +26,74 @@ import type { BranchingStoreBackend, OTStoreBackend } from './types.js';
  */
 type OTBranchStore = OTStoreBackend & BranchingStoreBackend;
 
+/**
+ * Minimum length of a re-inserted text run we treat as content *duplication* rather than a
+ * legitimate leading edit. Bodies/descriptions that double are far longer than this; a short
+ * genuine prepend never reaches it.
+ */
+const MERGE_DUP_MIN_CHARS = 16;
```
RyanHavoc: Flagging this against the repo's own scope boundary ([`CLAUDE.md`](https://github.com/dabblewriter/patches/blob/6fc5cb5c9958c56dbd4d4373c9ac93138dc887f3/CLAUDE.md#L19-L37)):

> Patches is a **generic primitives library**. It is composed into other systems (Pup, app-specific clients) — it is not itself an application. […] If a contribution adds something that "really only makes sense for Dabble" (or for any specific app), it does not belong here.

A hard-coded 16-character threshold deciding that text-shaped content is "duplicated", and unilaterally refusing the merge on that basis, is a policy call about one app's content — the kind CLAUDE.md places in the consuming server. Making it opt-in (an `OTBranchManager` option carrying the threshold and the refuse-vs-warn choice) would keep the mechanism here and the policy in pup, and would also give the false positives above an escape hatch.

Separately, `DAB-760` ticket refs appear in shipped source comments here and at `:320`, `:338`, `PatchesBranchClient.ts:177`, `PatchesDoc.ts:30`. Not load-bearing, but they read oddly in a library published for general use — the failure mode is worth describing without the ticket number.

src/server/OTBranchManager.ts
```diff
@@ -25,6 +26,74 @@ import type { BranchingStoreBackend, OTStoreBackend } from './types.js';
...
+
+/** The run of text a `@txt` delta prepends at position 0 (empty when it begins with retain/delete). */
+function leadingInsertText(value: unknown): string {
```
RyanHavoc, line 82: Breaking at the first non-string insert leaves any field whose body opens with an embed unguarded. Probed: body `[{insert:{image}},{insert:BODY1}]`, branch re-inserts the same at position 0 — merge succeeds, no error, doc doubled.

Minor on its own, but it's the same class of gap as the tail-chunk case: any scene opening with an image gets no protection at all.

src/server/OTBranchManager.ts
```diff
@@ -25,6 +26,74 @@ import type { BranchingStoreBackend, OTStoreBackend } from './types.js';
...
+  for (const op of ops) {
+    const insert = (op as { insert?: unknown }).insert;
+    if (typeof insert === 'string') text += insert;
+    else break; // a retain / delete / embed ends the leading prepend
+  }
+  return text;
+}
+
+/** Read the value at a JSON-Pointer `path` (e.g. `/docs/<id>/body/content`) in a plain doc state. */
+function valueAtPath(root: unknown, path: string): unknown {
```
RyanHavoc, line 105: `if (token === '') continue` skips *every* empty pointer token, not just the leading one:

```
'/'.split('/')     -> ['', '']
'/a//b'.split('/') -> ['', 'a', '', 'b']
```

Per RFC 6901 those reference the empty-string key `""`, so the loop resolves to the parent node instead and the guard checks the wrong value. Skipping only index 0 (or `path.split('/').slice(1)`) handles it.

The escape order on the next line is right — `~1` before `~0` — no issue there.

RyanHavoc (commented):
Reviewed the seeding fix and the merge guard separately, and probed both against the real `mergeBranch` / seeding paths rather than reading alone. Checks are green on the branch (43 tests, `tsc`, eslint, prettier all clean), so nothing below is a lint or typecheck nit.

The root-cause analysis is excellent and test D is a strong regression test for the desync. Two things I'd want resolved before this lands:

- The guard refuses legitimate merges, permanently and with no override — reproduced on branches whose floor is **correct**, so it isn't gated on the condition it backstops. Details inline on `OTBranchManager.ts`.
- `contentStartRev` is still derived from a different config object than the one that assigns the revs, so the desync can reappear even with the seeding fix. Details inline on `PatchesBranchClient.ts`.

The rest are smaller: the guard's blind spots (tail chunks, `replace` ops, embeds), a refused merge still orphaning version copies onto the source doc, and the error class not being exported for pup to catch.

src/client/PatchesBranchClient.ts
```diff
@@ -169,12 +169,19 @@ export class PatchesBranchClient {
...
-      initChanges = breakChanges(initChanges, options.maxStorageBytes, options.sizeCalculator);
-    }
+    const initChanges = breakChangesIntoBatches([rootReplace], {
+      maxPayloadBytes: options.maxPayloadBytes,
+      maxStorageBytes: options.maxStorageBytes,
+      sizeCalculator: options.sizeCalculator,
+    }).flat();
 
     // contentStartRev is the first revision after all init changes
     const contentStartRev = initChanges[initChanges.length - 1].rev + 1;
```
RyanHavoc: This makes the seed split match `PatchesSync`, which is the right fix — but `contentStartRev` is still derived from a *different* config object than the one that actually assigns the revs, so the desync can reappear.

`initChanges` is computed from `patches.docOptions`, then persisted at `:205` via `algorithm.handleDocChange(branchDocId, change.ops, undefined, {})` — passing only `.ops`, discarding both the computed rev and the returned `Change[]`. `OTAlgorithm._createChangesFromOps` ([`OTAlgorithm.ts:627`](https://github.com/dabblewriter/patches/blob/6fc5cb5c9958c56dbd4d4373c9ac93138dc887f3/src/client/OTAlgorithm.ts#L626-L630)) then runs its **own** split against `this._options.maxStorageBytes` — the options the algorithm was constructed with, not `patches.docOptions`.

Reproduced with the real `OTAlgorithm` + `OTInMemoryStore`:

```
patches.docOptions      = { maxStorageBytes: 3000, maxPayloadBytes: 6000, sizeCalculator: compressedSizeUint8 }
new OTAlgorithm(store,    { maxStorageBytes: 400,  sizeCalculator: compressedSizeUint8 })

seed persisted at revs 1..137
contentStartRev recorded as 8
-> 130 seed changes at/above the floor
```

That's DAB-760 again, post-fix. And the two configs do diverge in dw3 today: the editor-copy branch is created on the spoke `Patches` (built with no `docOptions`, `src/stores/patches.ts:343`) while the change is persisted by the hub's `OTAlgorithm` (built with `{maxStorageBytes: 900_000, compressedSizeUint8}`, `src/services/hub/PatchesHubService.ts:380`). 900KB *compressed* is loose enough that a 1MB-uncompressed piece won't normally re-split, so it's latent rather than routinely firing — but the fix's correctness currently rests on that coincidence, and nothing enforces it.

`handleDocChange` returns the changes it persisted. Collecting those and deriving `contentStartRev` from the last actually-persisted rev would make this correct by construction, independent of how many config sources disagree.

src/net/PatchesSync.ts
```diff
@@ -144,7 +144,7 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
     }
 
     // Use options if provided, otherwise fall back to patches.docOptions
-    this.maxPayloadBytes = options?.maxPayloadBytes;
+    this.maxPayloadBytes = options?.maxPayloadBytes ?? patches.docOptions?.maxPayloadBytes;
```
RyanHavoc, line 147: This closes a real gap, but the fallback only applies when the explicit option is *absent* — an explicit `PatchesSync` option still wins, while `_createBranchOffline` only ever reads `patches.docOptions`. So a stricter explicit limit re-splits the seed after `contentStartRev` has been frozen.

```
docOptions { maxPayloadBytes: 6000, maxStorageBytes: 3000, compressed }
new PatchesSync(..., { maxPayloadBytes: 2000 })

7-change seed re-splits to 23 changes, contentStartRev = 8
```

It isn't confined to the wire either: `flushDoc` (`PatchesSync.ts:857-870`) sees `flattened.length !== pending.length` and calls `replacePendingChanges`, rewriting the **store** to the renumbered changes — 16 of the 23 landed at rev >= `contentStartRev`. And the floor can never be corrected afterwards: `updateBranch` omits `contentStartRev` and `syncPendingBranchMetas` (`:363`) strips it.

Not live in dw3 (`PupSync` passes no size options), and the doc comment on `PatchesDocOptions.maxPayloadBytes` does warn about it — but a comment feels thin for a silent whole-document-doubling failure. Deriving the floor from the persisted revs (see my note on `PatchesBranchClient.ts`) would make this one moot too, since the split would no longer need to be predicted in advance.

tests/client/PatchesBranchClient.spec.ts
```diff
@@ -161,6 +164,36 @@ describe('PatchesBranchClient', () => {
       expect(patches.onChange.emit).toHaveBeenCalledWith('my-branch');
     });
 
+    it('splits the seed for the wire so contentStartRev matches the committed span (DAB-760)', async () => {
```
RyanHavoc: This test can't catch the bug class it's guarding.

`mockAlgorithm` is `{ handleDocChange: vi.fn().mockResolvedValue([]) }` (`:47-49`), so `_createChangesFromOps` never executes and no persisted rev is ever observed. The assertions are the call *count*, and that `contentStartRev === wireSeed.last.rev + 1` where `wireSeed` is re-derived from `breakChangesIntoBatches` — the same function under test. It would pass unchanged in the 137-revs-vs-floor-of-8 scenario I described on `PatchesBranchClient.ts`.

Swapping in a real `OTAlgorithm` + `OTInMemoryStore` and asserting against the revs actually in the store makes it real — I confirmed it still passes for the matched-config case.

More broadly, nothing in this PR wires fixed `_createBranchOffline` -> real server merge -> correct content. Server test A hand-sets a correct floor computed via `breakChanges` directly rather than via the fixed client path, and B and D both assert *refusal*. So the claim "the seeding fix prevents doubling" currently rests entirely on the guard tests. Test D is a genuinely good regression test for the desync itself — it's the end-to-end no-doubling case that's missing.

tests/server/OTBranchManager.integration.spec.ts
```diff
@@ -604,3 +607,212 @@ describe('OTBranchManager integration', () => {
...
+  return { op: '@txt' as const, path, value: ops };
+}
+
+/** A one-op `@txt` change. */
+function txtChange(id: string, baseRev: number, path: string, ops: any[]): ChangeInput {
+  return { id, baseRev, rev: baseRev + 1, ops: [txtOp(path, ops)] };
+}
+
+/** Concatenated text of a doc's `/body` delta field, tolerant of Delta / {ops} / Op[] shapes. */
+function bodyText(state: any): string {
```
RyanHavoc: `bodyText` is never called — every call site uses the nested `docBody` below. ESLint doesn't flag it (exit 0 on this file), so it'd sit here indefinitely.

RyanHavoc:
Pushed two commits addressing the review findings (07a630b, ee676a1 for the 0.21.1 bump).

**Client floor (root cause):** `_createBranchOffline` now persists the seed first and derives `contentStartRev` from the revisions the algorithm actually assigned, instead of predicting the split from `patches.docOptions`. The algorithm's own `maxStorageBytes` re-split can no longer desync the floor, regardless of how many config sources disagree. The `breakChangesIntoBatches` pre-split stays as the flush-stability layer so the persisted pieces aren't renumbered at sync time. Branch meta creation moved after seeding (rollback covers both orderings, with a new test for a create failure after seeding).

**Guard rework:** the check no longer asks "does this op's leading insert prefix the base field?". It walks each affected text field through the whole batch — composing `@txt` deltas and field-level `replace`/`add`/`remove` in order — against the source's **current head**, tracking which spans of the head survive and which runs the batch inserted. A field is flagged when the final content contains the head content twice (a field set to an already-doubled value), or when a substantial inserted run sitting ahead of all surviving head content duplicates text that still survives. Consequences, matching the inline findings:

- Insert-before-delete shapes (trim-tail paste, identical re-paste, delete+undo in one batch) net out and merge — each is now a passing test.
- Tail seed pieces are caught even when the floor is off by one (bare mid-body inserts are substrings of the surviving head, no prefix relation needed) — tested.
- Whole-field `replace`/`add` values are inspected, including deltas nested inside larger values; a doubled value is refused while an ordinary rewrite merges — tested.
- Comparing against the current head fixes both stale-`baseRev` directions; re-adding content the source has since deleted merges on a second merge — tested.
- A head-reconstruction failure now propagates (retryable) instead of silently skipping the check — tested.
- Fields opening with an embed are protected; the leading-insert scan treats embeds as placeholder content instead of stopping — tested.

**Policy and mechanics:** per the CLAUDE.md scope boundary the guard is now opt-in — `new OTBranchManager(store, server, { contentDuplicationGuard: { action: 'refuse' | 'warn', minLength } })` (default threshold 64 chars), with a per-merge override `mergeBranch(id, { contentDuplicationGuard: 'off' | 'warn' | 'refuse' })` as the recovery escape hatch for branches with an already-bad floor. The old third constructor param (`maxPayloadBytes`) still works. The guard runs before the version-copy loop, so a refused merge orphans no version copies (asserted in test B). `MergeContentDuplicationError` and the option types are exported from `@dabble/patches/server`. The JSON-Pointer walker only skips the leading empty token now, and ticket refs are out of shipped source comments.

**Tests:** the client spec's floor tests run against a real `OTAlgorithm` + `OTInMemoryStore` and assert against the revs actually in the store, including a diverging-config case that fails under the old prediction. A new end-to-end test runs the fixed client seeding into a real server merge with the guard armed and asserts the content is neither doubled nor refused. `bodyText` is gone. Full suite 3203 passing, tsc/eslint/prettier clean.

pup will need `{ contentDuplicationGuard: { action: 'refuse' } }` (or `'warn'` first, if we want an observation period) when it picks up 0.21.1, since the guard no longer arms itself.

PR #116, opened 2026-07-19
test: pin @txt transform behavior at array-element sub-paths

RyanHavoc:
**TL;DR:** Tests only — no runtime change, no version bump, no publish needed. Pins the transform/apply behavior of `@txt` ops addressed inside array elements (`/list/<i>/text`), which dabble-writer-3.0#1202 now depends on for loose-note (`doc.notes[]`) content batching.

17 tests covering:
- Positional shifting against concurrent `add`/`remove` at lower/same/higher indexes
- Following the op's own element through a `move` (both directions), and correct shifts when *other* elements move across it
- Dropping when the element is removed or wholly replaced; surviving (and composing after) a same-path wholesale `replace` — the mixed-fleet case, matching existing object-path semantics
- Concurrent same-path `@txt` transforming as text (not LWW)
- Apply: composing onto existing `{ ops }` content and seeding an empty document at an array sub-path

All pass against current main. pup (0.20.1) and dw3 (0.21.0) share identical json-patch code (the 0.20.1→0.21.0 delta was networking-only), so both sides of the sync already behave as pinned.

### DIFF
--- a/tests/json-patch/txt-array-subpath.spec.ts
+++ b/tests/json-patch/txt-array-subpath.spec.ts
@@ -1,3 +1,4 @@
+import { Delta } from '@dabble/delta';
 import { describe, expect, it } from 'vitest';
 import { applyPatch } from '../../src/json-patch/applyPatch.js';
 import { transformPatch } from '../../src/json-patch/transformPatch.js';
@@ -103,18 +104,41 @@ describe('@txt at array-element sub-paths', () => {
       ).toEqual([{ op: '@txt', path: '/list/1/text', value: [{ retain: 5 }, { insert: 'second' }] }]);
     });
 
+    it('threads through a batch of sequential @txt ops on the same sub-path', () => {
+      // Ops within one patch each apply after the previous, so ours has to be advanced over
+      // every one of them to stay in the same coordinate space — see the delta threading in
+      // `text.transform`. A queue that batches content edits emits exactly this shape, and a
+      // single-op test can't tell correct threading apart from transforming only against the
+      // first op.
+      const theirs = [
+        { op: '@txt', path: '/list/1/text', value: [{ retain: 1 }, { insert: 'AAA' }] },
+        { op: '@txt', path: '/list/1/text', value: [{ retain: 2 }, { insert: 'B' }] },
+      ];
+      const ours = [{ op: '@txt', path: '/list/1/text', value: [{ retain: 2 }, { insert: 'S' }] }];
+
+      const transformed = transformPatch(state, theirs, ours);
+      expect(transformed).toEqual([{ op: '@txt', path: '/list/1/text', value: [{ retain: 6 }, { insert: 'S' }] }]);
+
+      // 'bbb\n' + their two edits = 'bABAAbb\n', where the base chars now sit at 0, 5 and 6.
+      // Ours still lands between the two it was written between, not at its original (now
+      // stale) offset of 2.
+      const converged = applyPatch(applyPatch(state, theirs), transformed) as typeof state;
+      expect(converged.list[1].text).toEqual({ ops: [{ insert: 'bABAAbSb\n' }] });
+    });
+
     it('survives a wholesale replace of the same sub-path and composes onto the new content', () => {
-      // Same semantics as `@txt` vs `replace` at an object path (see transformPatch.spec):
+      // Same semantics as `@txt` vs `replace` at an object path (see transform.spec.ts):
       // the text op is not clobbered — it applies after the replace. This is what keeps a
       // mixed fleet safe: an old client replacing the whole text object concurrent with a
       // new client's incremental edit loses neither write.
-      expect(
-        transformPatch(
-          state,
-          [{ op: 'replace', path: '/list/1/text', value: { ops: [{ insert: 'rewritten\n' }] } }],
-          [txt('/list/1/text')]
-        )
-      ).toEqual([txt('/list/1/text')]);
+      const replace = [{ op: 'replace', path: '/list/1/text', value: { ops: [{ insert: 'rewritten\n' }] } }];
+      const transformed = transformPatch(state, replace, [txt('/list/1/text')]);
+      expect(transformed).toEqual([txt('/list/1/text')]);
+
+      // Surviving is only half of it — the op has to land on the replaced content, not the
+      // content it was written against. Neither write is lost.
+      const converged = applyPatch(applyPatch(state, replace), transformed) as typeof state;
+      expect(converged.list[1].text).toEqual({ ops: [{ insert: 'xrewritten\n' }] });
     });
 
     it('drops when its whole element was replaced', () => {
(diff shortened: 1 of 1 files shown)


tests/json-patch/txt-array-subpath.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    it('stays put when another element moves entirely above it', () => {
+      expect(transformPatch(state, [{ op: 'move', from: '/list/2', path: '/list/3' }], [txt('/list/1/text')])).toEqual([
+        txt('/list/1/text'),
+      ]);
+    });
+  });
+
+  describe('against concurrent writes to the same element', () => {
+    it('transforms against a concurrent @txt at the same sub-path as text, not LWW', () => {
```
RyanHavoc, line 97: Coverage gap worth closing: all 17 cases use single-op patches on both sides.

`text.transform` deliberately advances its delta across successive other-ops (`thisDelta = otherDelta.transform(thisDelta, false)` — [text.ts#L56-L72](https://github.com/dabblewriter/patches/blob/1a766cda1c034e74f2eefeec160c3b27d1bacb45/src/json-patch/ops/text.ts#L56-L72)), and that advancing is only covered at an object path today: [sequential-transform.spec.ts#L99-L110](https://github.com/dabblewriter/patches/blob/1a766cda1c034e74f2eefeec160c3b27d1bacb45/tests/json-patch/sequential-transform.spec.ts#L99-L110).

Since dabble-writer-3.0#1202 batches loose-note content through an id-keyed queue, two `@txt` ops in one patch at an array sub-path is the case this file exists to pin and is the one thing it doesn't.

tests/json-patch/txt-array-subpath.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+      expect(
+        transformPatch(
+          state,
+          [{ op: '@txt', path: '/list/1/text', value: [{ insert: 'first' }] }],
+          [{ op: '@txt', path: '/list/1/text', value: [{ insert: 'second' }] }]
+        )
+      ).toEqual([{ op: '@txt', path: '/list/1/text', value: [{ retain: 5 }, { insert: 'second' }] }]);
+    });
+
+    it('survives a wholesale replace of the same sub-path and composes onto the new content', () => {
```
RyanHavoc, line 129: The name claims two things but the assertion only covers the first — nothing is applied, so "composes onto the new content" is untested.

The behavior is right (applying the transformed op after the replace gives `xrewritten\n`), but composition is the half the mixed-fleet argument in the PR description rests on. An `applyPatch` follow-through would actually pin it.

tests/json-patch/txt-array-subpath.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+      // Same semantics as `@txt` vs `replace` at an object path (see transformPatch.spec):
```
RyanHavoc: `transformPatch.spec` doesn't exist. The equivalent object-path assertions live in `tests/json-patch/transform.spec.ts`: [#L890-L896](https://github.com/dabblewriter/patches/blob/1a766cda1c034e74f2eefeec160c3b27d1bacb45/tests/json-patch/transform.spec.ts#L890-L896) (`replace` as this-op) and [#L1404-L1410](https://github.com/dabblewriter/patches/blob/1a766cda1c034e74f2eefeec160c3b27d1bacb45/tests/json-patch/transform.spec.ts#L1404-L1410) (`@txt` as this-op).

tests/json-patch/txt-array-subpath.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+  });
+
+  describe('against concurrent remove', () => {
+    it('shifts down when an element is removed at a lower index', () => {
+      expect(transformPatch(state, [{ op: 'remove', path: '/list/0' }], [txt('/list/1/text')])).toEqual([
+        txt('/list/0/text'),
+      ]);
+    });
+
+    it('drops when its own element is removed', () => {
```
RyanHavoc, line 53: This and `drops when its whole element was replaced` (L120) restate existing coverage — same ops, same `/1/text` sub-path form: [transform.spec.ts#L1237-L1241](https://github.com/dabblewriter/patches/blob/1a766cda1c034e74f2eefeec160c3b27d1bacb45/tests/json-patch/transform.spec.ts#L1237-L1241) and [#L1245-L1249](https://github.com/dabblewriter/patches/blob/1a766cda1c034e74f2eefeec160c3b27d1bacb45/tests/json-patch/transform.spec.ts#L1245-L1249).

Reasonable to keep for a self-contained suite — flagging so it's a deliberate call. The add/remove/move shifting cases are genuinely new; existing `@txt` shift coverage is all at exact indexes, never a sub-path.

tests/json-patch/txt-array-subpath.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+      expect(result.list[1].text).toEqual({ ops: [{ insert: 'xbbb\n' }] });
+      expect(result.list[0]).toBe(state.list[0]); // untouched siblings keep identity
+    });
+
+    it('seeds an empty document when the element has no content yet', () => {
+      const bare = { list: [{ id: 'a' }] };
+      const result = applyPatch(bare, [{ op: '@txt', path: '/list/0/text', value: [{ insert: 'x' }] }]) as {
+        list: { id: string; text?: unknown }[];
+      };
+      expect(result.list[0].text).toEqual({ ops: [{ insert: 'x\n' }] });
```
RyanHavoc, line 165: `text.apply` seeds with `new Delta().insert('\n')` and stores a **Delta instance**, not a plain `{ ops }` object — see [text.ts#L27-L54](https://github.com/dabblewriter/patches/blob/1a766cda1c034e74f2eefeec160c3b27d1bacb45/src/json-patch/ops/text.ts#L27-L54). `toEqual` passes here; `toStrictEqual` would not.

Worth pinning the stored type if dabble-writer-3.0#1202 does structural comparison or serialization on these values — as written this wouldn't catch a change to it.

RyanHavoc (commented):
Reviewed the branch locally: 17/17 pass, full suite green (115 files / 3202 tests), `type:check` and `prettier --check` clean.

I also re-derived the expected indexes by hand and then applied the *transformed* patches to check the end state, rather than just the rewritten path strings. The moved-element case genuinely lands the text on the right element (`move /list/1 -> /list/3` then apply yields `b:"Xbbb\n"` in final position), and `replace` + `@txt` composes to `xrewritten\n` without losing either write. So the semantics being pinned here are correct.

No correctness defects. Five notes below — coverage and accuracy rather than bugs. Only the first is substantive.

RyanHavoc:
Pushed b75fc77 addressing the review feedback.

**Batched `@txt` threading** — added `threads through a batch of sequential @txt ops on the same sub-path`. Two `@txt` ops in one patch on the same array sub-path, which is the shape an id-keyed content queue emits and the one case the file didn't cover. It discriminates: transforming against only the first op yields `retain: 5`, against both yields `retain: 6`, and the test asserts 6, so it fails if the delta threading in `text.transform` stops advancing.

**Two tests now assert what their names claim** — the wholesale-replace case applies the transformed op and checks it composes onto the replaced content (`xrewritten\n`) instead of only surviving transform; the seed case pins that apply stores a Delta instance (`toEqual` can't distinguish that from a plain `{ ops }` literal) plus the structurally-cloned shape consumers actually read back.

**Stale reference** — `transformPatch.spec` corrected to `transform.spec.ts`.

Not changed: the two tests that overlap `transform.spec.ts:1239/1247`. Removing them would leave this suite covering add/remove/move but silently deferring the two drop cases elsewhere, which costs more in readability than the duplication does. Happy to drop them if you'd rather.

Gates: `type:check`, `lint`, `format:check` clean; full suite 115 files / 3203 tests passing.

PR #117, opened 2026-07-19
Extract SSE event buffering into a pluggable SSEEventStore

jacwright:
**TL;DR:** When a writer's connection blips, their app currently has to re-download the state of every project to be sure it missed nothing — slow and expensive. This change lets the server hand the reconnecting app exactly the events it missed instead. This PR adds the plumbing in the library; the companion pup PR stores the events in Redis so it works across the whole server fleet.

## What changed

`SSEServer` no longer owns event ids or the replay buffer. A new `SSEEventStore` interface owns both, plus the subscription record needed to rehydrate fan-out routing on another instance:

- `append(clientId, event, data)` → the store assigns the id written on the wire frame. Returning `null` (degraded store) delivers the frame **without** an id, so the client's `Last-Event-ID` cursor never advances past an unstored event.
- `replay(clientId, lastEventId)` → either a verified continuation (possibly empty) or `resync`, optionally carrying a store-assigned **re-anchor id** written on the resync frame so the browser's cursor moves into the store's current id space (base behavior wrote an id on resync for exactly this reason). All continuity decisions live in the store; the server no longer parses or compares ids (they're opaque strings end-to-end).
- `addSubscriptions` / `removeSubscriptions` / `loadSubscriptions` mirror the routing set so a reconnect landing on a cold instance can rebuild fan-out before replay. Mutations are serialized per client and `subscribe`/`addSubscriptions`/`unsubscribe` resolve once the mirror settles, so removes can't overtake adds and an acked unsubscribe is visible to cold hydration.
- `dropClient` is advisory: a shared store may no-op and rely on TTLs, because one instance's disconnect-expiry only proves the client hasn't reconnected _there_.

The default `InMemorySSEEventStore` reproduces the previous single-instance buffering behavior exactly — all 41 pre-existing SSEServer tests pass unmodified — so consumers that pass no `eventStore` see no change.

## Correctness details (from adversarial review)

- **Per-client pipeline**: appends and frame writes are serialized per client so store order always matches wire order — an id appended out of order could let the cursor leapfrog an undelivered event. The connect-time **replay read is issued inside the pipeline** too, so it lands after every queued append (whose events it must cover) and before anything enqueued later; the notify writer is captured at enqueue time so an event covered by a pending replay is delivered exactly once.
- **Store calls are timeout-bounded** (`storeTimeoutMs`, default 10s): a call that never settles degrades (append → id-less frame, replay → resync, hydration → empty) instead of permanently wedging the client's pipeline behind it.
- **Hydration vs. racing events**: routing for an event that arrives while a cold connect is still loading subscriptions is decided _inside_ the pipeline (after hydration), so the event is neither dropped nor left out of the store. An unsubscribe racing the hydration snapshot is recorded and can't be resurrected by the stale snapshot. Rehydration deliberately does not re-run authorization — only authorized subscribes reach the store, and the store TTL bounds the grant.
- **`releaseClient(clientId)`** (new): drops one instance's local ghost state for a client whose stream has been claimed elsewhere — cancels the expiry timer _without_ touching the store. Without it, a ghost instance would double-append every relayed event and its buffer-TTL cleanup would destroy a live client's shared state (this was the critical review finding). Queued pipeline tasks check state identity so a released ghost can't keep appending, and closes still owed for a force-closed stream are tallied server-side so a lagging close can't mark a fresh reconnection disconnected.
- Direct `sendToClient` frames are written synchronously and carry no id — they are never stored, so the boolean return is truthful and signaling latency never couples to store latency.

## Tests

3220 passing (36 new across two review rounds: pipeline ordering, replay-read-behind-appends, store timeouts, resync re-anchoring, epoch-cursor refusal, subscription mutation ordering, releaseClient lifecycle, orphaned closes, opaque-id replay, degraded-store id-less frames, hydration window, no-resurrection, TTL touch on reconnect, in-memory store parity). Type-check, lint, and format clean.

Consumed by the companion pup PR (`RedisSSEEventStore`), which should merge after this releases.

### DIFF
--- a/src/net/rest/SSEServer.ts
+++ b/src/net/rest/SSEServer.ts
@@ -19,6 +19,11 @@ interface ClientState {
   hydrating: Set<string> | null;
   /** Serializes this client's append→write pipeline and connect replay. */
   pipeline: Promise<void>;
+  /**
+   * Serializes this client's store subscription mutations so add/remove land
+   * in issue order (see the {@link SSEEventStore.addSubscriptions} contract).
+   */
+  subOps: Promise<void>;
   expiryTimer: ReturnType<typeof globalThis.setTimeout> | null;
   /**
    * Number of old connections force-closed by a reconnect whose close the
@@ -52,6 +57,15 @@ export interface SSEServerOptions {
    * original single-instance buffering behavior.
    */
   eventStore?: SSEEventStore;
+  /**
+   * Milliseconds to wait on an event store call before degrading it (append →
+   * id-less frame, replay → resync, hydration → empty set). Store calls ride
+   * each client's serialized pipeline, so a call that never settles (e.g. a
+   * Redis client that queues commands across a reconnect indefinitely) would
+   * otherwise wedge that client's event delivery permanently — and invisibly,
+   * since heartbeats bypass the pipeline. Default: 10000.
+   */
+  storeTimeoutMs?: number;
 }
 
 /**
@@ -102,15 +116,25 @@ export interface SSEServerOptions {
  */
 export class SSEServer {
   private clients = new Map<string, ClientState>();
+  /**
+   * Pending close callbacks for streams force-closed while their ClientState
+   * was being discarded (releaseClient / buffer-TTL cleanup). The per-state
+   * staleCloses counter dies with the state, so without this a lagging
+   * framework close for a released stream would mark a fresh reconnection
+   * (new state, staleCloses 0) as disconnected.
+   */
+  private orphanedCloses = new Map<string, number>();
   private heartbeatInterval: ReturnType<typeof globalThis.setInterval> | null = null;
   private heartbeatIntervalMs: number;
   private bufferTTLMs: number;
+  private storeTimeoutMs: number;
   private auth?: AuthorizationProvider;
   private eventStore: SSEEventStore;
 
   constructor(options?: SSEServerOptions) {
     this.heartbeatIntervalMs = options?.heartbeatIntervalMs ?? 30_000;
     this.bufferTTLMs = options?.bufferTTLMs ?? 300_000;
+    this.storeTimeoutMs = options?.storeTimeoutMs ?? 10_000;
     this.auth = options?.auth;
     this.eventStore =
       options?.eventStore ??
@@ -130,7 +154,8 @@ export class SSEServer {
    *
    * If `lastEventId` is provided (from the Last-Event-ID header on reconnect),
    * the event store decides continuity: a verified continuation is replayed,
-   * otherwise a `resync` event (no id) tells the client to do a full re-sync.
+   * otherwise a `resync` event tells the client to do a full re-sync (carrying
+   * a store-assigned re-anchor id when the store provides one).
    *
    * The framework MUST call `disconnect(clientId)` when the response closes.
    */
@@ -156,6 +181,7 @@ export class SSEServer {
         subscriptions: new Set(),
         hydrating: null,
         pipeline: Promise.resolve(),
+        subOps: Promise.resolve(),
         expiryTimer: null,
         staleCloses: 0,
       };
@@ -175,8 +201,13 @@ export class SSEServer {
       // hydrate the routing set from the store before replay so post-replay
       // events fan out. Unsubscribes racing the load are collected in
       // `hydrating` and excluded when the (stale) snapshot is applied.
+      // Rehydration deliberately does NOT re-run the AuthorizationProvider:
+      // only authorized subscribe() calls ever reach the store, and the
+      // store's retention TTL bounds how long that grant can outlive a
+      // revocation (matching base behavior, where revocation never stopped an
+      // already-live feed).
       client.hydrating = new Set();
-      const subscriptions = this.eventStore.loadSubscriptions(clientId).catch((): string[] => []);
+      const subscriptions = this._guardStoreCall(this.eventStore.loadSubscriptions(clientId), []);
       this._enqueue(client, async () => {
         const stored = await subscriptions;
         const removed = client.hydrating ?? new Set<string>();
@@ -197,13 +228,21 @@ export class SSEServer {
     }
 
     if (lastEventId) {
-      // Issued now so the store sees the cursor before any subsequent appends.
-      const replay = this.eventStore.replay(clientId, lastEventId).catch((): SSEReplayResult => ({ type: 'resync' }));
       this._enqueue(client, async () => {
-        const result = await replay;
+        // The read is ISSUED inside the pipeline task: it lands after every
+        // append already queued (whose events the replay must cover — a read
+        // issued at connect time could execute before an in-flight append
+        // lands, permanently skipping an event that was committed to the
+        // store but dropped from the replaced wire) and before anything
+        // enqueued later (whose appends must not be double-delivered).
+        const result = await this._guardStoreCall<SSEReplayResult>(this.eventStore.replay(clientId, lastEventId), {
+          type: 'resync',
+        });
         if (client.writer !== writer) return; // Replaced while awaiting
         if (result.type === 'resync') {
-          this._writeFrame(writer, client.encoder, null, 'resync', '{}');
+          // A store-assigned re-anchor id moves the client's cursor into the
+          // store's current id space (see SSEReplayResult).
+          this._writeFrame(writer, client.encoder, result.id ?? null, 'resync', '{}');
         } else {
           for (const event of result.events) {
             this._writeFrame(writer, client.encoder, event.id, event.event, event.data);
@@ -223,6 +262,16 @@ export class SSEServer {
    * stored events are replayed.
    */
   disconnect(clientId: string): void {
+    // A close reported for a stream that was force-closed while its state was
+    // being discarded (releaseClient / TTL cleanup) — any state found now
+    // belongs to a NEWER connection and must not absorb this close.
+    const orphaned = this.orphanedCloses.get(clientId);
+    if (orphaned !== undefined) {
+      if (orphaned <= 1) this.orphanedCloses.delete(clientId);
+      else this.orphanedCloses.set(clientId, orphaned - 1);
+      return;
+    }
+
     const client = this.clients.get(clientId);
     if (!client) return;
 
@@ -274,18 +323,20 @@ export class SSEServer {
    * Never call this with client-supplied docIds that haven't passed
    * authorization.
    *
+   * Resolves once the store mirror has settled (see {@link _mirrorSubs}).
+   *
    * @returns The registered docIds, or null when the client is unknown on this
    *   instance — distinguishable from success so the caller can surface the
    *   failure (and the client can reconnect/retry) instead of the subscription
    *   being silently dropped.
    */
-  addSubscriptions(clientId: string, docIds: string[]): string[] | null {
+  async addSubscriptions(clientId: string, docIds: string[]): Promise<string[] | null> {
     const client = this.clients.get(clientId);
     if (!client) return null;
     for (const docId of docIds) {
       client.subscriptions.add(docId);
     }
-    this.eventStore.addSubscriptions(clientId, docIds).catch(SSEServer.noop);
+    await this._mirrorSubs(client, () => this.eventStore.addSubscriptions(clientId, docIds));
     return [...docIds];
   }
 
@@ -309,7 +360,7 @@ export class SSEServer {
       for (const docId of docIds) {
         client.subscriptions.add(docId);
       }
-      this.eventStore.addSubscriptions(clientId, docIds).catch(SSEServer.noop);
+      await this._mirrorSubs(client, () => this.eventStore.addSubscriptions(clientId, docIds));
       return [...docIds];
     }
 
@@ -329,24 +380,29 @@ export class SSEServer {
       }
     }
     if (allowed.length) {
-      this.eventStore.addSubscriptions(clientId, allowed).catch(SSEServer.noop);
+      await this._mirrorSubs(client, () => this.eventStore.addSubscriptions(clientId, allowed));
     }
     return allowed;
   }
 
   /**
    * Unsubscribe a client from documents.
+   *
+   * Resolves once the store mirror has settled — callers answering an HTTP
+   * unsubscribe should await it, so a reconnect landing on a cold instance
+   * after the response can't rehydrate a subscription the client was told is
+   * gone.
    */
-  unsubscribe(clientId: string, docIds: string[]): void {
+  unsubscribe(clientId: string, docIds: string[]): Promise<void> {
     const client = this.clients.get(clientId);
-    if (!client) return;
+    if (!client) return Promise.resolve();
     for (const docId of docIds) {
       client.subscriptions.delete(docId);
       // A pending hydration snapshot may still contain this docId — record the
       // removal so the apply can't resurrect it.
       client.hydrating?.add(docId);
     }
-    this.eventStore.removeSubscriptions(clientId, docIds).catch(SSEServer.noop);
+    return this._mirrorSubs(client, () => this.eventStore.removeSubscriptions(clientId, docIds));
   }
 
   /**
@@ -377,17 +433,23 @@ export class SSEServer {
       // are neither subscribed nor hydrating are skipped without a task.
       if (!client.subscriptions.has(docId) && !client.hydrating) continue;
 
+      // The writer is captured at ENQUEUE time: a task that ends up running
+      // after a reconnect must not write to the replacement stream, because
+      // the reconnect's replay task is queued behind this append and would
+      // deliver the same event again — the mismatch guard below drops the
+      // direct write and lets the replay deliver it exactly once.
+      const writer = client.writer;
       this._enqueue(client, async () => {
+        // State replaced or discarded (releaseClient / TTL cleanup / destroy)
+        // while queued — a released ghost must not keep appending to a shared
+        // store for a client that is live on another instance.
+        if (this.clients.get(clientId) !== client) return;
         if (!client.subscriptions.has(docId)) return;
         // Appends run inside the pipeline so each client's store order matches
         // its wire order — the replay cursor is the LAST id on the wire, so an
-        // id appended out of order could leapfrog an undelivered event. The
-        // writer is captured before the append: a stream replaced while the
-        // append is in flight is not written to — its replay (whose store read
-        // was issued after this append) delivers the event instead.
-        const writer = client.writer;
-        const eventId = await this.eventStore.append(clientId, event, data).catch(() => null);
-        if (!writer || client.writer !== writer) return; // Disconnected at append time, or replaced since
+        // id appended out of order could leapfrog an undelivered event.
+        const eventId = await this._guardStoreCall(this.eventStore.append(clientId, event, data), null);
+        if (!writer || client.writer !== writer) return; // Disconnected at enqueue time, or replaced since
         this._writeFrame(writer, client.encoder, eventId, event, data);
       });
     }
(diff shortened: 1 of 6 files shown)


src/net/rest/SSEServer.ts
```diff
@@ -118,69 +129,87 @@ export class SSEServer {
...
-      } else {
-        // Trim confirmed events — the client has acknowledged receipt up to lastId
-        const knownClient = client.nextEventId > 1;
-        client.buffer = client.buffer.filter(e => e.id > lastId);
-
-        if (!knownClient && lastId > 0) {
-          // Client claims a lastEventId but we have no history — buffer expired or server restarted
-          this._writeEvent(client, { id: client.nextEventId++, event: 'resync', data: '{}', timestamp: Date.now() });
+      // Issued now so the store sees the cursor before any subsequent appends.
+      const replay = this.eventStore.replay(clientId, lastEventId).catch((): SSEReplayResult => ({ type: 'resync' }));
```
RyanHavoc: The replay *read* is issued here, synchronously in `connect()`, while appends ride `client.pipeline` — only the result is applied inside the pipeline (`:202-212`). For an async store the read can therefore execute before appends the pipeline orders ahead of it.

That breaks the compensation `notify()` relies on. [The note at `:385-387`](https://github.com/dabblewriter/patches/blob/3d548a677b15a62eb6907b2b3fc3c1602372f42c/src/net/rest/SSEServer.ts#L384-L388) says the writer-mismatch drop is safe because "its replay (whose store read was issued after this append) delivers the event instead" — but that's issue order, not landing order. Nothing guarantees the append has landed, and for a task still queued it hasn't even been sent to the store.

Reproduced against this head with a store whose second append stalls mid-round-trip — client sitting on `id: 1` reconnects with `Last-Event-ID: 1` while event 2's append is in flight:

```
stream2 frames: "retry: 5000\n\nid: 3\nevent: changesCommitted\ndata: {"n":3}\n\n"
store contents:  [{"id":"1",...},{"id":"2",...},{"id":"3",...}]
```

Event 2 is committed to the store, written to neither stream, and the cursor advances to `3` — every later reconnect replays `> 3`, so it is unreachable permanently. A *verifying* store doesn't help: the gap is created server-side after the read, so [the escape hatch at `SSEEventStore.ts:18-23`](https://github.com/dabblewriter/patches/blob/3d548a677b15a62eb6907b2b3fc3c1602372f42c/src/net/rest/SSEEventStore.ts#L17-L24) doesn't cover this case.

`InMemorySSEEventStore.append` mutates synchronously, so none of this is visible to the added test suite — it only appears with the Redis store the PR exists to enable.

Moving the read inside the enqueued task fixes it. The read is then ordered behind everything already queued and still ahead of anything enqueued later, which is what the "sees the cursor before any subsequent appends" note is after:

```ts
this._enqueue(client, async () => {
  const result = await this.eventStore
    .replay(clientId, lastEventId)
    .catch((): SSEReplayResult => ({ type: 'resync' }));
  ...
```

Verified locally: the repro passes and all 151 `tests/net/rest` tests still pass, except `should replay a store continuation with its opaque ids verbatim` (`tests/net/rest/SSEServer.spec.ts:660`), which asserts `store.replay` was called synchronously after `connect()` and needs an `await` in front of the assertion.

src/net/rest/SSEServer.ts
```diff
@@ -427,26 +492,19 @@ export class SSEServer {
...
-    const cutoff = now - this.bufferWindowMs;
-    const idx = client.buffer.findIndex(e => e.timestamp >= cutoff);
-    if (idx > 0) {
-      client.buffer = client.buffer.slice(idx);
-    } else if (idx === -1) {
-      // All events are older than the window
-      client.buffer = [];
-    }
+  private _enqueue(client: ClientState, task: () => Promise<void>): void {
+    client.pipeline = client.pipeline.then(task).catch(SSEServer.noop);
```
RyanHavoc, line 558: No timeout and no bound on this chain. `notify`'s task awaits `eventStore.append` at `:389`; the `.catch()` covers rejection but not a hang — and ioredis has no `commandTimeout` by default and queues commands across a reconnect indefinitely. A store call that never settles leaves `client.pipeline` permanently pending, so every later `notify`, `sendToClient` and connect-`replay` for that client is dead.

Two things make that silent rather than self-healing:

- The heartbeat at `:528` writes straight to `client.writer`, bypassing the pipeline. `PatchesREST` resets `_lastEventAt` on each `heartbeat`, so the 75s liveness watchdog never fires and the stream stays `connected` — the client reports healthy while receiving zero events, and `PatchesSync` never runs its catch-up.
- Reconnecting can't clear it. `connect()` reuses the existing `ClientState` and therefore the existing `pipeline` (only the `isNew` branch builds a fresh `Promise.resolve()` at `:158`), and it clears `expiryTimer` on every reconnect — so each EventSource retry cancels the `bufferTTLMs` path to `_cleanupClient` that would have rebuilt the state.

A per-task timeout resolving to `null` would reuse the degraded-append path this PR already defines and keep the wedge from being permanent.

src/net/rest/SSEServer.ts
```diff
@@ -407,7 +449,30 @@ export class SSEServer {
...
+   * (duplicates) until its buffer TTL fires — and that TTL cleanup would drop
+   * the live client's store state.
+   */
+  releaseClient(clientId: string): void {
+    const client = this.clients.get(clientId);
+    if (!client) return;
+    if (client.expiryTimer) {
+      globalThis.clearTimeout(client.expiryTimer);
+    }
+    client.writer?.close().catch(SSEServer.noop);
```
RyanHavoc: `connect()` increments `staleCloses` at `:150` precisely so the framework's later close callback for a force-closed stream doesn't mark the replacement disconnected. This closes the writer and then deletes the `ClientState` at `:469`, destroying the counter that would have absorbed that close.

If the client reconnects to *this* instance before the framework reports the released stream's close, `disconnect()` finds fresh state with `staleCloses === 0` and treats the stale close as the new connection dying:

1. `connect('c1')` on A; the network dies silently, so A's writer stays live (half-open) and no close is reported.
2. The client reconnects onto B; B broadcasts its claim; A runs `releaseClient('c1')`.
3. The client's next reconnect load-balances back to A. `A.getConnectionIds()` -> `['c1']`.
4. The framework finally reports the released stream's close -> `A.disconnect('c1')`. `getConnectionIds()` -> `[]`.

The brand-new live stream is now marked disconnected: heartbeats stop (the loop writes only when `client.writer` is set), every notify task returns at the `!writer` guard after appending, and A's buffer TTL eventually runs `_cleanupClient` -> `dropClient` against a client that is live on A itself. Half-open teardown is both the case this JSDoc calls out and where the close callback lags most.

src/net/rest/SSEServer.ts
```diff
@@ -407,7 +449,30 @@ export class SSEServer {
...
+   * claimed by ANOTHER instance (a multi-instance routing signal — e.g. pup
+   * broadcasts a claim when a stream connects). Cancels the expiry timer and
+   * closes any lingering local writer WITHOUT touching the event store: the
+   * client is live elsewhere and its stored events and subscriptions must
+   * survive for that stream to use. Without this, the abandoned instance's
+   * retained ClientState keeps appending every relayed event to a shared store
+   * (duplicates) until its buffer TTL fires — and that TTL cleanup would drop
+   * the live client's store state.
+   */
+  releaseClient(clientId: string): void {
```
RyanHavoc, line 524: Deleting from `this.clients` stops *new* tasks, but tasks already on `client.pipeline` still hold the `client` reference and still run — including `eventStore.append(clientId, ...)` at `:389`. So the released ghost keeps appending into the shared store for a client that is live elsewhere, which is the duplicate this JSDoc says the method exists to prevent.

The window is bounded by queue depth, but the queue is non-empty precisely when the store is slow — the same condition that makes the claim broadcast worth having in the first place. A `released` flag on the state that queued tasks check around their append would close it. `_cleanupClient` and `destroy()` have the same shape.

src/net/rest/SSEServer.ts
```diff
@@ -118,69 +129,87 @@ export class SSEServer {
...
-        if (!knownClient && lastId > 0) {
-          // Client claims a lastEventId but we have no history — buffer expired or server restarted
-          this._writeEvent(client, { id: client.nextEventId++, event: 'resync', data: '{}', timestamp: Date.now() });
+      // Issued now so the store sees the cursor before any subsequent appends.
+      const replay = this.eventStore.replay(clientId, lastEventId).catch((): SSEReplayResult => ({ type: 'resync' }));
+      this._enqueue(client, async () => {
+        const result = await replay;
+        if (client.writer !== writer) return; // Replaced while awaiting
+        if (result.type === 'resync') {
+          this._writeFrame(writer, client.encoder, null, 'resync', '{}');
```
RyanHavoc: Base wrote this frame with `id: client.nextEventId++` ([`:167`](https://github.com/dabblewriter/patches/blob/b5dd98edb7f75869c8bd612d0c74b4e598c9aa27/src/net/rest/SSEServer.ts#L166-L168) and [`:176`](https://github.com/dabblewriter/patches/blob/b5dd98edb7f75869c8bd612d0c74b4e598c9aa27/src/net/rest/SSEServer.ts#L175-L177)), and that id was load-bearing twice: it re-anchored the browser's `Last-Event-ID` into the new id space, and it made `knownClient` true on the next connect. Per the EventSource spec a frame with no `id:` leaves the last-event-ID buffer unchanged, and `PatchesREST` uses the native `EventSource`, so the client can't clear the cursor itself — its `resync` handler only cycles state.

After a restart or deploy (in-memory store resets to `nextEventId = 1` while browsers still hold pre-restart ids) the cursor therefore stays in the old epoch. `replay()` never calls `_ensure()`, so a resynced client leaves no store entry at all; once any event has been appended-but-not-delivered, `known` flips true and `filter(e => e.id > 137)` empties the buffer. Replay then returns `{type:'events', events: []}` — a positively "verified" empty continuation — and nothing is written: no replay *and* no resync.

That's a behavioural difference from base rather than the stated exact parity. Either give the resync frame a store-assigned id so the cursor is re-anchored, or have `replay` treat "no entry for this client" as `resync` unconditionally.

src/net/rest/SSEServer.ts
```diff
@@ -309,14 +342,21 @@ export class SSEServer {
     if (!client) return;
     for (const docId of docIds) {
       client.subscriptions.delete(docId);
+      // A pending hydration snapshot may still contain this docId — record the
+      // removal so the apply can't resurrect it.
+      client.hydrating?.add(docId);
     }
+    this.eventStore.removeSubscriptions(clientId, docIds).catch(SSEServer.noop);
```
RyanHavoc: `unsubscribe()` records the docId in `client.hydrating` at `:347` so an in-flight snapshot can't resurrect it — but that Set is a field on one `ClientState`, created at `:178` and discarded at `:183`. It protects exactly one hydration on one instance and carries nothing across connects or across instances, so it misses the multi-instance race the PR exists to serve:

1. Client C on A, subscribed to {D1, D2}; the store records both.
2. The app unsubscribes D2. This line fires `removeSubscriptions` unawaited; pup answers `{ok:true}` immediately.
3. Redis is briefly slow — the removal hasn't landed.
4. C's stream drops and reconnects onto cold instance B. `loadSubscriptions` at `:179` reads the stale set and `:185` re-adds D2, into a routing set that is never re-authorized.

Related, same seam: `addSubscriptions` (`:288`, `:332`) and `removeSubscriptions` here are independent unawaited writes with no ordering guarantee between them, so a subscribe/unsubscribe pair can land out of order against an async store and leave the doc subscribed. The PR built `client.pipeline` specifically to guarantee that kind of ordering for appends and frames; the subscription mutations don't use it, and [the `SSEEventStore` JSDoc](https://github.com/dabblewriter/patches/blob/3d548a677b15a62eb6907b2b3fc3c1602372f42c/src/net/rest/SSEEventStore.ts#L47-L53) doesn't state any ordering or idempotence requirement for implementers to compensate.

src/net/rest/SSEServer.ts
```diff
@@ -368,13 +411,12 @@ export class SSEServer {
...
-    if (!client || !client.writer) return false;
+    if (!client?.writer) return false;
 
-    this._writeEvent(client, {
-      id: client.nextEventId++,
-      event,
-      data,
-      timestamp: Date.now(),
+    const writer = client.writer;
+    this._enqueue(client, async () => {
```
RyanHavoc: On base this wrote synchronously ([`:369-381`](https://github.com/dabblewriter/patches/blob/b5dd98edb7f75869c8bd612d0c74b4e598c9aa27/src/net/rest/SSEServer.ts#L368-L382)), so `true` genuinely meant written. Now the write is deferred behind any pending appends while `return true` at `:422` is immediate — if the stream drops before the queued task runs, it hits the writer-mismatch guard at `:418` and returns without writing. The JSDoc at `:409` still says "true if the client was connected and the event was written".

This matters more here than for `notify` because these frames are deliberately never stored or replayed: a dropped signal is gone permanently, and the caller was told it was delivered, so it won't retry or fall back. Confirmed on this head — `sendToClient` returned `true` and the stream received only `["retry: 5000\n\n"]`.

It also couples WebRTC signaling latency to store latency: a signal now queues behind a slow append for the same client, even though it never touches the store.

src/net/rest/SSEServer.ts
```diff
@@ -118,69 +129,87 @@ export class SSEServer {
...
+      // events fan out. Unsubscribes racing the load are collected in
+      // `hydrating` and excluded when the (stale) snapshot is applied.
+      client.hydrating = new Set();
+      const subscriptions = this.eventStore.loadSubscriptions(clientId).catch((): string[] => []);
+      this._enqueue(client, async () => {
+        const stored = await subscriptions;
+        const removed = client.hydrating ?? new Set<string>();
+        client.hydrating = null;
+        for (const docId of stored) {
+          if (!removed.has(docId)) client.subscriptions.add(docId);
```
RyanHavoc, line 216: No change requested — flagging for whoever reviews the pup side.

This restores the fan-out routing set without consulting `AuthorizationProvider`, whereas `subscribe()` at `:318` is the gate. Only authorized `subscribe()` calls ever write to the store, so it isn't an escalation, and "revocation doesn't stop a live feed" is already base behaviour. What's new is that it survives an instance move: previously a cold-instance reconnect started with an empty set and had to re-`subscribe` through the gate, which was an implicit re-authorization point. Now the routing set outlives the check for as long as the store's TTL — a Redis config rather than an auth decision.

pup's `GET /events/:clientId` is unauthenticated ("auth at subscribe time") and `clientId` is `createId(22)`, so this isn't reachable by guessing. Still worth being a deliberate decision rather than an implicit one, given it's the same surface as the `/events` hijack concern.

tests/net/rest/SSEServer.spec.ts
```diff
@@ -428,6 +458,7 @@ describe('SSEServer', () => {
       server.notify('doc1', 'changesCommitted', { docId: 'doc1', changes: [{ id: 'c1' }] });
       server.notify('doc1', 'changesCommitted', { docId: 'doc1', changes: [{ id: 'c2' }] });
       server.notify('doc1', 'changesCommitted', { docId: 'doc1', changes: [{ id: 'c3' }] });
+      await flushAsync();
```
RyanHavoc, line 464: `flushAsync()` landed here and at `:489`/`:494`/`:521`, but not in `should buffer events for disconnected clients` at [`:341`](https://github.com/dabblewriter/patches/blob/3d548a677b15a62eb6907b2b3fc3c1602372f42c/tests/net/rest/SSEServer.spec.ts#L340-L342) — and that one is the headline test for the in-memory parity claim.

Without it, that test no longer exercises replay at all. `notify()` now only enqueues ([`SSEServer.ts:380`](https://github.com/dabblewriter/patches/blob/3d548a677b15a62eb6907b2b3fc3c1602372f42c/src/net/rest/SSEServer.ts#L379-L381)), and its notify -> notify -> `connect('client1','0')` sequence has no await, so neither append has run when `connect()` calls `store.replay` at `SSEServer.ts:201`. Replay sees an empty store and returns `{type:'events', events: []}`; the asserted `id: 1` / `id: 2` frames come from the two notify tasks running afterwards and writing to the newly attached writer.

Verified by wrapping the store and logging the replay result while the upstream assertions still passed — it returned `{"type":"events","events":[]}`. The test would keep passing if `replay()` were deleted entirely. One `await flushAsync()` after the two notify calls restores it.

RyanHavoc (commented):
Reviewed the extraction against a genuinely async store, since that's the case the PR exists to enable — the in-memory default mutates synchronously, so most of the below is invisible to the added tests and only shows up once pup's Redis store lands.

The main one is `SSEServer.ts:201`: the replay read is issued outside the pipeline, so an append still in flight at reconnect can be committed to the store, dropped from the wire, and then omitted from the replay meant to cover it — permanently, once the cursor advances past it. I reproduced that one and the fix. The rest are lifecycle and contract issues around `releaseClient`, the pipeline, and the resync cursor.

Nothing here blocks the shape of the design, which reads well — the store seam and the `hydrating` guard are the right primitives.

src/net/rest/SSEServer.ts
jacwright: Fixed in 634540c — the read is now issued inside the enqueued task, and your repro is checked in as `should issue the replay read behind an append still in the pipeline`. One knock-on this exposed: with the read ordered behind queued appends, a notify task that runs after the reconnect would have delivered its event directly to the new stream AND via the replay covering it. The writer is therefore now captured at enqueue time — the mismatch guard drops the direct write and the replay delivers exactly once (the test's sentinel frame asserts no duplicate). The spec assertion at 660 moved after the stream read as you noted.

src/net/rest/SSEServer.ts
jacwright: Added `storeTimeoutMs` (default 10s): store calls inside the pipeline are wrapped in a guard that resolves to the degraded fallback on timeout or rejection — append → null (id-less frame), replay → resync, hydration → empty set — so a store call that never settles can't wedge the pipeline. pup also sets `commandTimeout: 500` on its dedicated events Redis client, so the server-side guard is the backstop rather than the first line.

src/net/rest/SSEServer.ts
jacwright: Fixed — discarding state (releaseClient or TTL cleanup) now transfers the closes still owed by the framework (the force-closed live writer plus any pending staleCloses) into a server-level `orphanedCloses` tally, which disconnect() consumes before consulting per-state counters. Test: `should not mark a later reconnection disconnected when the released stream's close arrives late`.

src/net/rest/SSEServer.ts
jacwright: Fixed with an identity check (`this.clients.get(clientId) !== client`) at the top of each queued notify task — covers releaseClient, TTL cleanup, and destroy() without a flag. Test: `should suppress appends already queued when the client is released`.

src/net/rest/SSEServer.ts
jacwright: Went with the re-anchor: the resync variant of `SSEReplayResult` now carries an optional store-assigned `id`, written on the resync frame. The in-memory store re-anchors on every resync and now also refuses cursors at/past its counter, with verification moved BEFORE the destructive filter — the 'verified empty continuation' can no longer destroy the buffer. pup's Redis store re-anchors by appending a `resync` marker through the normal append path (which also resets a poisoned log) and filters markers out of continuations; a side benefit is the >2h-idle case, where the expired sse:last marker would otherwise force a full sync on every subsequent blip instead of just the first.

src/net/rest/SSEServer.ts
jacwright: Store subscription mutations now ride a per-client serialized chain, and subscribe/addSubscriptions/unsubscribe resolve only once the mirror settles (bounded by storeTimeoutMs). The ordering/idempotence/visibility contract is documented on SSEEventStore.addSubscriptions. pup awaits before answering both subscription routes, and the forwarded-subscribe ack now waits on the owner's registration landing. One window remains open by choice: a forwarded *unsubscribe* to another owner stays fire-and-forget (documented at the call site) — the stale-set exposure there is bounded by the forward hop, not closed.

src/net/rest/SSEServer.ts
jacwright: Restored the synchronous write — no pipeline, so the boolean is truthful again and signaling latency is decoupled from store latency. The 'empty continuation' test that used sendToClient as a queue sentinel now flushes the pipeline before writing it.

src/net/rest/SSEServer.ts
jacwright: Kept, but made deliberate: documented at the hydration site in SSEServer (authorized-at-subscribe-time, TTL-bounded) and on pup's SSE_SUBS_TTL_SECONDS as the explicit bound on how long a stored subscription can outlive a revocation.

tests/net/rest/SSEServer.spec.ts
jacwright: Added the missing flushAsync, with a comment on why it's load-bearing — the test exercises replay() again.

RyanHavoc (approved):

PR #118, opened 2026-07-19
feat(client): max-time guard for IndexedDB opens and transactions

jacwright:
**TL;DR:** On some computers the browser's local database freezes mid-save and never responds — and today that failure is completely silent: the app looks fine while everything typed after the freeze is quietly lost. With this change, a frozen local database now surfaces as a visible, catchable error within seconds, so the app can warn the writer instead of silently eating their words. Nothing changes for healthy machines — normal saves finish in a few milliseconds and never come near the guard.

## What changed

Every IndexedDB `open()` and every transaction in the client store layer now runs under a max-time guard, enabled by default:

- **Soft threshold (default 1000ms):** invokes an overridable `onSlowStorage({ operation, storeNames, elapsedMs })` hook once (default: `console.warn`). No throw — purely advisory.
- **Hard threshold (default 4000ms):** aborts the transaction (`tx.abort()` where possible) and rejects with the new exported `StorageTimeoutError` carrying `{ operation, storeNames, elapsedMs }`.

Both thresholds and the hook are configurable via store options, plumbed through `IndexedDBStore`, `OTIndexedDBStore`, and `LWWIndexedDBStore`:

```ts
new OTIndexedDBStore('my-db', { slowStorageMs: 1000, storageTimeoutMs: 4000, onSlowStorage: info => ... });
```

## Design notes

- Timers arm when the request is issued and are cleared the moment any settle event fires (`complete`/`error`/`abort` for transactions; `success`/`error`/`blocked` for opens), so the guard only trips on a genuinely unsettled operation — it cannot fire on slow-but-completing work after the fact.
- On hard timeout the rejection is issued **before** `tx.abort()`, so the typed `StorageTimeoutError` wins over the `AbortError` the abort itself fires; an abort that throws (dead connection) can't mask the rejection.
- Individual requests (`get`/`put`/…) race against the owning transaction's failure. Without this, a hung request would park its caller forever even after the transaction-level guard fired — the store methods await request promises before `tx.complete()`.
- A `blocked` open stands down the hang guard: blocked is an observable state with its own existing 5s grace period and heal path, not a silent hang.
- A hung `open()` rejects current waiters and re-arms, so callers arriving later also fail fast — while a success that eventually lands still heals the store (same shape as the blocked-grace recovery).
- `StorageTimeoutError` lives beside `StorageError` in `net/error.ts` and is re-exported from the client entry. It deliberately does not classify as `isStorageError` — it is a new failure class for consumers to branch on.

## Tests

18 new tests in `tests/client/IndexedDBStore-timeout.spec.ts` (fake timers + controllable fake IDB objects, plus fake-indexeddb end-to-end): soft hook fires exactly once at threshold, hard abort rejects typed, settle-before-threshold fires nothing, hung `open()` covered (defaults, custom thresholds, re-arm, late-success heal, blocked interplay), hung individual request, no unhandled rejections. Full suite: 3203 passed, type-check/lint/format clean.

### DIFF
--- a/src/client/IndexedDBStore.ts
+++ b/src/client/IndexedDBStore.ts
@@ -9,7 +9,7 @@ import type {
   QuarantinedChange,
 } from '../types.js';
 import { deferred, type Deferred } from '../utils/deferred.js';
-import { toStorageError } from '../net/error.js';
+import { StorageTimeoutError, storageOpLabel, toStorageError } from '../net/error.js';
 import { signal } from 'easy-signal';
 import type { BranchClientStore } from './BranchClientStore.js';
 import type { PatchesStore, TrackedDoc } from './PatchesStore.js';
@@ -20,6 +20,97 @@ interface StoredBranch extends Branch {
   _docId: string;
   /** Numeric pending flag for IndexedDB indexing (1 when pending, absent otherwise) */
   _pending?: 1;
+}
+
+/** Soft threshold: an operation still unsettled after this fires `onSlowStorage`. */
+export const DEFAULT_SLOW_STORAGE_MS = 1000;
+/** Hard threshold: an operation still unsettled after this is rejected with {@link StorageTimeoutError}. */
+export const DEFAULT_STORAGE_TIMEOUT_MS = 4000;
+
+/** Passed to `onSlowStorage` when an IndexedDB operation crosses the soft threshold without settling. */
+export interface SlowStorageInfo {
+  operation: 'open' | 'transaction' | 'delete';
+  storeNames: string[];
+  elapsedMs: number;
+}
+
+/**
+ * Max-time guard options for IndexedDB opens, transactions, and deletes. Some machines have
+ * IDB calls that never settle (no success, error, or abort event), silently losing every
+ * write. A healthy operation settles in single-digit milliseconds, so these thresholds only
+ * trip on a genuinely unsettled one. Ships enabled; omitting options uses the defaults.
+ */
+export interface StorageGuardOptions {
+  /** Soft threshold in ms before `onSlowStorage` is invoked (no throw). Default {@link DEFAULT_SLOW_STORAGE_MS}. */
+  slowStorageMs?: number;
+  /** Hard threshold in ms before the operation is rejected with {@link StorageTimeoutError}. Default {@link DEFAULT_STORAGE_TIMEOUT_MS}. */
+  storageTimeoutMs?: number;
+  /** Invoked once per operation at the soft threshold. Defaults to a console.warn. */
+  onSlowStorage?: (info: SlowStorageInfo) => void;
+}
+
+function warnSlowStorage({ operation, storeNames, elapsedMs }: SlowStorageInfo): void {
+  console.warn(`${storageOpLabel(operation, storeNames)} still unsettled after ${elapsedMs}ms`);
+}
+
+interface StorageGuard {
+  /** Push the hard deadline out to a full `storageTimeoutMs` from now. No-op once settled. */
+  extend(): void;
+  /** Stand down the hard timer only (observable activity owns the deadline); the soft timer stays. */
+  clearHard(): void;
+  /** Settled: clear all timers; `extend()` becomes a no-op. */
+  clear(): void;
+}
+
+/**
+ * Arm the soft/hard guard timers for one storage operation. The soft timer fires
+ * `onSlowStorage` once; the hard timer invokes `onTimeout` with the elapsed ms.
+ * `storeNames` is called lazily, only when a timer actually fires.
+ */
+function armStorageGuard(
+  options: StorageGuardOptions | undefined,
+  operation: SlowStorageInfo['operation'],
+  storeNames: () => string[],
+  onTimeout: (elapsedMs: number) => void
+): StorageGuard {
+  const started = Date.now();
+  const timeoutMs = options?.storageTimeoutMs ?? DEFAULT_STORAGE_TIMEOUT_MS;
+  let done = false;
+  let slowTimer: ReturnType<typeof setTimeout> | undefined = setTimeout(() => {
+    slowTimer = undefined;
+    (options?.onSlowStorage ?? warnSlowStorage)({
+      operation,
+      storeNames: storeNames(),
+      elapsedMs: Date.now() - started,
+    });
+  }, options?.slowStorageMs ?? DEFAULT_SLOW_STORAGE_MS);
+  const fireHard = () => {
+    hardTimer = undefined;
+    onTimeout(Date.now() - started);
+  };
+  let hardTimer: ReturnType<typeof setTimeout> | undefined = setTimeout(fireHard, timeoutMs);
+  const clearHard = () => {
+    if (hardTimer !== undefined) {
+      clearTimeout(hardTimer);
+      hardTimer = undefined;
+    }
+  };
+  return {
+    extend() {
+      if (done) return;
+      clearHard();
+      hardTimer = setTimeout(fireHard, timeoutMs);
+    },
+    clearHard,
+    clear() {
+      done = true;
+      clearHard();
+      if (slowTimer !== undefined) {
+        clearTimeout(slowTimer);
+        slowTimer = undefined;
+      }
+    },
+  };
 }
 
 /**
@@ -79,6 +170,13 @@ export class IndexedDBStore implements PatchesStore, BranchClientStore {
   protected dbName?: string;
   protected dbPromise: Deferred<IDBDatabase>;
   protected external: boolean;
+  protected options: StorageGuardOptions;
+  /**
+   * Cancels the in-flight open's guard and blocked timers. Every reopen path (setName,
+   * versionchange, suspend recovery) and close() must call this so a superseded open's
+   * timers can never reject or replace the current `dbPromise`.
+   */
+  protected cancelOpenGuard?: () => void;
   protected requiredStores = new Set(['docs', 'snapshots', 'branches', 'quarantinedChanges']);
 
   /**
@@ -87,7 +185,8 @@ export class IndexedDBStore implements PatchesStore, BranchClientStore {
    */
   readonly onUpgrade = signal<(db: IDBDatabase, oldVersion: number, transaction: IDBTransaction) => void>();
 
-  constructor(dbOrName?: string | IDBDatabase | Promise<IDBDatabase>) {
+  constructor(dbOrName?: string | IDBDatabase | Promise<IDBDatabase>, options?: StorageGuardOptions) {
+    this.options = options ?? {};
     this.dbPromise = deferred<IDBDatabase>();
 
     if (dbOrName != null && typeof dbOrName !== 'string') {
@@ -165,6 +264,8 @@ export class IndexedDBStore implements PatchesStore, BranchClientStore {
   /** Opens the database. `version: null` opens at whatever version currently exists. */
   protected async initDB(version: number | null = IndexedDBStore.DB_VERSION) {
     if (!this.dbName) return;
+    // A superseded open's timers must never touch this open's promise.
+    this.cancelOpenGuard?.();
     const request = indexedDB.open(this.dbName, version ?? undefined);
     let blockedTimer: ReturnType<typeof setTimeout> | undefined;
     const clearBlockedTimer = () => {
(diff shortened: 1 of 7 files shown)


RyanHavoc (commented):
The problem this solves is real and the mechanics are careful — reject-before-abort ordering is correct, unhandled-rejection coverage is complete on all three paths (`this.promise.catch(reject)`, `this.failure.catch()`, `request.catch()`), and the test suite is thorough. I couldn't construct a case where a legitimately-settled request gets raced away, and the VersionError reopen, missing-stores reopen, `onversionchange` and `beginTransaction`'s retry are all safe because each runs after `clearGuardTimers()`.

My concern is the **hard 4000ms abort as a shipped default**. The timer measures wall-clock from transaction creation, which is not the same thing as IndexedDB being unresponsive — it also covers scope-queue wait and synchronous CPU. Comments below, roughly in severity order. The soft 1000ms hook seems fine to ship on by default; the hard threshold needs one of: re-base on last IDB activity rather than creation, raise it substantially, or make it opt-in — plus the `factories.ts` plumbing so a consumer can actually reach the knobs.

src/client/IndexedDBStore.ts
```diff
@@ -670,23 +750,75 @@ function stripInternal(stored: StoredBranch): Branch {
...
+        if (slowTimer !== undefined) clearTimeout(slowTimer);
+        // Reject BEFORE aborting so the typed timeout wins over the AbortError the abort
+        // fires (the later onabort reject is a no-op on the settled promise).
+        reject(new StorageTimeoutError('transaction', storeNames, Date.now() - started));
+        try {
+          tx.abort();
+        } catch {
+          // Connection already dead/finished — nothing left to abort.
+        }
+      }, options?.storageTimeoutMs ?? DEFAULT_STORAGE_TIMEOUT_MS);
```
RyanHavoc: The budget is wall-clock from transaction *creation* ([`started` at :764](https://github.com/dabblewriter/patches/blob/24c7f5862aeab7b410509e24e733068540790bb7/src/client/IndexedDBStore.ts#L763-L765)), but `db.transaction()` returns immediately while IndexedDB doesn't *run* a transaction until all overlapping earlier ones commit. So the 4s covers queue wait and synchronous CPU, not IDB liveness. Three ways healthy work gets aborted:

- **Queue wait.** `blockable` serialises per-docId, not per-store, so `Promise.all(docIds.map(id => saveDoc(id, ...)))` creates N transactions in one tick with N timers armed at t=0 while IDB runs them one at a time.
- **Same-database neighbours.** In DW3 the patches stores live in the same `dabble/{uid}` database as the export tooling, whose cursor walk deliberately holds a transaction active for its whole duration. Patches writes queued behind it burn the full budget with storage perfectly healthy — and those dumps get run on the largest, already-troubled accounts.
- **Suspension / long main-thread block.** Timers don't fire while frozen but the deadline passes in wall-clock; on resume the timer fires in the first task, ahead of the pending `oncomplete`. This file already documents Safari suspending backgrounded tabs aggressively ([:69-70](https://github.com/dabblewriter/patches/blob/24c7f5862aeab7b410509e24e733068540790bb7/src/client/IndexedDBStore.ts#L68-L71), DABBLE-WRITER-3-CA).

The last case has a nasty sub-mode: because `reject()` fires before `tx.abort()`, a transaction that *actually committed* reports `StorageTimeoutError` to the caller. Re-basing the timer on last IDB activity (reset on each request settle) rather than on creation would distinguish "hung" from "slow or queued".

src/client/LWWIndexedDBStore.ts
```diff
@@ -64,8 +64,8 @@ interface Snapshot {
 export class LWWIndexedDBStore implements LWWClientStore {
   public db: IndexedDBStore;
 
-  constructor(db?: string | IndexedDBStore) {
-    this.db = !db || typeof db === 'string' ? new IndexedDBStore(db) : db;
+  constructor(db?: string | IndexedDBStore, options?: IndexedDBStoreOptions) {
+    this.db = !db || typeof db === 'string' ? new IndexedDBStore(db, options) : db;
```
RyanHavoc, line 72: Worth checking this store's own bulk path against the 4s hard threshold before enabling it by default. [`applyServerChanges` :485-488](https://github.com/dabblewriter/patches/blob/24c7f5862aeab7b410509e24e733068540790bb7/src/client/LWWIndexedDBStore.ts#L484-L489) is two IDB round-trips per op, strictly sequential, inside one readwrite transaction:

```ts
for (const op of sortOpsByCommitOrder(serverChanges.flatMap(change => change.ops))) {
  await committedOps.delete([docId, op.path + '/'], [docId, op.path + '/\uffff']);
  await committedOps.put<CommittedOp>({ ...op, docId });
}
```

The input is unbounded — the caller fetches changes-since with no limit and no chunking — and the same transaction then does a `count` and often `compactSnapshot` (a `getAll` plus synchronous `applyPatch` plus N deletes). LWW is DW3's default algorithm. A user offline for a week reconnects with a few thousand queued ops; at even 1ms per round-trip, 2000 ops is 4s before compaction work.

The failure mode is worse than one aborted write: the abort rolls the transaction back, so `committedRev` never advances, and the next sync re-fetches the identical backlog and times out again. That's a permanent per-doc wedge loop, landing on exactly the cohort already prone to wedges.

src/client/OTIndexedDBStore.ts
```diff
@@ -38,8 +38,8 @@ interface StoredChange extends Change {
 export class OTIndexedDBStore implements OTClientStore {
   public db: IndexedDBStore;
 
-  constructor(db?: string | IndexedDBStore) {
-    this.db = !db || typeof db === 'string' ? new IndexedDBStore(db) : db;
+  constructor(db?: string | IndexedDBStore, options?: IndexedDBStoreOptions) {
+    this.db = !db || typeof db === 'string' ? new IndexedDBStore(db, options) : db;
```
RyanHavoc, line 46: The options are plumbed through the three store constructors but not through `factories.ts`, which is how consumers actually build these. [`createOTIndexedDBPatches` :69](https://github.com/dabblewriter/patches/blob/24c7f5862aeab7b410509e24e733068540790bb7/src/client/factories.ts#L68-L70), [`createLWWIndexedDBPatches` :103](https://github.com/dabblewriter/patches/blob/24c7f5862aeab7b410509e24e733068540790bb7/src/client/factories.ts#L102-L104) and [`createMultiAlgorithmIndexedDBPatches` :141-143](https://github.com/dabblewriter/patches/blob/24c7f5862aeab7b410509e24e733068540790bb7/src/client/factories.ts#L140-L144) all construct with `options.dbName` and no second argument, and none of the `*FactoryOptions` interfaces gained guard fields.

So DW3 is hard-wired to 4000ms with no kill switch and can never replace the `console.warn` soft hook — configuring it would mean abandoning the factories and hand-constructing all three stores. Given the guard ships enabled, the config surface should be reachable from the same entry point.

src/client/IndexedDBStore.ts
```diff
@@ -218,6 +297,7 @@ export class IndexedDBStore implements PatchesStore, BranchClientStore {
     };
     request.onsuccess = () => {
       clearBlockedTimer();
+      clearGuardTimers();
```
RyanHavoc: `onblocked`, `onerror` and `onsuccess` all clear the guard timers, but [`onupgradeneeded` :336-343](https://github.com/dabblewriter/patches/blob/24c7f5862aeab7b410509e24e733068540790bb7/src/client/IndexedDBStore.ts#L335-L344) does not — which contradicts the guard's own comment at :219-221 ("cleared the moment any event fires").

`onUpgrade` is a public signal, so consuming apps run migrations inside that version-change transaction. An upgrade that fires `onupgradeneeded` at 50ms and takes 5s to migrate rejects its waiters at t=4000 on a demonstrably progressing open, then re-arms and swaps `dbPromise` on every 4s tick until `onsuccess` lands. The same reasoning you applied to `blocked` ("an observable state, not a silent hang") applies here — `onupgradeneeded` firing is proof the open is alive.

src/client/IndexedDBStore.ts
```diff
@@ -174,6 +214,43 @@ export class IndexedDBStore implements PatchesStore, BranchClientStore {
...
+        if (this.closed) return;
+        // Same shape as the blocked-grace expiry: hand the current waiters a typed error and
+        // install a fresh deferred, so a success that eventually lands still heals the store.
+        // Re-arm so callers arriving after this rejection also fail fast instead of parking
+        // forever on the fresh deferred while the open stays hung.
+        const hung = this.dbPromise;
+        this.dbPromise = deferred<IDBDatabase>();
+        hung.reject(new StorageTimeoutError('open', [], Date.now() - openStarted));
+        hung.promise.catch(() => undefined);
+        armHangTimer();
```
RyanHavoc: The re-arm callback grabs `this.dbPromise` unconditionally — it never checks that the open request it belongs to is still the current one, nor that the deferred is still pending (`deferred()` exposes `status`).

[`setName()` :355-366](https://github.com/dabblewriter/patches/blob/24c7f5862aeab7b410509e24e733068540790bb7/src/client/IndexedDBStore.ts#L354-L367) is the path that exposes it. Called while an open is in flight, `this.db` is still `null`, so `dbPromise` is not replaced and `initDB()` starts a second request alongside the first:

1. Open R1 hangs. At t=4000 the guard rejects D1, installs D2, re-arms.
2. `setName('other')` starts R2 while R1 is still live.
3. R2 succeeds, clears *R2's* timers, sets `this.db`, resolves D2. Store is healthy.
4. At t=8000 *R1's* stale timer fires, swaps the resolved D2 for a fresh pending D3, and re-arms.

`this.db` now holds a live connection but `getDB()` never resolves again — a permanent 4s failure cycle that only `close()` stops. Pre-PR the stale in-flight open was inert; the re-arm turns it into active poison. An ownership check (bail if `this.dbPromise` is no longer the deferred this request owns) fixes it.

src/client/IndexedDBStore.ts
```diff
@@ -174,6 +214,43 @@ export class IndexedDBStore implements PatchesStore, BranchClientStore {
...
+      }, this.options.storageTimeoutMs ?? DEFAULT_STORAGE_TIMEOUT_MS);
```
RyanHavoc: There's no documented way to disable the guard, and the obvious workaround is destructive: `setTimeout` coerces any delay above 2^31-1 — including `Infinity` — to **1ms**. Verified in Node:

```
$ node -e "const t=Date.now();setTimeout(()=>console.log(Date.now()-t),Infinity)"
TimeoutOverflowWarning: Infinity does not fit into a 32-bit signed integer. Timeout duration was set to 1.
1
```

So a consumer reaching for `{ storageTimeoutMs: Infinity }` to opt out gets every transaction aborted at ~1ms and every open rejected — total write loss, the exact inverse of intent. Same for any value over ~24.8 days. Either clamp/validate the option, or accept a sentinel (`0`, `null`, `Infinity`) that skips arming the timer entirely.

Separately, worth a thought against [CLAUDE.md's scope boundary](https://github.com/dabblewriter/patches/blob/24c7f5862aeab7b410509e24e733068540790bb7/CLAUDE.md): a hard abort policy defaulting to on, with no off switch, is closer to app UX policy than to a generic primitive. The typed error and the soft hook are clearly in scope; the decision to destroy a slow-but-live transaction feels like the consuming app's call.

src/net/error.ts
```diff
@@ -184,6 +184,27 @@ export class StorageError extends Error {
...
 
+/**
+ * Error for an IndexedDB `open` or transaction that never settled — no success, error, or
+ * abort event within the hard threshold ({@link IndexedDBStore}'s `storageTimeoutMs`,
+ * default 4000ms). Some machines have IDB calls that simply hang forever, which would
+ * otherwise silently swallow every write with no error and no log. The guard converts that
+ * infinite hang into this fast, typed, catchable failure so the consuming app can surface
+ * it; a healthy operation settles in single-digit milliseconds and never comes near it.
+ */
+export class StorageTimeoutError extends Error {
```
RyanHavoc, line 197: Flagging the downstream consequence of the deliberate "does not classify as `isStorageError`" choice, since the TL;DR's goal is "the app can warn the writer": as of this PR it won't.

DW3's `isStorageDegradedError` matches on names `{StorageError, UnknownError, QuotaExceededError}`. `StorageTimeoutError` isn't one, so `isStorageDegraded` stays false, `setLocalPersistenceBlocked(true)` never fires, and the local-persistence-blocked banner — the one piece of UI built to say "your writes aren't reaching disk" — stays hidden for exactly that condition. It does reach Sentry at error level as a new fingerprint, so it's visible to us and invisible to the writer.

Not necessarily a change to this PR (the new class is defensible), but it needs a paired DW3 change to land alongside, otherwise the guard converts a silent hang into a silent typed rejection. Also worth knowing: DW3's `toCloneSafeSyncError` keeps only `name`/`message`/`stack`/`code`, so `operation`, `storeNames` and `elapsedMs` are stripped crossing worker to tab and survive only as message text.

src/client/IndexedDBStore.ts
```diff
@@ -192,6 +269,7 @@ export class IndexedDBStore implements PatchesStore, BranchClientStore {
     // resolves that new deferred and the store heals with no reopen.
     request.onblocked = () => {
       clearBlockedTimer();
+      clearGuardTimers();
```
RyanHavoc: Minor asymmetry: standing the hang guard down here is the right call, but the blocked path it defers to rejects **once** and installs a fresh deferred without re-arming, unlike the hang path which deliberately re-arms. So if a blocked open never settles, callers arriving after the 5s grace expiry park forever with the guard now disarmed — I advanced 10 minutes of timers with nothing settling.

Not a regression (it hung pre-PR too), but it's a hole in the new guard's coverage created by this `clearGuardTimers()`. Same shape at [`close()` :380](https://github.com/dabblewriter/patches/blob/24c7f5862aeab7b410509e24e733068540790bb7/src/client/IndexedDBStore.ts#L379-L381): during a hung open it early-returns on `!this.db`, so the fresh deferred the guard installed is never rejected and its waiters park indefinitely while the next re-arm tick bails on `this.closed`.

RyanHavoc (approved):

PR #119, opened 2026-07-19
DAB-773: send clientId on mutating REST requests for sender-excluded SSE fan-out

jacwright:
**TL;DR:** When you type in Dabble, the server currently sends your own edits back to you over the live-update stream, wasting bandwidth and creating opportunities for subtle sync glitches. This change makes the client identify itself on every save so the server can stop echoing a client's own changes back to it. (DAB-773)

## What changed

- `PatchesREST._fetch` appends a `clientId=<this.clientId>` query parameter to every **mutating (non-GET)** request. GETs — the bulk of launch traffic (doc reads, change fetches) — carry no `clientId`, so nothing about read traffic changes. A query parameter (unlike the custom header this started as) needs no CORS allow-list, so it adds no preflight and, crucially, can't be rejected by a server that doesn't know about it yet (see Rollout). `clientId` already appears in the path of sibling endpoints (`/events/:clientId`, `/subscriptions/:clientId`, `/signal/:clientId`), so this stays consistent with the existing API.
- `clientId` is now per-instance (`options.clientId ?? createId()`), **no longer persisted to sessionStorage**. Browsers copy sessionStorage on Duplicate Tab and restore it on reopen-closed-tab, so two live tabs could share one clientId — previously that just made them fight over the `/events` stream; with sender exclusion it would make one tab's commits permanently invisible to the other (excluded events are never buffered for replay). Reload continuity lost nothing real: a fresh page has no `Last-Event-ID` to replay from anyway.

The server-side seam already exists on main:

- `setAuthContext`/`getClientId` (`src/net/serverContext.ts`) let a consuming server bind the request's clientId to the commit synchronously
- `OTServer`/`LWWServer` capture `getClientId()` as their first synchronous statement and emit it as `originClientId` on `onChangesCommitted`/`onDocDeleted`
- `SSEServer.notify(..., exceptClientId)` skips the sender before writing **or buffering**

A consuming server (pup) reads the `clientId` query parameter and passes `originClientId` through to `notify` — that lands in pup#199.

## Rollout (order-independent)

No deploy-ordering constraint. A query parameter isn't part of the CORS preflight (which only vets method and request headers), so mutations preflight exactly as they do today (JSON Content-Type / DELETE method) and the currently-deployed pup simply ignores the unknown `clientId` param — no CORS failure, no behavior change. Sender exclusion switches on once pup#199 deploys and starts reading the param. Client and server can ship in either order.

This is the reason for the switch from a header: the earlier `X-Client-Id` approach had a hard constraint — shipping the client before pup added `X-Client-Id` to CORS `allowHeaders` would have hard-failed the preflight on every commit and delete. The query parameter removes that coupling entirely.

## Behavior

- The `clientId` param is present on all mutating requests (explicit `options.clientId` or a per-instance generated one). Today's per-tab ids get sender-exclusion as soon as pup#199 is live; the sync-v2 stable per-uid clientId gets it the same way. Callers passing `options.clientId` must keep it unique per live connection.
- Servers that ignore the param see exactly today's behavior. Exclusion keys on exact clientId equality only — the same user's other tabs/devices with different clientIds still receive every event.

## Tests

- `clientId` query param sent on commit and delete with the configured id; generated clientId sent when none is configured
- GETs carry no `clientId` param (stay CORS-simple)
- existing per-endpoint URL assertions updated for the appended param
- two instances never share a clientId
- `SSEServer.notify` exclusion also prevents buffering the echo for a disconnected sender

## Note for #117

Touches `docs/sse-rest.md` (one paragraph) and `tests/net/rest/SSEServer.spec.ts` (one added test), both of which #117 (SSEEventStore) rewrites. Both touches are minimal; whichever merges second has a trivial rebase.

### DIFF
--- a/docs/sse-rest.md
+++ b/docs/sse-rest.md
@@ -56,7 +56,7 @@ doc.change(draft => {
 
 ```typescript
 interface PatchesRESTOptions {
-  // Explicit client ID. If not provided, restored from sessionStorage or generated.
+  // Explicit client ID. If not provided, a random id is generated per instance.
   clientId?: string;
 
   // Static headers for every request (e.g. Authorization)
@@ -67,20 +67,13 @@ interface PatchesRESTOptions {
 }
 ```
 
-### Client ID Persistence
+### Client ID
 
-By default, `PatchesREST` persists the client ID in `sessionStorage`. This means:
+By default, `PatchesREST` mints a random client ID per instance. It is deliberately never persisted: browsers copy `sessionStorage` on Duplicate Tab and restore it on reopen-closed-tab, and two live clients sharing an id would fight over the `/events` stream and, with sender exclusion, miss each other's changes.
 
-- **Page refresh** → same client ID → seamless buffer replay, no full re-sync
-- **New tab** → new client ID (each tab is a separate client)
-- **Tab close** → sessionStorage cleared
-- **Web Worker** → no sessionStorage, falls back to random UUID per construction
-
-To control this yourself, pass `clientId` in the options:
+To control the id yourself, pass `clientId` in the options. It must be unique per live connection:
 
 ```typescript
-const clientId = localStorage.getItem('my-client-id') ?? crypto.randomUUID();
-localStorage.setItem('my-client-id', clientId);
 const rest = new PatchesREST(url, { clientId });
 ```
 
@@ -102,6 +95,8 @@ Both implement the `PatchesConnection` interface, so everything downstream (`Pat
 ## REST API
 
 All document operations use standard HTTP methods. Document IDs can contain slashes (they're hierarchical). Sub-resources use `_`-prefixed path segments to avoid collisions.
+
+Mutating (non-GET) requests carry the client ID as a `clientId` query parameter so the server can bind mutations to their origin (via `setAuthContext`) and exclude the sender from its own `changesCommitted` fan-out. A query parameter (unlike a custom header) needs no CORS allow-list and adds no preflight requirement of its own, and a server that predates the feature simply ignores it. GETs omit it, and `POST /signal/:clientId` already carries the id in its path.
 
 ### SSE & Subscriptions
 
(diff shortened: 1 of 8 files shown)


RyanHavoc:
Reviewed and pushed two follow-up commits:

- 63a860f `fix(rest)`: the mutation query param was percent-encoded but the `/events/`, `/subscriptions/`, and `/signal/` paths interpolated the raw clientId. A caller-supplied id containing `#` or `?` would register the SSE stream under a truncated id while mutations advertised the full one — sender exclusion then misses by exact match and the client is silently echoed its own changes (unreachable with the generated `createId(22)`, reachable via `options.clientId`, which the JSDoc only constrains to "unique"). The id is now encoded once per instance and used for every carrier, with a test locking in path/query symmetry.

- 3824146 `feat(net)`: the new docs paragraph promises origin binding "via setAuthContext", but the doc's own commit-route example never read the query param or set the context — copying it gives `getClientId() === undefined`, `notify()` excludes nobody, and the echo this PR exists to prevent comes back silently. `setAuthContext`/`clearAuthContext` were also unexported (internal), so REST consumers had no supported way to feed OTServer's origin capture. Both are now exported from `@dabble/patches/net` and the example shows the synchronous set / try-finally clear pattern.

One reviewed-and-accepted cost, no change made: with per-instance ids, each reload strands the old id's SSEServer record until `bufferTTLMs` (default 5 min), and disconnected buffers are exempt from window-trimming until then. Bounded and self-healing, and clearly better than the sessionStorage sharing hazard; a `pagehide` disconnect beacon could shrink it later if it ever matters.

Also verified during review, for the record: the scarier candidate — reload plus a lost commit ack re-sending under a new clientId and defeating dedup — does not happen; the store's write-time `[docId, change.id]` uniqueness guard is origin-agnostic, and the read-side `sameOrigin` filter was never the guard for that case.

PR #120, opened 2026-07-20
In-transaction rev minting, trust-the-open-doc reads, and onServerCommit (DAB-783)

jacwright:
## TL;DR

Apps that use patches get three improvements with no API changes to worry about:

- **Multi-tab editing no longer loses a keystroke batch to a rare same-instant write collision.** When the same document is open in two tabs, they share one local database. Previously two tabs writing at the exact same moment could stamp the same slot and one tab's batch would silently vanish. Now the local store assigns each change its slot inside the write transaction itself, so concurrent tabs can never collide.
- **Sync does far less database work per flush.** Receiving server changes used to fully rebuild the document from disk several times per flush cycle. It now trusts the already-open document as the source of truth and touches disk only to persist, so steady-state sync is meaningfully lighter.
- **Apps get a reliable "your change is committed" signal.** `patches.onServerCommit` now fires exactly once per committed change, at the moment it becomes durable on disk, with re-deliveries filtered out. Word counts, journal drains, and any per-change bookkeeping can rely on it.

---

## Technical detail

Three contract changes, each replacing an invariant that used to be enforced in application-instance memory with one the shared store enforces structurally.

### 1. In-transaction rev minting (replaces per-instance rev assignment)

`savePendingChanges` now assigns each change's `rev` **inside the persist transaction**, re-stamping the incoming objects from the tail already on disk (`max(committedRev, existing pending rev) + 1, +2, ...`) rather than trusting whatever the caller computed. IndexedDB serializes same-store readwrite transactions, so two contexts racing a mint over one database each read the other's row and number past it instead of clobbering at the `[docId, rev]` key. `OTInMemoryStore` mirrors this. `rev` is only the pending-queue ordering key; `baseRev` and `id` are untouched.

Previously the minting instance read `committedRev`/pending tail and assigned rev itself. That was correct only under a per-instance lock, which does not exist across tabs -- the collision this closes.

### 2. Conflict-checked pending replace (closes the foreign-mint wipe race)

`applyServerChanges` and `quarantinePendingChange` take an optional `pendingTailRev`: the max pending rev the caller's rebase/ejection input covered. When supplied, the store refuses the delete-and-replace and returns `'conflict'` if the current queue holds any change with `rev > pendingTailRev` (a foreign tab that minted after the caller's read). The caller (`applyServerChanges` / `replacePendingChanges` / `reconcilePending` / `ejectPendingChange` in `OTAlgorithm`) re-reads the now-larger queue and rebases again, bounded by `APPLY_CONFLICT_RETRIES = 10`; each pass anchors to a strictly larger tail so it converges in one or two, and exhausting the bound throws rather than looping. Omitting `pendingTailRev` keeps the unconditional replace and the `void` return (back-compat).

This closes the race where a receive-side rebase in one tab wipes a fresh pending mint from another that landed between the rebase read and the write.

### 3. Trust the open doc on receive (removes per-flush materialization)

`applyServerChanges` and `getPendingToSend` now take the open `PatchesDoc` and treat it as the live source of truth. Gap detection is pure rev arithmetic (a `MissingChangesError` shaped like `applyCommittedChanges`' so `PatchesSync` still routes gaps to `syncDoc` recovery) -- **no state materialization**. The aligned path is store-read-free; the store is re-read only on the rare misaligned path (root-replace catchup, stale re-delivery) or to pick up a foreign tab's mint delta via a ranged `getPendingChanges({ startAfterRev })`. This removes ~5 full doc materializations per flush cycle and all state materialization for closed docs.

### Deleted

- **Strand apparatus** (`_recentCommittedIds`, `_noteCommittedIds`, `_recentCommittedChanges`, `_dropCommittedStragglers`, `RECENT_COMMITTED_ID_WINDOW`): the per-instance record of committed ids that detected an already-committed pending copy re-sending as a duplicate. In-txn mint plus conflict-checked replace make the stranding it guarded against unreachable; the server's id dedup remains the eventual backstop.
- **`isRetry` pre-scan** (`_findChangeById` on the retry path, the `isRetry` param through `ClientAlgorithm`/`LWWAlgorithm`/`Patches`): the idempotent-retry lookup that returned an already-accepted change instead of re-minting. A rejected IndexedDB transaction is atomic, so a failed `savePendingChanges` persisted nothing and a re-submit cannot duplicate; the stable change id still backstops the server's dedup for the ambiguous landed-but-timed-out case. `_collectPending` replaces `_findChangeById`.
- **Per-doc lock at all former call sites**: `_withDocLock` is now kept at exactly one seam -- mint (`handleDocChange`) vs receive (`applyServerChanges`) on this instance -- the same-instance hole the rev-only store contracts cannot express (a mint reading `committedRev` while a concurrent receive advances it). Every other former holder (`replacePendingChanges`, `reconcilePending`, `dropResolvedPending`, `verifyPendingChange`, `ejectPendingChange`, `deleteDoc`, `confirmDeleteDoc`) is unlocked and relies on the conflict-checked replace instead.

### onServerCommit

`PatchesSync._applyServerChanges` captures the durable tip before applying and emits `onServerCommit` only for `rev > priorRev` -- the newly-durable changes. DAB-773 echoes every flushed commit back to its sender and catchup can overlap a broadcast, so the raw batch would re-fire an already-durable rev range and a per-change consumer would double-count. This is the single choke point covering every path where server-committed changes become locally durable (flush echoes, broadcasts, catchup, merges).

### Consumer notes

- **dw3**, on the release that includes this, deletes its `inTxnRevMint` wrapper, its `otDocReadCache`, and its absorption subsystem -- all of which existed to compensate for the contracts this PR moves into patches.
- **Server / pup surface untouched**: this is entirely client-store and client-algorithm; the wire protocol, commit RPC, and server dedup are unchanged.

### Tests

Full suite: 118 files, 3295 passed, 4 skipped. New coverage:

- `tests/client/OTStore-crossContext.spec.ts` (new) -- two store instances over one fake-indexeddb database, no shared lock:
  - concurrent mints get distinct, strictly increasing revs (R1);
  - `savePendingChanges` re-stamps the passed objects in place so callers carry the persisted rev;
  - `applyServerChanges` retries and folds in a foreign mint that lands between the rebase read and the replace (R2), and applies in a single attempt when nothing races;
  - `applyServerChanges` / `ejectPendingChange` do not wipe a foreign mint whose rev collides a doc-only rev (store-tail vs doc-merged-max).
- `tests/net/PatchesSync.spec.ts` -- `onServerCommit` fires once on durable apply (broadcast and flush-ack paths), never for an empty batch, never re-fires a re-delivered rev, and emits only the newly-durable revs in a mixed batch.
- Reworked `OTAlgorithm` suites (`committedStrands`, `getPendingToSend`, `reconcilePending`, `stableId`) to the trust-the-open-doc and conflict-replace contracts.

Closes DAB-783.

### DIFF
--- a/docs/operational-transformation.md
+++ b/docs/operational-transformation.md
@@ -140,32 +140,27 @@ The server also broadcasts Alice's change to all other clients (including Bob).
 
 ### 5. Clients Apply Updates
 
-This is where the `applyCommittedChanges` algorithm handles the complexity:
+When server changes arrive, `PatchesSync` hands them to the store's `applyServerChanges`, passing the open doc if there is one:
 
 ```typescript
 // In PatchesSync when server changes arrive
-const currentSnapshot = await store.getDoc(docId);
-const newSnapshot = applyCommittedChanges(currentSnapshot, serverChanges);
-
-// Update storage and notify open documents
-await store.applyServerChanges(docId, serverChanges);
-
 const doc = patches.getOpenDoc(docId);
-if (doc) {
-  doc.applyCommittedChanges(serverChanges, newSnapshot.changes);
-}
+const rebasedPending = await store.applyServerChanges(docId, serverChanges, doc);
+patches.onServerCommit.emit(docId, serverChanges);
 ```
+
+There is no separate `getDoc` reload. When a doc is open the algorithm **trusts it** as the live source of truth: it detects gaps by rev arithmetic (no state materialization), applies the server changes, and rebases the doc's pending changes against them in the same transaction. The store is a durability log and a cross-tab mailbox, not a snapshot to reload on every frame.
+
+Persisting the rebased pending is **conflict-checked**. The write only lands if the pending tail the store held still matches what the rebase started from. A foreign tab that minted in the meantime wins that slot, the store returns `'conflict'`, and the algorithm re-reads and rebases again (bounded retry, each pass anchored to a strictly larger tail so it converges). `onServerCommit` fires at this durable choke point: once it emits, the changes are on disk.
 
 Alice's `PatchesSync`:
 
-- Uses `applyCommittedChanges` to merge Bob's change with her pending changes
-- The algorithm calls `rebaseChanges` internally to transform her pending changes
-- Updates her `PatchesDoc` with the new state
+- Rebases Bob's change against her pending changes via `rebaseChanges`
+- Updates her open `PatchesDoc` in place
 
 Bob's `PatchesSync`:
 
-- Receives Alice's change and applies it normally
-- If he has pending changes, they get rebased automatically
+- Receives Alice's change; if he has pending changes, they get rebased automatically
 
 Both Alice and Bob see the same document, with both of their changes applied in the correct order.
 
@@ -332,7 +327,7 @@ interface BranchingStoreBackend {
 }
 ```
 
-Patches provides `InMemoryStore` for testing and simple use cases, and `IndexedDBStore` for browser-based client persistence. See [persist.md](persist.md) for details.
+Patches provides `InMemoryStore` for testing and simple use cases, and `IndexedDBStore` for browser-based client persistence. Both mint each pending change's `rev` **inside the persist transaction**, re-stamping it from the tail already on disk rather than from whatever the caller passed. This is the default: it keeps a doc open in two tabs from stamping the same rev twice, and is what makes the conflict-checked replace above safe. See [persist.md](persist.md) for details.
 
 ## How Transformation Works
 
--- a/src/client/OTAlgorithm.ts
+++ b/src/client/OTAlgorithm.ts
@@ -1,4 +1,4 @@
-import { applyCommittedChanges } from '../algorithms/ot/client/applyCommittedChanges.js';
+import { MissingChangesError } from '../algorithms/ot/client/applyCommittedChanges.js';
 import { applyChanges } from '../algorithms/ot/shared/applyChanges.js';
 import { breakChanges } from '../algorithms/ot/shared/changeBatching.js';
 import { computePendingEjection } from '../algorithms/ot/shared/ejectPendingChange.js';
@@ -14,13 +14,12 @@ import type { PatchesDoc, PatchesDocOptions } from './PatchesDoc.js';
 import type { TrackedDoc } from './PatchesStore.js';
 
 /**
- * How far back in the committed tail to scan for a change id — on an idempotent retry (in
- * case the original submit committed between attempts) and when checking a suspect pending
- * change for an already-committed copy (DAB-607). Also bounds the per-doc in-memory record
- * of committed ids. Bounded so the scans stay cheap; a just-committed change sits at the
- * head, so a small window catches the realistic cases.
+ * Bound on the conflict-safe replace retries (applyServerChanges / replacePendingChanges /
+ * reconcilePending / ejectPendingChange). Each retry reads a strictly larger pending tail and
+ * mints are human-paced, so the loop converges in one or two passes; the bound only guards
+ * against a pathological store, and exceeding it throws rather than looping forever.
  */
-const RECENT_COMMITTED_ID_WINDOW = 200;
+const APPLY_CONFLICT_RETRIES = 10;
 
 /**
  * OT (Operational Transformation) algorithm implementation.
@@ -28,6 +27,11 @@ const RECENT_COMMITTED_ID_WINDOW = 200;
  * OT uses revision-based history and rebasing for concurrent edits.
  * This algorithm owns an OT-compatible store and handles all OT-specific
  * logic.
+ *
+ * Cross-context safety lives in the store, not in this class: in-transaction rev assignment
+ * (savePendingChanges) is the sole sequencer, and conflict-safe replace (applyServerChanges with
+ * a pendingTailRev) keeps a foreign tab's mint from being wiped by a rebase. Any tab may mint;
+ * the receive-side mutations here run only in the elected writer.
  */
 export class OTAlgorithm implements ClientAlgorithm {
   readonly name = 'ot';
@@ -35,15 +39,14 @@ export class OTAlgorithm implements ClientAlgorithm {
 
   protected readonly _options: PatchesDocOptions;
 
-  /** Per-doc FIFO mutex (see {@link _withDocLock}). */
-  private readonly _docLocks = new Map<string, Promise<unknown>>();
-
   /**
-   * Per-doc record of recently committed change ids applied by this instance (see
-   * {@link _noteCommittedIds}). Used to recognize a stranded pending copy of an
-   * already-committed change regardless of how rebases have re-stamped it.
+   * Minimal per-doc mutex, kept at exactly the mint-vs-receive seam (see {@link _withDocLock}).
+   * Cross-context safety is the store's job (R1 in-txn rev mint, R2 conflict-safe replace); this
+   * lock only closes the same-instance hole those rev-only contracts cannot express — a mint
+   * reading committedRev while a concurrent receive advances it (stale baseRev) or persisting ops
+   * the receive rebased away in place.
    */
-  private readonly _recentCommittedIds = new Map<string, Set<string>>();
+  private readonly _docLocks = new Map<string, Promise<unknown>>();
 
   constructor(store: OTClientStore, options: PatchesDocOptions = {}) {
     this.store = store;
@@ -68,53 +71,37 @@ export class OTAlgorithm implements ClientAlgorithm {
     ops: JSONPatchOp[],
     doc: PatchesDoc<T> | undefined,
     metadata: Record<string, any>,
-    id?: string,
-    isRetry?: boolean
+    id?: string
   ): Promise<Change[]> {
     if (ops.length === 0) return [];
 
-    // Serialize per-doc so a concurrent receive-rebase and this mint can't read the
-    // same rev and clobber each other at the [docId, rev] store key (silent change loss).
     return this._withDocLock(docId, async () => {
-      // Re-check under the lock: ops arrays are shared with the doc's optimistic queue,
-      // and a receive-rebase that ran while we waited may have rebased them away.
+      // Re-check under the lock: ops arrays are shared with the doc's optimistic queue, and a
+      // receive-rebase that ran while we waited may have rebased them away.
       if (ops.length === 0) return [];
 
-      // Idempotent retry: if this exact caller-minted change id was already accepted on a
-      // prior attempt (the submit RPC timed out *after* the hub had persisted it), return the
-      // existing change instead of minting+persisting+applying a duplicate. Only runs on a
-      // retry, so the first submit keeps its fast path.
-      if (isRetry && id) {
-        const existing = await this._findChangeById(docId, doc, id);
-        if (existing.length > 0) return existing;
-      }
-
-      // Get revision info from doc if available, otherwise from store
+      // Revision info from the open doc; else from the store (no state materialization).
+      // Provisional only — savePendingChanges re-stamps rev in its own transaction from the
+      // persisted tail, the sole cross-context sequencer.
       let committedRev: number;
       let pendingRev: number;
-
       if (doc) {
         const otDoc = doc as OTDoc<T>;
         const pendingChanges = otDoc.getPendingChanges();
         committedRev = otDoc.committedRev;
         pendingRev = pendingChanges[pendingChanges.length - 1]?.rev ?? committedRev;
       } else {
-        // Worker scenario: get from store
-        const snapshot = await this.store.getDoc(docId);
-        committedRev = snapshot?.rev ?? 0;
-        const pendingChanges = snapshot?.changes ?? [];
-        pendingRev = pendingChanges[pendingChanges.length - 1]?.rev ?? committedRev;
+        committedRev = await this.store.getCommittedRev(docId);
+        const pending = await this.store.getPendingChanges(docId);
+        pendingRev = pending[pending.length - 1]?.rev ?? committedRev;
       }
 
-      // Create changes from ops
       const changes = this._createChangesFromOps(committedRev, pendingRev, ops, metadata, id);
-
       if (changes.length === 0) return [];
 
-      // Save to store
+      // Re-stamps each change's rev in place from the persisted tail; the objects below carry it.
       await this.store.savePendingChanges(docId, changes);
 
-      // Apply changes to doc if provided (uncommitted changes have committedAt === 0)
       if (doc) {
         (doc as OTDoc<T>).applyChanges(changes);
       }
@@ -128,15 +115,26 @@ export class OTAlgorithm implements ClientAlgorithm {
     return pending.length > 0;
   }
 
-  async getPendingToSend(docId: string): Promise<Change[] | null> {
-    // getDoc reads pending and the rev it's based on in one store transaction, so a concurrent
-    // receive-rebase can't advance committedRev between the two reads. Heal only the outgoing
-    // batch; the next server echo rebases the stored copy into agreement.
-    const snapshot = await this.store.getDoc(docId);
-    if (!snapshot || snapshot.changes.length === 0) return null;
-    const pending = await this._dropCommittedStragglers(docId, snapshot.changes, snapshot.rev);
-    if (pending.length === 0) return null;
-    return this._withConsistentBaseRev(docId, pending, snapshot.rev);
+  async getPendingToSend(docId: string, doc?: PatchesDoc<any>): Promise<Change[] | null> {
+    let batch: Change[];
+    if (doc) {
+      // Trust the open doc for pending; a ranged store read picks up any foreign tab's mints
+      // (rev past the doc's tail) and appends them raw. The serial flush and the commit echo
+      // rebase make that safe — the doc learns of them through the echo.
+      const otDoc = doc as OTDoc<any>;
+      const pending = otDoc.getPendingChanges();
+      const tail = pending[pending.length - 1]?.rev ?? otDoc.committedRev;
+      const delta = await this.store.getPendingChanges(docId, { startAfterRev: tail });
+      // Filter the delta against the doc's own pending by id: a custom store that ignores
+      // startAfterRev (back-compat) would return the whole queue, folding the doc's pending in
+      // twice — a duplicate commit the server's id dedup would otherwise have to catch.
+      const have = new Set(pending.map(c => c.id));
+      batch = [...pending, ...delta.filter(c => !have.has(c.id))];
+    } else {
+      batch = await this.store.getPendingChanges(docId);
+    }
+    if (batch.length === 0) return null;
+    return this._withConsistentBaseRev(docId, batch);
   }
 
   async applyServerChanges<T extends object>(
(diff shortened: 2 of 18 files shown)


src/client/OTAlgorithm.ts
```diff
@@ -128,15 +115,26 @@ export class OTAlgorithm implements ClientAlgorithm {
...
+  async getPendingToSend(docId: string, doc?: PatchesDoc<any>): Promise<Change[] | null> {
+    let batch: Change[];
+    if (doc) {
+      // Trust the open doc for pending; a ranged store read picks up any foreign tab's mints
+      // (rev past the doc's tail) and appends them raw. The serial flush and the commit echo
+      // rebase make that safe — the doc learns of them through the echo.
+      const otDoc = doc as OTDoc<any>;
+      const pending = otDoc.getPendingChanges();
+      const tail = pending[pending.length - 1]?.rev ?? otDoc.committedRev;
+      const delta = await this.store.getPendingChanges(docId, { startAfterRev: tail });
```
RyanHavoc, line 127: The ranged read anchors on the open doc's tail, but R1 mints from the shared store tail — when a foreign tab mints first, its row lands below this doc's tail and is structurally invisible to the strict `>` here. The comment above says "rev past the doc's tail", which only holds when this tab minted last.

Repro against a real `OTServer` (two `OTIndexedDBStore`s over one DB, doc `{items:['A','B','C']}`): tab B appends X (mints rev 1), tab A prepends Y (mints rev 2). A's flush sends only rev 2; the echo splices Y out and relabels B's row to the new baseRev without transforming it (https://github.com/dabblewriter/patches/blob/f9af8e7c38591f8226a4087e4d556f82c861f492/src/algorithms/ot/shared/rebaseChanges.ts#L43-L49), and the next flush ships X at the stale index — final server state `["Y","A","B","X","C"]` instead of `["Y","A","B","C","X"]`. On main the same interleaving lost X outright to the rev collision, so this is not a strict regression — but main's unranged read did send the foreign row, so the doc-tail range is a new loss channel, and the docs now advertise the two-tab open-doc case as safe.

Sharper variant: after tab B's own `applyServerChanges` correctly transforms its store row in place (same rev), A's flush still ships the stale ops — the correction never raises the rev past the strict `>`.

Suggest unioning the full store queue by id (store rows winning) instead of range-reading past the doc tail, or stating explicitly that concurrent cross-context mints converge on rev order only, not intent.

https://github.com/dabblewriter/patches/blob/f9af8e7c38591f8226a4087e4d556f82c861f492/src/client/OTAlgorithm.ts#L120-L132

RyanHavoc (commented):
Reviewed the three contract changes with runnable repros against this head (f9af8e7) and main. Nine inline comments: three regressions confirmed with pass-on-main / fail-on-head differentials (cross-tab transform gap in `getPendingToSend`, the R2 watermark's blindness to rev-neutral receives, and the `deleteDoc` unlock), two medium items on the `isRetry` deletion and the closed-doc receive path, and four smaller ones on `onServerCommit` coverage and docs. What held up under scrutiny: the in-txn mint genuinely stays inside one IDB transaction, the retry anchors strictly increase, the gap arithmetic is faithful (checked off-by-one, backward revs, root-replace catchup), and `dropResolvedPending`/`verifyPendingChange` are safe unlocked.

src/client/OTAlgorithm.ts
```diff
@@ -219,64 +236,54 @@ export class OTAlgorithm implements ClientAlgorithm {
...
+      const committedRev = await this.store.getCommittedRev(docId);
       const current = await this.store.getPendingChanges(docId);
-      let rev = newChanges.length > 0 ? newChanges[newChanges.length - 1].rev : await this.store.getCommittedRev(docId);
+      const tailRev = current.length > 0 ? current[current.length - 1].rev : committedRev;
+      let rev = newChanges.length > 0 ? newChanges[newChanges.length - 1].rev : committedRev;
       const mintedSince = current.filter(c => !oldIds.has(c.id)).map(c => ({ ...c, rev: ++rev }));
-      // applyServerChanges with no server changes atomically replaces the pending queue
-      await this.store.applyServerChanges(docId, [], [...newChanges, ...mintedSince]);
-    });
+      const result = await this.store.applyServerChanges(docId, [], [...newChanges, ...mintedSince], tailRev);
```
RyanHavoc, line 250: R2's predicate (`rev > pendingTailRev` in both stores) only fires when the pending tail rises. A concurrent receive is invisible to it when it is rev-neutral: `rebaseChanges` renumbers survivors from `lastServerRev + 1` (https://github.com/dabblewriter/patches/blob/f9af8e7c38591f8226a4087e4d556f82c861f492/src/algorithms/ot/shared/rebaseChanges.ts#L60-L69), so `tail_new − tail_old = #serverChanges − #droppedPending` — and the client's own commit echo (one in, one out) leaves the tail exactly where it was. With `_withDocLock` gone from this method, nothing serializes it against `_receiveCommittedChanges` (`@blockable` is inert client-side; nothing ever calls `blocking`).

Repro without touching product code (the test only controls when the deferred store write lands): store at committedRev 4, pending `[S@5 sent-unacked, C@6]`; `flushDoc` splits C; the echo committing S lands between this method's reads and its write. The replace returns success and reverts the receive — S is resurrected into pending at rev 5 == committedRev and the splits are left at baseRev 4, a frame stale. On main the lock forces the correct order.

The eject path has the same hole with a wider window (it can span the user-consent prompt via `Patches.ejectPendingChange`), which contradicts the comment at L379-381 claiming the re-corroboration is atomic with the ejection.

The lock and R2 solve different problems — same-instance interleaving vs cross-tab mints — and this PR treats them as substitutes. Folding `committedRev` (or a pending id-set) into the watermark, or keeping the lock on the replace-style callers, would close it.

https://github.com/dabblewriter/patches/blob/f9af8e7c38591f8226a4087e4d556f82c861f492/src/client/OTAlgorithm.ts#L244-L251

src/client/OTAlgorithm.ts
```diff
@@ -444,34 +454,26 @@ export class OTAlgorithm implements ClientAlgorithm {
   }
 
   async deleteDoc(docId: string): Promise<void> {
-    this._recentCommittedIds.delete(docId);
-    // Under the lock too: a delete must not race an in-flight mint/apply for the doc.
-    return this._withDocLock(docId, () => this.store.deleteDoc(docId));
+    return this.store.deleteDoc(docId);
```
RyanHavoc, line 457: The justification at L471-478 ("every other former holder relies on the R2 contract") does not hold for `deleteDoc`: `store.deleteDoc` takes no `pendingTailRev` and can never return `'conflict'` — it is outside R2 entirely. The lock here was added deliberately by 061b853 ("a delete must not race an in-flight mint/apply").

With it gone, a mint whose store write is still in flight when `deleteDoc` runs revives the tombstone (`Revived document ... by saving pending changes.` in `OTIndexedDBStore.savePendingChanges`): the docs row loses `deleted`, so the queued server delete — `allTracked.filter(t => t.deleted)` in `PatchesSync` is its only producer — is cancelled permanently, and the doc returns to `listDocs()` as live with one orphan pending row. Reproduced deterministically on this head vs main.

The dangerous shape is a doc that is not open: `Patches.deleteDoc` only drains `_changeQueues` via `closeDoc`, which is skipped then, so this lock was the sole serialization (bulk scene deletes, `PatchesBranchClient`'s post-merge `deleteDoc`).

Suggest restoring the lock on `deleteDoc` — `confirmDeleteDoc` genuinely doesn't need it — and correcting the L476 comment.

https://github.com/dabblewriter/patches/blob/f9af8e7c38591f8226a4087e4d556f82c861f492/src/client/OTAlgorithm.ts#L455-L463

src/client/Patches.ts
```diff
@@ -673,18 +677,19 @@ export class Patches {
    *   hiccup): the write MAY have landed (or may land on a later attempt), and
    *   nothing rejected it — discarding the ops would silently delete the user's
    *   un-persisted work. The ops stay applied (and queued in `_optimisticOps`)
-   *   and the submit is re-issued with backoff. Every attempt carries the same
-   *   stable change id, so id-based idempotency (OTAlgorithm._findChangeById,
-   *   server commit dedup) makes the re-submission exactly-once.
+   *   and the submit is re-issued with backoff. Re-submitting cannot duplicate:
+   *   a rejected IndexedDB transaction is atomic, so a failed savePendingChanges
```
RyanHavoc, line 681: This justification covers only the rejected-write subcase, and the same docstring contradicts it five lines up: the retried class is "the write MAY have landed". The spoke retries on a 30s hub RPC timeout where `savePendingChanges` often completed — the retry then persists a second pending row with the same stable id. `rebaseChanges` splices only the first id match (https://github.com/dabblewriter/patches/blob/f9af8e7c38591f8226a4087e4d556f82c861f492/src/algorithms/ot/shared/rebaseChanges.ts#L43-L49), so the echo clears one copy and the survivor is re-stamped to a baseRev behind its own commit — invisible to the server's baseRev-scoped read-side dedup. Both client backstops that caught exactly this (`_recentCommittedIds`, `_dropCommittedStragglers`, DAB-607) are deleted in this PR, and both were added after pup's write-time `_changeIds` marker existed, i.e. the server guard alone was judged insufficient at the time.

Reproduced: this head persists two rows and resends the survivor. Pup's write-time marker does block the double commit, so today's terminal cost is a wasted commit round-trip plus a strand in the queue between flushes — but correctness now rests solely on that one Firestore marker. Either keep the isRetry pre-scan, or restate this comment honestly as "the server's write-time id guard is the sole backstop for the landed-but-unacked case".

(Also: `PatchesHubService.handleDocChange`'s call site still passes `isRetry` — a TS2554 waiting on the version bump — and `legacyHeal.ts` documents its crash-rerun idempotency as resting on "isRetry plus the server pre-check".)

https://github.com/dabblewriter/patches/blob/f9af8e7c38591f8226a4087e4d556f82c861f492/src/client/Patches.ts#L676-L684

src/client/OTAlgorithm.ts
```diff
@@ -146,65 +144,84 @@ export class OTAlgorithm implements ClientAlgorithm {
...
+        let pendingSet = pending;
+        // A pending copy of a change already reflected in committedRev (stale echo) must be
+        // dropped before the rebase, matching applyCommittedChanges; rebaseChanges drops the new
+        // echoes.
+        if (staleServerChanges.length > 0 && pendingSet.length > 0) {
+          const staleIds = new Set(staleServerChanges.map(c => c.id));
+          pendingSet = pendingSet.filter(c => !staleIds.has(c.id));
+        }
+        rebased = newServerChanges.length > 0 ? rebaseChanges(newServerChanges, pendingSet) : pendingSet;
+        const result = await this.store.applyServerChanges(docId, serverChanges, rebased, tailRev);
```
RyanHavoc, line 205: On the closed-doc path this persists server changes that were never applied anywhere. main materialized via `getDoc` and applied before persisting, so an unappliable committed change threw `ApplyChangesError` into the `syncDoc` reload; here the poison is durably persisted and committedRev advances past it.

Repro over the IDB store: a poison change plus 250 good ones all persist silently; the first throw is at snapshot compaction (~rev 201), which does abort and heal — so there is a delayed backstop — but inside that window `openDoc` is a hard failure with no `ApplyChangesError` handler (`Patches.openDoc` rethrows), and on a quiet doc the window is weeks. Unappliable committed changes are real in this codebase: unknown op types on older pinned clients, and historically-lenient ops (`applyChangesForReconstruction`'s doc, DAB-601).

The rev arithmetic itself checked out (off-by-one, backward revs, root-replace catchup all correct). Suggest keeping the perf win but routing `getDoc`/`openDoc`'s `ApplyChangesError` into the same reload recovery, so deferred detection still lands somewhere that heals.

https://github.com/dabblewriter/patches/blob/f9af8e7c38591f8226a4087e4d556f82c861f492/src/client/OTAlgorithm.ts#L199-L206

src/net/PatchesSync.ts
```diff
@@ -1094,12 +1095,26 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
...
+      this.docStates.state[docId]?.committedRev ??
+      0;
+
     // Delegate to algorithm - it handles store updates and doc updates
     await algorithm.applyServerChanges(docId, serverChanges, doc);
 
     if (serverChanges.length > 0) {
       const lastRev = serverChanges[serverChanges.length - 1].rev;
       this._updateDocSyncState(docId, { committedRev: lastRev });
+      // The one choke point covering every path where server-committed changes became locally
```
RyanHavoc, line 1114: Not quite every path: `_reloadDocFromServer`'s default branch makes changes durable through `reconcilePending` and `saveDoc` (https://github.com/dabblewriter/patches/blob/f9af8e7c38591f8226a4087e4d556f82c861f492/src/net/PatchesSync.ts#L740-L750) without passing through here — and that branch is the target of `ApplyChangesError` recovery, the stale-doc 409, and `docReloadRequired`. A consumer keyed on `onServerCommit` silently misses every commit made durable by recovery. Either emit from the reload path too, or scope this comment and the "exactly once per committed change" framing in the `onServerCommit` JSDoc to the non-recovery paths.

https://github.com/dabblewriter/patches/blob/f9af8e7c38591f8226a4087e4d556f82c861f492/src/net/PatchesSync.ts#L1113-L1118

src/net/PatchesSync.ts
```diff
@@ -1094,12 +1095,26 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
       this._initDocSyncState(docId, {});
     }
 
+    // Durable tip BEFORE applying: onServerCommit fires only for changes that become newly
+    // durable now. A re-delivery carries revs already applied — DAB-773 echoes every flushed
+    // commit back to its sender, and catchup can overlap a broadcast — and applyServerChanges
+    // drops those, so emitting the raw batch would re-fire the same rev range and a per-change
+    // consumer (word counts, journal drain) would double-count.
+    const priorRev =
+      (doc as { committedRev?: number } | null | undefined)?.committedRev ??
```
RyanHavoc, line 1104: `??` only falls through on null/undefined, so an open doc's `committedRev` always beats `docStates` — including the torn-reload case documented at https://github.com/dabblewriter/patches/blob/f9af8e7c38591f8226a4087e4d556f82c861f492/src/client/OTAlgorithm.ts#L171-L177 ("a frame behind the store (never ahead)"). A stale-low `priorRev` over-includes: a re-delivery overlapping that window re-fires `onServerCommit` for revs already emitted, against the exactly-once JSDoc. `Math.max(doc.committedRev, docStates rev)` keeps the fallback and closes it. Note all four new `onServerCommit` tests run with `getOpenDoc` returning null, so the doc-present branch of this chain is untested.

https://github.com/dabblewriter/patches/blob/f9af8e7c38591f8226a4087e4d556f82c861f492/src/net/PatchesSync.ts#L1102-L1107

src/client/OTClientStore.ts
```diff
@@ -15,9 +15,11 @@ export interface OTClientStore extends PatchesStore {
...
    * @param docId Document identifier
+   * @param options.startAfterRev Only return pending changes with rev > startAfterRev (a ranged
+   *   read for foreign-mint deltas; no state materialization)
    * @returns Array of pending changes in chronological order
    */
-  getPendingChanges(docId: string): Promise<Change[]>;
+  getPendingChanges(docId: string, options?: { startAfterRev?: number }): Promise<Change[]>;
 
   /**
    * Appends new pending changes to the document's local change queue.
```
RyanHavoc, line 25: This docblock still describes the pre-R1 contract — "Changes should have sequential revision numbers starting after the last pending change" — but both implementations now re-stamp `rev` inside the persist transaction, overwriting the caller's values. Since this interface is what an external store implements, following it as written rebuilds exactly the cross-tab `[docId, rev]` clobber R1 exists to fix. Worth stating R1 here the way `applyServerChanges`' docblock states R2: rev is minted by the store inside the transaction from `max(committedRev, pending tail)`, and the passed objects are re-stamped in place.

https://github.com/dabblewriter/patches/blob/f9af8e7c38591f8226a4087e4d556f82c861f492/src/client/OTClientStore.ts#L24-L34

docs/operational-transformation.md
```diff
@@ -140,32 +140,27 @@ The server also broadcasts Alice's change to all other clients (including Bob).
...
-const newSnapshot = applyCommittedChanges(currentSnapshot, serverChanges);
-
-// Update storage and notify open documents
-await store.applyServerChanges(docId, serverChanges);
-
 const doc = patches.getOpenDoc(docId);
-if (doc) {
-  doc.applyCommittedChanges(serverChanges, newSnapshot.changes);
-}
+const rebasedPending = await store.applyServerChanges(docId, serverChanges, doc);
```
RyanHavoc, line 148: Both lines of this sample contradict the code it documents: the store's `applyServerChanges` third parameter is `rebasedPendingChanges: Change[]` — it is `algorithm.applyServerChanges` that takes the doc — and `onServerCommit` is not emitted unconditionally with the raw batch; `PatchesSync` filters to `newlyDurable` (`rev > priorRev`) and emits only when non-empty (https://github.com/dabblewriter/patches/blob/f9af8e7c38591f8226a4087e4d556f82c861f492/src/net/PatchesSync.ts#L1113-L1118), which is the exact re-delivery dedup this PR's own JSDoc promises. A consumer porting from this sample double-fires on every echo.

https://github.com/dabblewriter/patches/blob/f9af8e7c38591f8226a4087e4d556f82c861f492/docs/operational-transformation.md#L145-L150

jacwright:
One more from the wedge audit, on this branch's receive path: the convergence-exhaustion throw in `OTAlgorithm.applyServerChanges` ("did not converge after 10 attempts", OTAlgorithm.ts:225) is a plain `Error`, so `_receiveCommittedChanges` matches neither `MissingChangesError` nor `ApplyChangesError` and falls through to `onError.emit` — the batch is dropped, committedRev freezes, and the client keeps reporting synced. Under mint-anywhere, a foreign-mint race storm is exactly the condition that can exhaust the conflict loop, so the silent-stall case is reachable in v2.

Suggest routing it into the same `syncDoc(docId)` recovery as the other two (typed error, or a third recognized class): a fresh syncDoc re-enters the conflict loop with a fresh frame, which is the right retry for a race-driven exhaustion.

RyanHavoc (approved):

src/client/OTAlgorithm.ts
jacwright: Fixed in f5f2790. Confirmed: R1 mints from the shared store tail, so the foreign row lands below the doc's tail and the strict `>` never sees it — the ranged read was wrong at its anchor. `getPendingToSend` now unions the full store queue with the doc's in-memory pending via the same `_collectPending` the receive path uses: store rows win by id (so your sharper variant — a receive that transformed a store row in place — ships the transformed ops, not the doc's stale copy), doc-only rows (torn store write) are appended. Since that was the only consumer, the `startAfterRev` option is deleted from `getPendingChanges` (new in this PR, unreleased). Regression tests: foreign-row-below-doc-tail and store-copy-wins-by-id in OTAlgorithm-getPendingToSend.spec.ts.

src/client/OTAlgorithm.ts
jacwright: Fixed in f5f2790 — took the keep-the-lock option. `replacePendingChanges`, `reconcilePending`, and `ejectPendingChange` are back under `_withDocLock`; the docblocks now state the split you identified: R2 catches cross-context mints raising the store tail, the lock covers same-instance interleavings a rev-only watermark can't see (rev-neutral echo, and the eject window spanning the consent prompt — the L379 atomicity claim is true again under the lock). Deterministic regression test for your repro (echo landing between the replace's reads and its write) in OTAlgorithm-interleave.spec.ts. One residual, same on main: the replace's inputs come from flushDoc's unlocked getPendingToSend read, so an echo completing fully BEFORE the replace acquires the lock still writes the stale queue — that resurrection self-heals (server read-side id dedup at the old baseRev catches the resend, dropResolvedPending clears it), and main has the identical window.

src/client/OTAlgorithm.ts
jacwright: Fixed in f5f2790. Lock restored on `deleteDoc` exactly as 061b853 had it; `confirmDeleteDoc` left unlocked as you note. The `_withDocLock` docblock no longer claims every former holder relies on R2 — it names deleteDoc as outside R2 entirely, and documents why untrackDocs stays unlocked deliberately (eviction racing a mint should keep the minted row; the doc re-tracks next boot with the work intact). Regression test (tombstone survives an in-flight mint, no revive, no orphan row) in OTAlgorithm-interleave.spec.ts.

src/client/Patches.ts
jacwright: Fixed in f5f2790 — ended up taking your first option, and I want to flag why the dedup-only route I tried first was insufficient, since it's subtle: a per-row store dedup keyed on the stable id cannot pair a re-split batch — split pieces after the first used to get fresh random ids, so a maxStorageBytes-split retry would append pieces 2..N under new ids that neither the store nor pup's `_changeIds` marker can catch. Two pieces now: (1) breakChanges derives piece ids from the original (`id~2`, `id~3`, ...), so the whole batch is discoverable by the stable id — the store's per-row dedup pairs a re-split piece-for-piece, and it also makes flush-split resends visible to the server's id dedup; (2) the isRetry pre-scan is restored: `handleDocChange` regained the param (fixing the PatchesHubService TS2554 you flagged), `_processDocChange` passes `attempt > 0`, and `_findChangeById` prefix-scans store pending (the store, not the doc — an unacked landed write lives only in the store) then a bounded recent-committed window, returning the full landed batch instead of re-minting. On a pending hit with an open doc the pre-scan also applies the batch to the doc (append + optimistic shift, exactly what the failed attempt skipped) — without that the commit echo reads as foreign and re-applies the user's own ops. savePendingChanges dedups already-pending ids in-transaction as a second net for direct callers, and a fully-deduped call can't revive a tombstone. The `_processDocChange` docstring states the honest residuals: a landed batch that committed AND compacted before the retry ran rests on the server's write-time guard, and LWW has neither net (a retried `@inc` can double-consolidate — pre-existing, filed as DAB-797). `legacyHeal.ts`'s doc reference is dw3-side; it gets corrected in the dw3 PR when this releases. Tests: split-retry full-batch return, committed-retry idempotence, and open-doc heal in OTAlgorithm-stableId.spec.ts; per-row dedup incl. derived-id re-splits in OTIndexedDBStore-dedup.spec.ts.

src/client/OTAlgorithm.ts
jacwright: Fixed in f5f2790, keeping the perf win. Detection now lands somewhere that heals: `Patches._openDocOnce` emits a new `onDocLoadFailed` signal on `ApplyChangesError` (the open still rejects; pre-tracked docs only, so a heal can't race the failed-open unwind), and PatchesSync arms a per-doc poison flag consumed at the top of `syncDoc` — inside its serial gate — by throwing into syncDoc's existing ApplyChangesError recovery. That routing matters: a raw reload from the delivery path would race an in-flight flush (rewinding committedRev past commits its echo just installed), and it inherits the full retry ladder for free. The flag is armed at the `_applyServerChangesToDoc` choke point, so every delivery path is covered — broadcasts, flush echoes, branch merges (which rethrow to the caller but now also kick recovery), and reload drains — and it clears only when the recovery reload succeeds, so a transient failure mid-recovery keeps it armed for the ladder instead of orphaning the poison. After a successful recovery, surviving pending (e.g. offline edits predating the poison) gets a queued follow-up flush rather than stranding at synced+hasPending. A flag armed while disconnected waits for the reconnect's syncAllKnownDocs, so a poisoned doc that first surfaces during boot still heals. Compaction stays the quiet-doc backstop; the closed-doc receive comment states the deferred-detection trade explicitly.

src/net/PatchesSync.ts
jacwright: Fixed in f5f2790 — both options: `_reloadDocFromServer`'s default branch now emits the durable tail (committedTail when reconcile fetched it, else the envelope's installed changes, filtered to rev > the pre-reload baseRev), and the `onServerCommit` JSDoc is scoped honestly — changes subsumed into a snapshot body before the client ever saw them individually are not re-emittable, so it's "each delivered change exactly once", not a complete change log. One more gap your comment surfaced indirectly: a delivery whose store write lands but whose doc apply throws would have been emitted by NEITHER site (the throw skips the choke point; the recovery reload's baseRev already includes those revs). `_applyServerChangesToDoc` now emits the landed prefix — checked against the store's actual committedRev, never assumed — before rethrowing into recovery. Since the delivery paths aren't serialized against each other (DAB-773 echoes every flushed commit), both emit sites re-read the docStates floor immediately before emitting, so an echo and a broadcast of the same revs in flight together can't double-fire a per-change consumer. Tests for all three in PatchesSync.spec.ts.

src/net/PatchesSync.ts
jacwright: Fixed in f5f2790: `Math.max(doc.committedRev ?? 0, docStates rev ?? 0)`. Doc-present-branch test added (torn-reload shape: doc a frame behind the store, re-delivery must not re-fire).

src/client/OTClientStore.ts
jacwright: Fixed in f5f2790: `savePendingChanges`' interface docblock now states R1 the way `applyServerChanges`' states R2 — the store mints rev in-txn from max(committedRev, pending tail) and re-stamps the passed objects in place, callers' revs are provisional — plus the id-dedup and no-tombstone-revive obligations, so an external store implemented from the interface can't rebuild the [docId, rev] clobber.

docs/operational-transformation.md
jacwright: Fixed in f5f2790: the sample now goes through `algorithm.applyServerChanges(docId, serverChanges, doc)` and emits `onServerCommit` with the newly-durable filter; the surrounding text describes the store's actual third parameter (rebasedPendingChanges).

PR #123, opened 2026-07-20
feat(net): resume the SSE stream from a cursor on cross-tab hand-off

jacwright:
## TL;DR

When you close (or reload) the browser tab that's driving sync, another open tab takes over. Today that successor does a full re-sync of every document, which shows a "syncing" spinner and re-checks everything even though nothing changed. This makes the successor **resume the live stream from where the previous tab left off** instead: within a short window it replays only the gap of changes it missed, adds no per-document round-trips, and a clean hand-off never flashes "syncing". If the successor waited too long, it safely falls back to the old full re-sync. No behavior change for anyone who isn't handing off between tabs.

The server side (pup's `?lastEventId` replay + Redis-backed event store) already shipped; this is the client half that finally uses it.

## What changed

**`PatchesREST`**
- `connect(lastEventId?)` appends the cursor as a `?lastEventId=` query param. A freshly constructed `EventSource` can't set the `Last-Event-ID` request header (the browser only sends it on the same instance's own auto-reconnect), so a cross-tab resume has to pass the cursor in the URL, which pup reads as a `Last-Event-ID` equivalent.
- Tracks the id of each `changesCommitted` / `docDeleted` frame and exposes it via `lastEventId` for the app to persist across tabs.

**`PatchesSync`**
- `connect(lastEventId?)` arms a single-shot resume for the next `connected` transition.
- On a resumed connect, `syncAllKnownDocs({ resume: true })` skips the re-subscribe and the per-doc re-pull (the server restored this client's subscriptions from its own store, 2h TTL, and replays the committed gap over the stream) and only flushes docs still holding local pending — subscribing just those, since one may have been created offline and never subscribed. It stays on `synced` unless there's pending to flush, so a clean hand-off doesn't flash `syncing`.
- The server's `resync` frame (cursor too old / buffer gone) re-enters `connected` with the flag already cleared, so it naturally falls back to a full cold sync. No new failure path.

**`PatchesConnection` / `PatchesWebSocket`**: `connect(lastEventId?)` widened on the interface; WebSocket accepts and ignores it (no resumable stream).

## Tests

New resume specs in `tests/net/rest/PatchesREST.spec.ts` (cursor query param, `lastEventId` tracking) and `tests/net/PatchesSync.spec.ts` (resume skips re-subscribe/re-pull for clean docs, flushes + subscribes only pending docs, defers tombstone deletes, single-shot flag, cold-connect unchanged). Full suite green (3308 passed), lint + format clean.

## Consumer follow-up

dabble-writer-3.0 wires the cross-tab cursor persistence (localStorage stamp on the writer's unload/demotion, restore on the successor's takeover) against this API; it lands once this releases as 0.23.0 and the pin bumps.

### DIFF
--- a/src/net/PatchesConnection.ts
+++ b/src/net/PatchesConnection.ts
@@ -10,8 +10,20 @@ export interface PatchesConnection extends PatchesAPI {
   /** The server URL. Readable and writable — setting while connected triggers reconnection. */
   url: string;
 
-  /** Establish the connection to the server. */
-  connect(): Promise<void>;
+  /**
+   * Establish the connection to the server. `lastEventId`, when supplied, asks the
+   * server to resume the event stream after that id (SSE replay) instead of a cold
+   * start — the successor of a closed leader tab passes the predecessor's last id so
+   * the gap is replayed rather than re-synced. Transports without a resumable stream
+   * (WebSocket) ignore it.
+   */
+  connect(lastEventId?: string): Promise<void>;
+
+  /**
+   * The id of the last event received on the current stream, or undefined before the
+   * first event. Persisted cross-tab so a successor can resume from it (see `connect`).
+   */
+  readonly lastEventId?: string;
 
   /** Tear down the connection. */
   disconnect(): void;
--- a/src/net/PatchesConnection.ts
+++ b/src/net/PatchesConnection.ts
@@ -14,8 +14,11 @@ export interface PatchesConnection extends PatchesAPI {
    * Establish the connection to the server. `lastEventId`, when supplied, asks the
    * server to resume the event stream after that id (SSE replay) instead of a cold
    * start — the successor of a closed leader tab passes the predecessor's last id so
-   * the gap is replayed rather than re-synced. Transports without a resumable stream
-   * (WebSocket) ignore it.
+   * the gap is replayed rather than re-synced. The cursor is only honored for the
+   * clientId that earned it: the server keys its replay log and restored subscriptions
+   * by clientId, so the successor must also connect with the predecessor's clientId. A
+   * fresh clientId with an old cursor degrades safely to a full sync (`resync`), but
+   * the resume never fires. Transports without a resumable stream (WebSocket) ignore it.
    */
   connect(lastEventId?: string): Promise<void>;
 
--- a/src/net/PatchesConnection.ts
+++ b/src/net/PatchesConnection.ts
@@ -25,6 +25,18 @@ export interface PatchesConnection extends PatchesAPI {
    */
   readonly lastEventId?: string;
 
+  /**
+   * Whether the currently open stream was opened as a resume (with a `lastEventId`
+   * cursor) rather than a cold start — i.e. the server is replaying the missed gap.
+   * This is the authority PatchesSync consults on the `connected` transition to decide
+   * whether to run a resume-mode sync: it reflects what the transport *actually did*,
+   * so a cursor that never opened a resumed stream (offline defer, already-connected
+   * no-op, a `resync` re-anchor after the buffer expired, or a plain cold reconnect)
+   * correctly reads `false`. Transports with no resumable stream (WebSocket) leave it
+   * undefined, which reads as `false`.
+   */
+  readonly resumedStream?: boolean;
+
   /** Tear down the connection. */
   disconnect(): void;
 
--- a/src/net/rest/PatchesREST.ts
+++ b/src/net/rest/PatchesREST.ts
@@ -127,6 +127,8 @@ export class PatchesREST implements PatchesConnection {
   private _lastEventAt = 0;
   /** Id of the last id-bearing SSE frame received; the resume cursor (see `lastEventId`). */
   private _lastEventId: string | undefined;
+  /** Whether the currently open stream was opened with a resume cursor (see `resumedStream`). */
+  private _streamResumed = false;
   /** Interval handle for the half-open-stream watchdog; null while not connected. */
   private livenessTimer: ReturnType<typeof globalThis.setInterval> | null = null;
   /** Pending reconnect attempt scheduled after a fatal error / teardown; null when none. */
@@ -161,6 +163,16 @@ export class PatchesREST implements PatchesConnection {
     return this._lastEventId;
   }
 
+  /**
+   * Whether the currently open stream was opened as a resume (with a `lastEventId`
+   * cursor). Set true only when an EventSource is actually opened with the cursor, and
+   * reset to false on every cold open and on a server `resync` re-anchor — so it always
+   * reflects the live stream, never a stale intent. See {@link PatchesConnection.resumedStream}.
+   */
+  get resumedStream(): boolean {
+    return this._streamResumed;
+  }
+
   /** Current connection state of the underlying SSE stream. */
   get state(): ConnectionState {
     return this._state;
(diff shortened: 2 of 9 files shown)


src/net/rest/PatchesREST.ts
```diff
@@ -176,8 +183,16 @@ export class PatchesREST implements PatchesConnection {
 
     this._setState('connecting');
 
+    // EventSource cannot set a `Last-Event-ID` request header on a freshly
+    // constructed instance (the browser only sends it on its own auto-reconnect),
+    // so a caller-supplied resume cursor rides as a query param the server reads
+    // as a `Last-Event-ID` equivalent. Seed `_lastEventId` too, so a resume that
+    // yields no frames before the next hand-off still carries the cursor forward.
+    if (lastEventId) this._lastEventId = lastEventId;
```
jacwright, line 203: The resume cursor doesn't survive either early return above, but PatchesSync's resume flag does — and the two disagree.

Offline-deferred connect (the documented "wait for online" path): `connect('42')` returns at the offline check before this line runs, so the cursor is neither sent nor seeded. When the online listener fires, it calls `this.connect()` with no args, the stream opens cold, and the server neither replays nor sends `resync` (a cold open is normal from its side). Meanwhile `PatchesSync._resumeCursorPending` stayed armed — its `connect` resolved without error — so the eventual `connected` transition runs `syncAllKnownDocs({ resume: true })`: re-pull skipped, no replay to cover it. Changes committed by other devices during the window are silently missed until the next reconnect. Wake-from-sleep is the most common hand-off trigger (the browser discards the leader tab during sleep), and the successor often takes over while `navigator.onLine` is still false, so this is a mainline hand-off scenario rather than a corner.

Same shape via the `this.eventSource` early return: `connect(cursor)` while a stream exists leaves the flag armed with nothing sent, and a later watchdog/backoff rebuild (`_scheduleReconnect` → `connect()` no-args) opens cold yet consumes `resume: true` on its `connected` transition.

Suggested fix: a one-shot `_pendingResumeCursor` set before both early returns and consumed where the EventSource URL is built. The deferred/rebuilt stream then carries the cursor, and a by-then-stale cursor degrades safely through the server's `resync` → full sync. Optionally also clear `_resumeCursorPending` in `PatchesSync._handleConnectionChange` on a disconnected/error transition so the flag can't outlive the attempt it was armed for. A test for connect-while-offline → online reconnect would pin this.

src/net/PatchesConnection.ts
```diff
@@ -10,8 +10,20 @@ export interface PatchesConnection extends PatchesAPI {
...
-  /** Establish the connection to the server. */
-  connect(): Promise<void>;
+  /**
+   * Establish the connection to the server. `lastEventId`, when supplied, asks the
+   * server to resume the event stream after that id (SSE replay) instead of a cold
+   * start — the successor of a closed leader tab passes the predecessor's last id so
+   * the gap is replayed rather than re-synced. Transports without a resumable stream
+   * (WebSocket) ignore it.
+   */
+  connect(lastEventId?: string): Promise<void>;
```
jacwright, line 23: Worth one more sentence here: resume only works when the successor connects with the predecessor's clientId — the server keys both the replay log and the restored subscriptions by clientId (`replay(clientId, cursor)` in pup). Getting it wrong degrades safely to `resync` → full sync, but the optimization silently never fires, which is the kind of contract a future consumer loses an afternoon to.

jacwright (commented):
Verdict: request changes (GitHub blocks the formal state on the author's own PR).

Verified the diff against source on the branch and ran both touched suites (256 passing). The core design is right: flag armed before `connection.connect()` covers the synchronous transition, the unconditional `synced` at the tail of `syncAllKnownDocs` covers the resume pass, the server's `resync` re-anchor lands after the flag is consumed so the fallback is a genuine full sync, and pup reads `Last-Event-ID` header before the query param, so a browser-native auto-reconnect (fresh header, stale query in the URL) resumes from the right cursor.

One real correctness gap, inline on `PatchesREST.ts`: the resume cursor doesn't survive the offline-deferred connect (or a connect-while-connected no-op), but PatchesSync's resume flag does — the eventual reconnect runs in resume mode with no replay covering the gap. Wake-from-sleep hand-offs hit that path routinely, so requesting changes for that one.

Also worth doing in this PR: `docs/sse-rest.md` still documents the header-only contract — its sample handler reads only the `Last-Event-ID` header and the reconnection section doesn't mention `?lastEventId` or cross-tab resume. It's the protocol doc for this transport, so it should learn the query param alongside this change.

The brief window where a too-old cursor reads 'synced' until the `resync` frame arrives is acknowledged in the description and seems like the right trade.

RyanHavoc:
Pushed a follow-up (be46ead) addressing a resume-leak in the flag handling.

The resume decision was armed in `PatchesSync._resumeCursorPending` before the transport connected and consumed on the next `connected` transition. But `PatchesREST.connect` early-returns without opening a stream when offline (deferred) or already connected — so the flag could survive and be consumed by a later *cold* reconnect, running a resume-mode sync (skipping re-subscribe + per-doc re-pull for clean docs) over a stream the server never replayed, silently stranding those docs until the next resync. Reachable via a cross-tab hand-off that fires while briefly offline: `connect(cursor)` defers, then the online reconnect (`connect()`, no cursor) opens cold but consumes the armed flag.

Fix makes the transport the authority instead of caller intent: `PatchesREST` tracks whether the *live* stream was opened with a cursor (`resumedStream`), set only at a real open and cleared on every cold open and on a server `resync` re-anchor. `PatchesSync` reads `connection.resumedStream` on the `connected` transition rather than a pre-armed flag, so a cursor that never opened a resumed stream falls back to a full sync. WebSocket leaves `resumedStream` undefined (→ false). Also removes the resync-vs-flag coupling the old comment relied on.

Added regression coverage: the offline-defer case runs a full sync (PatchesSync), and `resumedStream` reflects cursor-open vs cold-open and clears on `resync` (PatchesREST). Full suite green (3311), type/lint/format clean.

jacwright:
Verified be46ead — it resolves the request-changes finding, and the transport-authority design is better than the pending-cursor carry I suggested: it eliminates the whole intent-vs-stream desync class instead of patching one leak.

Traced the lifecycle edges: offline-defer and already-connected no-ops now cold-sync (correct); the `resync` re-anchor clears the flag before cycling state; a failed initial handshake can leave `_streamResumed` stale-true but nothing can consume it — every path to a new `connected` goes through `connect()` (which resets it) or the resync handler (which clears it).

One behavior widening worth stating as intended: a browser-native auto-reconnect re-fires `onopen` without passing through `connect()`, so a transient blip on a resumed stream now runs a resume pass where the old code full-synced. I traced it as correct under pup's header-over-query precedence — a fresh `Last-Event-ID` header wins, a headerless reconnect falls back to the original query cursor (a superset of the gap), and too-old degrades through `resync` to a full sync. Net improvement: blips on a resumed stream stop paying the full re-sync tax.

Still open before merge, both small: the `docs/sse-rest.md` update (query param + resume semantics; it still documents the header-only contract) and the clientId-contract line on `connect()`'s JSDoc. Optional follow-up, not blocking: carry the cursor into the offline-deferred open (one-shot, consumed at EventSource construction) so the wake-from-sleep hand-off resumes too — it composes cleanly with `resumedStream`, since the deferred open would then legitimately mark itself resumed.

Re-verified locally at be46ead: full suite 3311 passed, type/lint/format clean.

jacwright:
Docs items from the review are addressed in 51f9992: sse-rest.md gains the `?lastEventId` query param in the sample route (header first, query fallback — matching pup), a Cross-Tab Hand-Off section covering the client wiring, the same-clientId contract, and the `resumedStream` authority, plus a tier-table row and a pointer from the Client ID section for the sanctioned id-adoption exception. `PatchesConnection.connect`'s JSDoc now states the clientId contract too. Full suite green at 51f9992 (3311), type/lint/format clean. Nothing left open from my review — merge-ready from my side.

jacwright:
Added `chore(release): 0.23.0` (286e03e) — bumps package.json / package-lock.json, folding the resume feature + the fixes on this branch into a single release, so merging publishes 0.23.0 directly (no separate release PR). dabble-writer-3.0 then bumps its pin to `^0.23.0`.

RyanHavoc (approved):

PR #124, opened 2026-07-20
Anchor the SSE resume cursor with a connected frame

jacwright:
## TL;DR

When you close or reload the tab that's driving sync, another open tab takes over. Today, in the common case (you're the only person editing your own docs), that hand-off does a full re-sync every time — spinner, round-trips, the works — even though it missed a few seconds at most. This makes the hand-off resume where the previous tab left off instead. The trick: the server now hands every stream its current position the moment it connects (in a `connected` event that also carries the Pup version), so a tab always has something to hand off, even one that never received a change.

## The bug this closes

The resume cursor only ever came from a delivered `changesCommitted`/`docDeleted` frame. But a client that only *sends* edits — the sole editor of its own docs, with its own commits excluded from its own stream (DAB-773) — receives zero id-bearing frames, holds no cursor, and every cross-tab hand-off cold-syncs. All the resume machinery shipped in 0.23.0 is inert in exactly that case. (`"0"` as a synthetic cursor doesn't work: Pup's Redis store keys resume off real stream ids and treats a floor value as unverifiable → resync.)

## What changed

- **`SSEEventStore.currentId(clientId)`** (optional): returns the id a fresh connect should anchor to — one its own `replay` accepts as up-to-date (empty continuation, not resync) — minting a durable marker when the client has none. The in-memory store returns the last assigned id, or `"0"` before the first append.
- **`SSEServer`** emits an id-bearing `connected` frame (id from `currentId`, opaque `connectedData` payload) at the start of every connect, **before** any replay so replayed events advance the cursor past the anchor rather than the reverse. A store without `currentId`, or a degraded call, yields an id-less frame and the client cold-syncs — exactly as before.
- **`PatchesREST`** adopts the cursor from the `connected` frame and exposes the payload as `serverInfo`; it also adopts the re-anchor id from a `resync` frame it previously dropped, so the next reconnect verifies instead of resyncing again.
- Docs: `docs/sse-rest.md` gains a "Connected Anchor" section.

## Release

Version bumped to **0.24.0** on this branch (`chore(release): 0.24.0`) — merging publishes it. Pup's matching half (`RedisSSEEventStore.currentId` + `connectedData: { version }`) is ready on its own branch and depends on 0.24.0; it bumps the dep and deploys after this publishes. dw3 only needs a pin bump — no logic change, the cursor now populates itself.

## Tests

All 3319 pass. New coverage: in-memory `currentId`, the `connected` frame (id from `currentId`, `connectedData`, id-less fallbacks, ordering before replay), and `PatchesREST` adopting both the connected cursor + server info and the resync re-anchor id.

### DIFF
--- a/docs/sse-rest.md
+++ b/docs/sse-rest.md
@@ -333,6 +333,20 @@ This is where SSE + REST actually shines compared to WebSocket.
 
 The cursor reaches the server two ways. The browser's own auto-reconnect sends the `Last-Event-ID` header — but only on the same `EventSource` instance. A freshly constructed `EventSource` can't set that header at all, so a caller-supplied cursor (see [Cross-Tab Hand-Off](#cross-tab-hand-off)) rides as a `?lastEventId=` query param instead, which the server reads as a header equivalent. Header wins when both are present — it's always the fresher of the two.
 
+### The Connected Anchor
+
+There's a hole in "EventSource remembers the last id it received": what if it never receives one? Ids ride `changesCommitted` and `docDeleted` frames. A client that only ever _sends_ edits — the sole editor of its own docs, with its own commits excluded from its own stream — gets zero id-bearing frames and holds no cursor. Its hand-off then full-syncs every time, because there's nothing to hand off.
+
+So every stream opens with a `connected` frame carrying the client's current cursor, which the store mints when the client has no history yet:
+
+```
+id: 1784516082085-0
+event: connected
+data: {"version":"3.15.0"}
+```
+
+Now the cursor is set the moment the stream opens, before any change arrives. `PatchesREST` exposes it as `lastEventId` (the resume cursor) and the frame's `data` as `serverInfo` (opaque — the server puts whatever it likes there; Pup sends its version). The store decides what the id is via `SSEEventStore.currentId`: it must be one a later `replay` accepts as up-to-date, so resuming from it is an empty continuation, not a resync. A store that omits `currentId`, or a degraded call, yields an id-less `connected` frame and the client cold-syncs — exactly as before.
+
 ### The Three Tiers
 
 | Scenario            | Duration | What happens                                                             |
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@dabble/patches",
-  "version": "0.23.0",
+  "version": "0.24.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@dabble/patches",
-      "version": "0.23.0",
+      "version": "0.24.0",
       "license": "MIT",
       "dependencies": {
         "@dabble/delta": "^1.2.5",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dabble/patches",
-  "version": "0.23.0",
+  "version": "0.24.0",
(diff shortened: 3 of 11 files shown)


jacwright (commented):
**Review: LGTM.** Clean fix for a real, well-diagnosed gap. Verified locally on `feat/sse-connected-anchor`: `tsc --noEmit` clean, SSE/REST suites 171/171 green.

### What it does
Closes the inert-resume hole from 0.23.0: a client that only *sends* edits (sole editor of its own docs, own commits excluded per DAB-773) never received an id-bearing frame, so it held no cursor and every cross-tab hand-off cold-synced. Now every stream opens with an id-bearing `connected` frame carrying the store's `currentId`, so a tab always has a cursor to hand off, even one that never saw a change. The frame's `data` is opaque server info (Pup sends `{version}`), surfaced as `serverInfo`.

### Strengths
- **Ordering is correct and tested.** The `connected` anchor is enqueued on the same per-client pipeline *before* the replay task, so replayed higher ids advance the cursor past the floor, never the reverse. The `sends the connected anchor before any replayed events` test locks this in.
- **Graceful degradation.** `currentId` is optional; a missing method or a `_guardStoreCall` timeout/rejection yields an id-less frame → client cold-syncs exactly as before. Both fallbacks are tested.
- **`"0"` floor is sound.** `(nextEventId ?? 1) - 1` returns one below the first real id, and `replay('c1','0')` is asserted to be an empty continuation, not a resync — matching the contract the JSDoc spells out.
- **Scope-clean.** Stays inside the transport seam (`SSEEventStore`/`SSEServer`/`PatchesREST`), no app vocabulary. `connectedData` serialized once at construction. Docs updated in-house style.

### Non-blocking notes
- **`resync` cursor carry-over.** The new `if (e.lastEventId) this._lastEventId = e.lastEventId` in the resync handler: when the server sends a resync with *no* re-anchor id, native EventSource still carries the last seen id on `e.lastEventId` (persists across frames), so the client adopts a carried-over value rather than "nothing". Not a hazard — the very next connect's `connected` anchor overwrites it with a verified-current `currentId`, so it self-heals. Worth a one-liner in the comment noting EventSource's persist-across-frames semantics, since the mock passes an explicit id and hides this.
- **`serverInfo` typed `unknown`.** Fine as an opaque payload, but consumers reading `.version` will cast. Acceptable given it's explicitly informational.

### Merge sequencing (from the PR body — flagging, not blocking)
`chore(release): 0.24.0` is on the branch, so merge publishes. Order holds: merge#124 → publish (OTP) → pup dep bump + PR + deploy (`RedisSSEEventStore.currentId` + `connectedData:{version}`) → dw3 pin `^0.24.0`. Pup's half must ship before dw3 relies on the populated cursor.

No changes requested.

jacwright:
Addressed the review notes in `c47df08`:

- **resync cursor carry-over** — expanded the comment on the resync handler to spell out EventSource's persist-across-frames `lastEventId` semantics: an id-less resync carries the last real id (not nothing), which self-heals via the next connect's `connected` anchor. Behavior unchanged; just documents what the explicit-id mock hides.
- **`serverInfo: unknown`** — left as-is by design. Typing it to a `{ version }` shape would couple the transport to Pup's payload; it's explicitly opaque/informational, so consumers cast.

No functional change since the LGTM — `tsc` + SSE/REST suites still green.

PR #125, opened 2026-07-21
fix(lww): refresh open follower doc when the shared store advanced past a fan

jacwright:
## TL;DR

Open project docs in a background tab now receive live updates. Before this, if you had the same project open in two tabs (or windows), settings-style data (project meta, goals, trash, and other LWW-synced fields) that changed in one tab would silently fail to update in the other until you reloaded it. This makes those background tabs stay current on their own.

## What was wrong

LWW docs (everything except document content: project meta, settings, goals, trash) share one IndexedDB store across a tab group, but each tab holds its own in-memory copy of an open doc. One tab is elected the writer; it applies incoming server changes and then fans the same batch to the other tabs as a latency optimization.

Crucially, the writer persists the commit to the shared store **before** it fans it. So by the time a follower tab runs `applyServerChanges` for that fanned batch, its own store watermark (`committedRev`) has already been advanced by the writer. Every op in the batch then stale-skips against the cross-batch ordering guard, `effectiveChanges` comes out empty, and the method returned early — never touching the follower's open in-memory doc. That doc was left behind permanently, since no later fan re-delivers a batch below the advanced rev.

(OT content docs don't have this problem: a later fan heals a misaligned OT open doc via a full reload+import. The follower reconciliation surface for LWW was simply empty.)

## The mechanism

When the stale-skip leaves `effectiveChanges` empty, instead of returning empty-handed we now refresh the open doc from the store, rebasing it onto current committed state while **preserving any local optimistic edit** (the existing `ejectPendingChange` precedent). The whole thing stays inside the per-doc lock.

The refresh snapshot is built from committed-only state (`getCommittedState`) plus the doc's own sending/pending layers, handed to `doc.import`. This matters: `store.getDoc` folds sending+pending **into** its snapshot state _and_ returns them again in `snapshot.changes`, and `import` re-applies the changes on top — idempotent for `replace` ops but doubling a combinable op like `@inc`. Building the snapshot from committed-only state means each in-flight op applies exactly once, so a pending `@inc` reads correctly through the heal and stays correct when the commit echo pure-matches its in-flight key. After the store reads we re-check `rev > doc.committedRev` before importing, so a concurrent `applySnapshot` (which runs outside this lock) that already advanced the doc to the store rev can't trigger an equal-rev rebuild and its needless churn emit.

The stale batch itself is never force-applied — its old field values would regress newer committed fields, which is exactly why the stale-skip guard exists. We refresh from the store's current truth rather than replaying the old batch.

## The watermark decision

The refresh is gated on `doc.committedRev < storeCommittedRev` (the doc being genuinely behind the shared store). The RCA flagged a false-negative risk: a minting follower applies its own edit optimistically, and if that advanced the doc's committed watermark, a behind-doc could read as current and skip the refresh.

Verified in code that it does **not**: `handleDocChange` mints a change with `committedAt: 0`, and `LWWDoc.applyChanges` only advances `_committedRev` for changes with `committedAt > 0`. So `committedRev` is a clean **server-only** watermark — optimistic/pending applies never move it. That makes the simple `doc.committedRev < storeCommittedRev` predicate both correct and minimal:

- A behind follower (writer already advanced the shared store) → refreshes.
- An own-echo re-delivery to an already-current doc (doc and store at the same rev) → does not refresh, so no pointless full-state rebuild per commit.

This is pinned by four tests (`tests/client/LWWAlgorithm-followerRefresh.spec.ts`): the fan-blindness heal (fails without the fix), the own-echo no-rebuild guard (asserts `store.getDoc` is never called), optimistic-edit preservation across the heal, and delta-fold-exactly-once through the commit echo (the last two fail without the fix).

## Relationship to dw3 sync-v2

dw3's sync-v2 ships an app-side fallback in `PupSync._applyServerChangesToDoc` that reconciles this same follower gap from the app layer. Once this library fix releases and dw3 picks it up, that app-side fallback becomes redundant and can be removed.

## Validation

- `npm run type:check`, `npm run lint`, `npm run format:check` — all clean
- `npm run test` — 120 files, 3323 passed, 4 pre-existing skips
- `npm run build` — succeeds

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@dabble/patches",
-  "version": "0.24.0",
+  "version": "0.24.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@dabble/patches",
-      "version": "0.24.0",
+      "version": "0.24.1",
       "license": "MIT",
       "dependencies": {
         "@dabble/delta": "^1.2.5",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dabble/patches",
-  "version": "0.24.0",
+  "version": "0.24.1",
   "description": "Immutable JSON Patch implementation based on RFC 6902 supporting operational transformation and last-writer-wins",
   "author": "Jacob Wright <jacwright@gmail.com>",
   "bugs": {
--- a/src/client/LWWAlgorithm.ts
+++ b/src/client/LWWAlgorithm.ts
@@ -196,7 +196,23 @@ export class LWWAlgorithm implements ClientAlgorithm {
         effectiveChanges.push(change);
         cursor = Math.max(cursor, change.rev);
       }
-      if (effectiveChanges.length === 0) return [];
+      if (effectiveChanges.length === 0) {
+        // Follower fan-blindness heal. A SECOND context sharing this store can hold the doc
+        // open while a FIRST context (the writer) already advanced the shared store past the
+        // doc's watermark — the writer persists before it fans, so every fanned batch then
+        // stale-skips above and the open doc, whose in-memory state predates the store, would
+        // never catch up. Refresh it from the store (which folds committed + sending + pending,
+        // preserving local optimistic edits) rather than returning empty-handed. Never
+        // force-apply the stale batch itself — its old field values would regress newer
+        // committed fields. Guard on the doc being genuinely behind: committedRev is a
+        // server-only watermark (optimistic applies never advance it), so an own-echo
+        // re-delivery to an already-current doc can't trip this into a needless rebuild.
+        if (doc && doc.committedRev < committedRev) {
+          const snapshot = await this.loadDoc(docId);
+          if (snapshot) (doc as LWWDoc<T>).import(snapshot);
+        }
+        return [];
+      }
 
       // Apply server changes to store (preserves sendingChange and pendingOps)
       await this.store.applyServerChanges(docId, effectiveChanges);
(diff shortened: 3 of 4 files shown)


RyanHavoc:
Reviewed — **approve** (leaving the formal approval to Ryan). Verified against the branch, not just the description:

- The heal sits inside `_withDocLock` (LWWAlgorithm.ts:160), so it can't interleave with another apply on the same doc.
- The **committed-only base + sending/pending re-applied via `snapshot.changes`** construction is the important call here, and it's right: `store.getDoc` folds in-flight layers into state *and* returns them as changes, which `import` re-applies — idempotent for `replace`, doubling for `@inc`. Building from `getCommittedState` makes each in-flight op fold exactly once, and the delta-fold test pins it.
- The `doc.committedRev < committedRev` gate is proven server-only (mints carry `committedAt: 0` and `LWWDoc.applyChanges` only advances the watermark for `committedAt > 0`), so an own-echo re-delivery can't trigger a needless full-state rebuild — and the no-rebuild test asserts `store.getDoc` is never called on that path.
- Post-await `rev > doc.committedRev` re-check correctly covers a concurrent `applySnapshot` that ran outside this lock.
- The stale batch is never force-applied — refreshing from store truth instead of replaying old field values is the right resolution of the stale-skip guard's tension.

All 4 tests pass on the branch. Two notes:

1. **Cross-repo interplay verified:** dw3's per-delivery belt (dabblewriter/dabble-writer-3.0#1211, e1d50b70b) checks `storeRev > doc.committedRev` after each fan — once this releases and the doc heals in-library, that check reads equal and the belt no-ops per delivery. Removing the belt after the version bump is exactly as described, with no window where both run destructively.
2. **Small ask while you're in here:** dw3's reconciliation reaches the protected `patches.docs` map through a structural cast to enumerate open docs (`PupSync.openDocIds`). A public read-only accessor (e.g. `openDocIds()` or an iterator on `Patches`) in this or a follow-up release would let that cast — and its rename-risk latch — go away.

RyanHavoc (approved):

PR #126, opened 2026-07-22
fix(ot): reject interior-gapped committed batches instead of skipping content

RyanHavoc:
**TL;DR:** An OT follower could advance its in-memory `committedRev` past committed content it never applied — a doc reading `committedRev == store rev` while its state is behind, un-healable by reconciliation (which gates on a strictly-higher store rev). Root cause: the receive path checked batch contiguity on the **first** change only. This rejects interior-gapped committed batches at three layers. Bumps 0.24.1 → 0.24.2.

## The bug (confirmed via sync-v2 production stress testing)

A follower receives a committed batch with an interior hole — e.g. `[148, 151]` with 149/150 dropped by a partial fan / self-echo exclusion. `OTDoc.applyChanges` guarded only `serverChanges[0].rev` (the leading edge), then applied whatever ops were present and set `_committedRev` to the **last** change's rev. So 148 and 151 applied, **149/150 were silently skipped**, and the watermark jumped to 151.

Because the store (written by a peer) stays complete at 151, the follower reads `committedRev == storeRev == 151` yet is missing content. The reconciliation audit heals only when `storeRev > doc.committedRev`, so `151 > 151` is false and it never fires — only a reload recovers. We reproduced this on `beta` (production backend): an open follower stayed content-behind through 66 s of genuine foreground; reload always healed.

## The fix — three layers

- **`OTAlgorithm.buildFrame`** now flags an interior hole in the new frame (not just the leading edge). A store-behind delivery therefore throws `MissingChangesError` **before** the store write, routing to `getChangesSince` recovery instead of writing a gapped batch into the store.
- **`OTAlgorithm.applyServerChanges`** routes an internally non-contiguous batch to a full store rebuild (`loadDoc` + `import`) rather than the incremental apply, so an open doc converges from the complete committed store (the observed case, where the store is already complete).
- **`OTDoc.applyChanges`** refuses an interior-gapped committed batch outright — a last-line invariant that makes "watermark advanced past unapplied content" unrepresentable for any caller.

The contiguity check starts at **index 1**, so a root-replace catchup (a legitimate first-vs-base rev jump, internally contiguous afterward) is unaffected.

## Tests

`tests/client/OTAlgorithm-batchContiguity.spec.ts` (5): the interior-gap invariant throws with no mutation; a contiguous batch applies normally; a store-behind gapped batch throws `MissingChangesError` without advancing the store; **the observed case — an open follower at rev 100 receiving `[101, 104]` against a complete store rebuilds to 104 with the skipped content present**; a contiguous fan still applies incrementally.

Full suite green (3328 passed), heavy convergence fuzz green (2073), `type:check` / `lint` / `format:check` clean.

## Deliberately scoped out (follow-up)

`applyCommittedChanges` (the reconstruction path — version builds, scrub, branch) has the **identical** first-element-only check and the same skip-to-last-rev shape. I did **not** mirror the guard there in this PR because that path may legitimately take sparse inputs; it needs a caller analysis first. Flagging it as the obvious next step.

## Context

Found during sync-v2 production stress testing (dabblewriter/dabble-writer-3.0#1211). A dw3-side backstop (reconcile on content divergence at equal rev) ships separately; this is the root-cause fix. Full writeup + mechanism in the handover/remediation docs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dabble/patches",
-  "version": "0.24.1",
+  "version": "0.24.2",
   "description": "Immutable JSON Patch implementation based on RFC 6902 supporting operational transformation and last-writer-wins",
   "author": "Jacob Wright <jacwright@gmail.com>",
   "bugs": {
--- a/src/client/OTAlgorithm.ts
+++ b/src/client/OTAlgorithm.ts
@@ -22,6 +22,21 @@ import type { TrackedDoc } from './PatchesStore.js';
 const APPLY_CONFLICT_RETRIES = 10;
 
 /**
+ * Index of the first change whose rev is not exactly one past its predecessor's — i.e. the first
+ * interior hole in a rev run — or -1 when the run is fully dense. Server revs are dense per doc, so
+ * any such gap in a committed batch is a delivery defect, not a legitimate sparse batch. Shared by
+ * the two rev-contiguity scans in this file (buildFrame's new-change frame and applyServerChanges'
+ * branch selector). OTDoc.applyChanges deliberately keeps its own inline scan as an independent
+ * last-line invariant that does not lean on this layer.
+ */
+function firstGapIndex(changes: Change[]): number {
+  for (let i = 1; i < changes.length; i++) {
+    if (changes[i].rev !== changes[i - 1].rev + 1) return i;
+  }
+  return -1;
+}
+
+/**
  * OT (Operational Transformation) algorithm implementation.
  *
  * OT uses revision-based history and rebasing for concurrent edits.
@@ -160,6 +175,10 @@ export class OTAlgorithm implements ClientAlgorithm {
         const staleC: Change[] = [];
         for (const change of serverChanges) (change.rev > base ? newC : staleC).push(change);
         let gap = false;
+        // The actual hole boundary for an INTERIOR gap, so the throw below can report the real
+        // missing revs. Undefined for a leading-edge gap, where newServerChanges[0].rev is already
+        // the right diagnostic.
+        let gapAt: { expected: number; got: number } | undefined;
         if (newC.length > 0 && newC[0].rev !== base + 1) {
           const first = newC[0];
           const isRootReplaceCatchup =
@@ -172,14 +191,13 @@ export class OTAlgorithm implements ClientAlgorithm {
         // an interior hole too so it routes to MissingChangesError recovery (or the store-rev
         // re-check below) rather than being written to the store and skipping content.
         if (!gap) {
-          for (let i = 1; i < newC.length; i++) {
-            if (newC[i].rev !== newC[i - 1].rev + 1) {
-              gap = true;
-              break;
-            }
+          const i = firstGapIndex(newC);
+          if (i !== -1) {
+            gap = true;
+            gapAt = { expected: newC[i - 1].rev + 1, got: newC[i].rev };
           }
         }
-        return { newC, staleC, gap };
+        return { newC, staleC, gap, gapAt };
       };
 
       // Trust the open doc's committedRev optimistically. The one case it is wrong is a torn
@@ -188,16 +206,24 @@ export class OTAlgorithm implements ClientAlgorithm {
       // re-check the store's committedRev (the authority) before declaring one; the aligned path
       // stays store-read-free.
       let committedRev = otDoc ? otDoc.committedRev : await this.store.getCommittedRev(docId);
-      let { newC: newServerChanges, staleC: staleServerChanges, gap } = buildFrame(committedRev);
+      let { newC: newServerChanges, staleC: staleServerChanges, gap, gapAt } = buildFrame(committedRev);
       if (gap && otDoc) {
         const storeRev = await this.store.getCommittedRev(docId);
         if (storeRev > committedRev) {
           committedRev = storeRev;
-          ({ newC: newServerChanges, staleC: staleServerChanges, gap } = buildFrame(committedRev));
+          ({ newC: newServerChanges, staleC: staleServerChanges, gap, gapAt } = buildFrame(committedRev));
         }
       }
       if (gap) {
-        throw new MissingChangesError(committedRev + 1, newServerChanges[0].rev, committedRev);
+        // sinceRev MUST stay committedRev — recovery pulls the tail from there and that is correct
+        // regardless of where the hole is. Only the diagnostic expected/got reflect the actual
+        // hole: an interior gap reports its real boundary (gapAt), a leading-edge gap the first
+        // new rev.
+        throw new MissingChangesError(
+          gapAt?.expected ?? committedRev + 1,
+          gapAt?.got ?? newServerChanges[0].rev,
+          committedRev
+        );
       }
 
       // Rebase pending and persist, retrying if a foreign mint raced the replace (R2). Each retry
@@ -229,17 +255,14 @@ export class OTAlgorithm implements ClientAlgorithm {
 
       if (otDoc) {
         // `serverChanges` is internally contiguous when the frame passed the gap check above,
-        // EXCEPT when a store-rev re-check absorbed an interior-gapped frame (newC emptied), in
-        // which case the original batch is still non-contiguous — verify it here so such a batch
-        // rebuilds from the store instead of advancing the in-memory watermark past skipped
-        // content via the incremental apply.
-        let contiguous = true;
-        for (let i = 1; i < serverChanges.length; i++) {
-          if (serverChanges[i].rev !== serverChanges[i - 1].rev + 1) {
-            contiguous = false;
-            break;
-          }
-        }
+        // EXCEPT when the store-rev re-check re-anchored the frame off a higher store rev and so
+        // absorbed an interior-gapped batch. That is broader than the newC-emptied case: if the
+        // store rev sits at the hole's trailing edge (store at 150 for a batch [148, 151]), the
+        // rebuilt frame passes with a NON-empty newC = [151] — leading edge clean off the store
+        // rev — yet the original batch [148, 151] is still non-contiguous. Re-scan the actual
+        // `serverChanges` array (not newC) here so either shape rebuilds from the store instead of
+        // advancing the in-memory watermark past skipped content via the incremental apply.
+        const contiguous = firstGapIndex(serverChanges) === -1;
         if (contiguous && otDoc.committedRev === serverChanges[0].rev - 1) {
           otDoc.applyChanges(changesToBroadcast);
         } else {
--- a/src/client/OTAlgorithm.ts
+++ b/src/client/OTAlgorithm.ts
@@ -166,6 +166,19 @@ export class OTAlgorithm implements ClientAlgorithm {
             first.ops.length === 1 && first.ops[0].op === 'replace' && first.ops[0].path === '';
           gap = !isRootReplaceCatchup;
         }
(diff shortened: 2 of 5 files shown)


src/client/OTAlgorithm.ts
```diff
@@ -166,6 +166,19 @@ export class OTAlgorithm implements ClientAlgorithm {
...
           gap = !isRootReplaceCatchup;
         }
+        // Interior contiguity: server revs are dense, so a hole *between* new changes (e.g.
+        // [148, 151] with 149/150 dropped by a partial fan) is a delivery defect, not a
+        // root-replace catchup. The first-element check above only sees the leading edge; catch
+        // an interior hole too so it routes to MissingChangesError recovery (or the store-rev
+        // re-check below) rather than being written to the store and skipping content.
+        if (!gap) {
+          for (let i = 1; i < newC.length; i++) {
+            if (newC[i].rev !== newC[i - 1].rev + 1) {
```
matalina: Diagnostics nit: when this interior check is what trips (leading edge clean, no absorption), the `MissingChangesError` below goes out as `(committedRev + 1, newServerChanges[0].rev, committedRev)` — and `newServerChanges[0].rev === committedRev + 1`, so the message reads "Expected rev 101, got 101." Recovery only consumes `sinceRev`, so it heals fine, but the log line is confusing next to a real leading-edge gap. Consider capturing the hole boundary here (`newC[i - 1].rev + 1` / `newC[i].rev`) and passing that pair through when the gap is interior.

src/client/OTAlgorithm.ts
```diff
@@ -215,11 +228,24 @@ export class OTAlgorithm implements ClientAlgorithm {
       const changesToBroadcast = [...serverChanges, ...rebased];
 
       if (otDoc) {
-        if (otDoc.committedRev === serverChanges[0].rev - 1) {
+        // `serverChanges` is internally contiguous when the frame passed the gap check above,
+        // EXCEPT when a store-rev re-check absorbed an interior-gapped frame (newC emptied), in
```
matalina: Comment nit: absorption isn't only the `newC`-emptied case. If the store rev sits at the hole's trailing edge — e.g. store at 150 for `[148, 151]` — the rebuilt frame passes with `newC = [151]` (leading edge clean off the store rev), and the original batch is still non-contiguous. The code below handles both correctly since it scans the actual array; just widen the wording so a future reader doesn't narrow this check to the emptied case.

src/client/OTAlgorithm.ts
```diff
@@ -215,11 +228,24 @@ export class OTAlgorithm implements ClientAlgorithm {
...
+        // which case the original batch is still non-contiguous — verify it here so such a batch
+        // rebuilds from the store instead of advancing the in-memory watermark past skipped
+        // content via the incremental apply.
+        let contiguous = true;
```
matalina: This is the third copy of the same scan (the interior check in `buildFrame`, and `OTDoc.applyChanges`' guard). Keeping OTDoc's copy independent makes sense — it's a last-line invariant that shouldn't lean on the algorithm layer — but the two in this file could share a tiny `firstGapIndex(changes)` helper so the contiguity definition can't drift.

src/client/OTDoc.ts
```diff
@@ -225,6 +225,21 @@ export class OTDoc<T extends object = object> extends BaseDoc<T> {
...
+      // Committed revs are dense per doc, so an interior hole in this batch (e.g. [148, 151]
+      // with 149/150 dropped by a partial fan / self-echo exclusion) is always a delivery
+      // defect. Without this guard the present ops apply and `_committedRev` advances to the
+      // LAST rev, silently skipping the missing content — a doc that reads as caught up
+      // (committedRev == store rev) while its state is behind, which the reconciliation audit
+      // cannot detect (it gates on store rev being strictly greater). Refuse it instead; the
+      // OT receive path routes a non-contiguous batch to a full store rebuild.
+      for (let i = 1; i < serverChanges.length; i++) {
+        if (serverChanges[i].rev !== serverChanges[i - 1].rev + 1) {
+          throw new Error(
```
matalina, line 244: Worth a line in the comment noting this is intentionally a plain `Error`, not `MissingChangesError`: it signals a caller bug (the algorithm layer must never hand this method a gapped committed batch), not a recoverable transport gap, so it shouldn't trigger PatchesSync's `getChangesSince` routing. Nothing depends on the type today — the only committed-path caller is `OTAlgorithm` — but the note stops a future reader from "fixing" it into the recoverable class.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dabble/patches",
-  "version": "0.24.1",
+  "version": "0.24.2",
```
matalina, line 3: 0.24.1 was bumped in a standalone `chore(release): 0.24.1` commit; bundling the 0.24.2 bump into the fix PR diverges from that flow — double-check it won't collide with the release tooling's own bump. Also flagging that the unreleased #125 (LWW follower refresh) rides along in 0.24.2.

tests/client/OTAlgorithm-batchContiguity.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+      committed(createChange(100, 101, [{ op: 'add', path: '/a', value: 1 }])),
+      committed(createChange(101, 102, [{ op: 'add', path: '/b', value: 2 }])),
+      committed(createChange(102, 103, [{ op: 'add', path: '/c', value: 3 }])),
+    ]);
+    expect(doc.committedRev).toBe(103);
+    expect(doc.state).toEqual({ base: true, a: 1, b: 2, c: 3 });
+  });
+});
+
+describe('OTAlgorithm.applyServerChanges — interior-gap handling', () => {
```
matalina, line 46: Two nice-to-have cases (neither blocks):

- a root-replace catchup batch with an interior hole *after* the snapshot change — exercises the `if (!gap)` interior scan behind the `isRootReplaceCatchup` exemption;
- partial absorption: store at 103, follower doc at 100, batch `[101, 104]` — the frame re-check passes with `newC` non-empty, and the apply branch must still take the rebuild path.

matalina (commented):
Reviewed with a full trace of the receive-path branches against base. **The fix is correct and well-layered — LGTM.** Inline comments are all minor: a diagnostics nit on the interior-gap `MissingChangesError` params, two comment-wording notes, a dedup suggestion, and a release-flow question on the version bump.

What I verified:

- **Store-behind gapped delivery** throws `MissingChangesError` before the store write, and `PatchesSync` routes it to `syncDoc`/`getChangesSince` recovery.
- **The observed case** (complete store, gapped fan to an open follower) takes the rebuild path and converges with the skipped content present.
- **Partial absorption** (store rev at the hole's trailing edge, e.g. store at 150 receiving `[148, 151]`): the rebuilt frame passes with `newC` non-empty, the store write stays dense (only revs past the store's edge are new), and the apply branch still routes to rebuild — the store can never be gapped by an absorbed frame.
- **Root-replace catchup** is unaffected: the interior scan starts at index 1, and `OTDoc.applyChanges` already rejected leading-edge jumps (catchup goes through `import`).
- **No false positives**: `getChangesSince` is a raw `listChanges({ startAfter })` over the dense committed log, and history scrubbing is playback, not deletion — no legitimate sparse batch reaches `applyServerChanges`. Leaving `applyCommittedChanges` (which can take genuinely sparse reconstruction inputs) for a caller analysis first is the right scoping.

Test coverage is good — the five tests pin all three layers plus contiguous regressions, and the observed-case test reproduces the production scenario end-to-end. Two nice-to-have cases inline; neither blocks.

src/client/OTAlgorithm.ts
RyanHavoc: Done in 23b20a2. `buildFrame` now captures `gapAt = { expected: newC[i-1].rev + 1, got: newC[i].rev }` on an interior hole and threads it into `MissingChangesError`, so an interior gap reports the real boundary instead of `(committedRev + 1, committedRev + 1)`. `sinceRev` stays `committedRev` since that is what recovery consumes — only the diagnostic pair changes.

src/client/OTAlgorithm.ts
RyanHavoc: Done in 23b20a2. Reworded to spell out the trailing-edge case: store at the hole's trailing edge (e.g. store 150 for `[148, 151]`) rebuilds with a non-empty `newC = [151]`, leading edge clean off the store rev, yet the original batch is still non-contiguous. The check already scanned the real `serverChanges` array; this is comment-only so a future reader does not narrow it to the emptied case.

src/client/OTAlgorithm.ts
RyanHavoc: Done in 23b20a2. Extracted `firstGapIndex(changes)` and pointed both in-file scans (buildFrame's interior check and the apply-branch selector) at it. Left OTDoc.applyChanges' copy inline on purpose and added a note there that it is a deliberately-independent last-line invariant, per your point.

src/client/OTDoc.ts
RyanHavoc: Done in 23b20a2. Added the note: intentionally a plain `Error`, not `MissingChangesError`, because it signals a caller-bug invariant (the algorithm layer must never hand this method a gapped committed batch), not a recoverable transport gap — so it must not route through PatchesSync's `getChangesSince` recovery and mask the bug as a transient hole.

tests/client/OTAlgorithm-batchContiguity.spec.ts
RyanHavoc: Both added in 23b20a2: `catches an interior hole behind a root-replace catchup leading edge` (exercises the `if (!gap)` interior scan behind the `isRootReplaceCatchup` exemption) and `rebuilds an open doc when the store re-anchor absorbs a partial-edge gap (non-empty newC)` (store at 103, follower at 100, batch `[101, 104]` — frame re-check passes with non-empty `newC`, apply branch takes the rebuild). Full suite 3330 pass.

package.json
RyanHavoc: Checked — this matches the #125 flow rather than diverging from it. #125's own `chore(release): 0.24.1` (3970905) was a standalone bump commit carried *in* its fix branch and merged along with the fixes; this PR does the same (`4996384 chore(release): 0.24.2` sitting on top of the fix commit). There is no automated release tooling to collide with — patches has no release-please config and no publish workflow, so every `chore(release)` bump is manual and in-branch.

On the ride-along: 0.24.1 is now published on npm and `main` is at 0.24.1, so #125 has shipped. This branch is based on that `main`, so 0.24.2 carries only this PR's fix — #125 no longer rides along.

RyanHavoc:
Review round addressed in 23b20a2 (all checks green: 3330 tests pass, build + prettier + eslint clean):

- **Shared scan** — extracted `firstGapIndex(changes)`; both in-file scans use it, OTDoc keeps its independent last-line copy (with a note on why).
- **Gap diagnostics** — interior gaps now report the real hole boundary via `gapAt`; `sinceRev` unchanged.
- **Absorption comment** — reworded to cover the non-empty-`newC` trailing-edge case.
- **Plain `Error` rationale** — noted on OTDoc.applyChanges (caller-bug invariant, must not route through `getChangesSince` recovery).
- **Tests** — added both suggested cases (interior hole behind a root-replace catchup; partial-edge store re-anchor with non-empty newC).
- **Version bump** — this matches the #125 flow (standalone in-branch `chore(release)` commit); no release tooling to collide with, and 0.24.1 is now published so 0.24.2 carries only this fix. Details in-thread.

PR #128, opened 2026-07-22
DAB-834: don't abort queued-healthy IndexedDB transactions

jacwright:
## TL;DR

Some users hit a burst of false "couldn't save" errors that then snowballed into a retry storm (one user generated ~600 of them in under three minutes). Their writes were actually fine, our safety timer was just too aggressive: when a lot of saves happen at once the browser makes them wait in line, and we were treating "waiting in line" as "storage is broken" and cancelling a perfectly good save. This change makes the timer smart enough to tell a busy-but-healthy database apart from a genuinely stuck one, so those false failures (and the storms they triggered) stop. It also stops backgrounded browser tabs from wrongly reporting stalls after the OS freezes them.

Feeds the DAB-830 RCA.

## The problem

`IndexedDBStore` has a storage guard (soft 1s warn, hard 4s reject) that converts a genuinely never-settling IndexedDB operation into a typed `StorageTimeoutError`. Two defects made it fire on healthy operations:

1. **Queue congestion read as a stall.** The hard deadline is measured from `IDBTransactionWrapper` construction, which is queue *entry*, not transaction *start*. IndexedDB serializes overlapping-scope readwrite transactions, and every save path includes the `docs` store, so under a write burst a healthy transaction can sit queued for >4s with zero of its own requests settling. The guard only extends on that transaction's *own* request settles, which a queued transaction never gets, so it rejected **and** `tx.abort()`ed a write that was about to run. Sentry: ~1,500 events / 33 users in 4.5h on 3.0.47, elapsed tightly clustered 4000-4150ms (timer fired on time → main thread responsive → congestion, not disk failure), with self-sustaining retry storms.

2. **Suspended tabs fire the timer grossly late.** A backgrounded/throttled tab freezes the event loop; on resume the timer dispatches far past its budget and misreports (Sentry issue with elapsed up to 20 minutes, a pure background artifact).

## The fix

**Connection-level activity beacon.** `IndexedDBStore` owns a small mutable `{ lastSettleAt }` beacon shared with every `IDBTransactionWrapper` it creates (new optional constructor param; direct construction still works with a private fallback beacon). It's bumped on every observable sign of storage life across the whole connection: each request settle (success and error) via the store wrappers, and each transaction `oncomplete`/`onerror`/`onabort`.

**Gate the hard fire on connection silence.** In the transaction guard's hard-timeout callback, if the beacon shows activity within the last `storageTimeoutMs`, the backend is alive and this transaction is merely queued, so it extends for the remaining silence window instead of firing. It only rejects + aborts (exactly as before) once the entire connection has been silent for the full window. The extension is unbounded while the queue drains, mirroring the existing extend-on-own-settle design.

**Late-fire re-arm.** `armStorageGuard` now records each hard timer's scheduled time and budget. If the timer fires later than ~2x its budget (event loop was suspended/throttled), it re-arms once with a fresh full window rather than firing, giving queued IDB events a chance to dispatch first; if the next window also passes silent, it fires normally. Applies to both the `open` and `transaction` guards.

Soft warn, error-message format, open-path re-arm behavior, and `tx.abort()` on genuine fires are all unchanged. No public API changes beyond internal wiring.

## Tests

Added to `tests/client/IndexedDBStore-timeout.spec.ts`:
- A queued transaction with no own settles defers while a sibling transaction keeps settling requests (beacon activity), and does time out once the whole connection goes silent for the window.
- A hard timer that fires grossly late (wall clock decoupled from the fake timer clock) re-arms once instead of firing, then fires when the re-armed window is still silent — at both the transaction and open guard levels.
- Existing coverage unchanged: a genuinely silent transaction still rejects with `StorageTimeoutError` and aborts.

Full suite green: 3333 passed / 4 skipped. `type:check`, `lint`, `format:check` all pass.

### DIFF
--- a/src/client/IndexedDBStore.ts
+++ b/src/client/IndexedDBStore.ts
@@ -32,6 +32,20 @@ export const DEFAULT_STORAGE_TIMEOUT_MS = 4000;
  * real stall. See the late-fire re-arm in {@link armStorageGuard}.
  */
 const LATE_FIRE_LATENESS_FACTOR = 2;
+/**
+ * Absolute backstop, in hard-timeout windows, on how long one transaction may keep deferring purely
+ * on OTHER transactions' activity (see {@link StorageActivityBeacon}) without ever settling one of
+ * its OWN requests. The connection-silence gate is unbounded by design — correct for a transaction
+ * merely queued behind a healthy write burst — but from inside the wrapper that case is
+ * indistinguishable from a transaction wedged at the HEAD of the queue while unrelated-scope
+ * transactions keep the connection observably alive; the latter would otherwise never settle its
+ * caller. Past this many windows without own progress it fires anyway, bounding that (exotic) wedge
+ * to a finite wait. Any own request settle resets the count, so a legitimately long-running bulk
+ * transaction that keeps settling requests is never capped. Deliberately generous: DAB-834's real
+ * queue-drain waits clustered at ~1 window, so a healthy queue wait never reaches this, while a
+ * false fire here is the very retry-storm class the beacon exists to prevent.
+ */
+const MAX_DEFER_WINDOWS = 30;
 
 /** Passed to `onSlowStorage` when an IndexedDB operation crosses the soft threshold without settling. */
 export interface SlowStorageInfo {
@@ -927,6 +941,11 @@ export class IDBTransactionWrapper {
     const beaconRef = this.beacon;
     const storeNames = () => Array.from(tx.objectStoreNames ?? []);
     const timeoutMs = options?.storageTimeoutMs ?? DEFAULT_STORAGE_TIMEOUT_MS;
+    // How many consecutive hard windows this transaction has ridden on connection-wide activity
+    // without settling a single one of its OWN requests. Reset by every own settle (see
+    // `extendDeadline`); bounds the beacon defer so a wedged-at-head transaction can't ride
+    // healthy sibling traffic forever. See {@link MAX_DEFER_WINDOWS}.
+    let deferWindows = 0;
     let guard!: StorageGuard;
     this.promise = new Promise((resolve, reject) => {
       // Max-time guard: some machines have transactions that never fire ANY of the settle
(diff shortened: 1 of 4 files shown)


src/client/IndexedDBStore.ts
```diff
@@ -857,16 +916,36 @@ export class IDBTransactionWrapper {
...
+        // settling. Before treating that as a stall, consult the connection-wide beacon — if ANY
+        // transaction settled a request within the window, the backend is alive and this one is
+        // merely queued, so extend for the remaining silence and let the queue drain (unbounded,
+        // mirroring extend-on-own-settle). Only fire once the ENTIRE connection has been silent
+        // for the full window.
+        const silentFor = Date.now() - beaconRef.lastSettleAt;
+        if (silentFor < timeoutMs) {
+          guard.extend(timeoutMs - silentFor);
+          return;
+        }
```
RyanHavoc, line 968: Residual behavior worth a note (not blocking): now that the hard timer only fires once the whole connection has been silent for `storageTimeoutMs`, a transaction genuinely wedged at the IDB level — but whose connection stays active via non-overlapping transactions (e.g. readonly `snapshots`/`branches` settling every <4s keeps `lastSettleAt` fresh) — defers indefinitely and never rejects its `complete()` caller, where pre-change it would have timed out + `tx.abort()`ed at 4s.

This reads as the intended tradeoff (fire "only once the ENTIRE connection has been silent"), and the trigger is exotic since a real backend stall usually wedges the whole connection — so no change needed. Flagging only so a future "stuck writer, no timeout" report has a breadcrumb. If a backstop ever feels warranted, an absolute cap on total defer time (distinct from the per-silence-window extend) would bound the worst case without reintroducing the false-positive.

RyanHavoc:
Pushed d3b00aa addressing the residual I flagged on the beacon defer.

The connection-silence gate is unbounded, which is right for a transaction queued behind a healthy write burst — but from inside the wrapper that's indistinguishable from a transaction wedged at the head of the queue while unrelated-scope transactions keep the connection observably alive, which would then never settle its caller. Added a finite backstop: a transaction may ride connection activity for at most `MAX_DEFER_WINDOWS` (30) hard windows without settling any of its **own** requests, after which it fires as if silent. Any own settle resets the count, so a legitimately long-running bulk transaction that keeps settling requests is never capped, and the threshold is deliberately generous (the DAB-834 queue waits clustered at ~1 window) so no healthy queue wait reaches it — a false fire here would be the exact retry-storm class the beacon exists to prevent.

Added a test alongside it: a transaction riding perpetual connection activity with no own settles defers past the first window (beacon defer intact) but still fires at the backstop instead of hanging forever. Full suite green (3334 passed / 4 skipped); `type:check`, `lint`, `format:check` all pass.

This is a judgment call over the intentional unbounded design — if you'd rather keep it purely unbounded and lean on the app-layer write watchdog for the wedged-at-head case, it's a clean revert of just this commit.

PR #129, opened 2026-07-22
DAB-835: three-bucket change-failure classes — defective changes terminal, ambiguous retries bounded

jacwright:
## TL;DR

When Dabble saves your typing, a save can fail. Today two very different failures are handled the same wrong way: a change the app can *never* save (a subtle bug producing un-storable data) retries forever and, worse, permanently jams every later save for that document for the rest of the session — so the rest of what you type only lives in memory and quietly vanishes on reload, with no warning. And when your device's storage is genuinely broken, saves also retry silently forever while you keep typing words that are never persisted.

This change makes both fail *loudly and safely*:

- A change that can never be saved now stops immediately and tells the app, so the app can alert you instead of silently blocking all future saves.
- Broken storage now stops silently retrying after a few tries and tells the app that saving is stuck, so the app can warn you and offer a "try again" — and your typed text is kept (not thrown away) so nothing is lost when saving resumes.

No visible behavior changes when saving is working normally. The user-facing alerting is a follow-up (DAB-836, app side); this PR is the library half that makes it possible.

## Technical detail

`Patches._processDocChange` classified change-submit failures into two buckets: authoritative rejection (rollback) and everything-else (silent re-submit with backoff, forever). Two production consequences motivated this: a `DataCloneError` from a non-cloneable value can never persist yet retried forever and permanently blocked the doc's change queue (Sentry DABBLE-WRITER-3-PQ); and a genuinely broken storage environment retried silently forever while the user kept typing memory-only words (the DAB-830 loss pattern).

### Three failure classes (`_processDocChange`)

1. **`rejection`** — terminal `StatusError` (401/402/403/404/410). Unchanged behavior (epoch bump + `rollbackOptimistic`), now tagged `kind: 'rejection'`.
2. **`defective`** (new, terminal) — `isDefectiveChangeError`: `DataCloneError` / `DataError` / `ConstraintError`. The change's own data can never be saved — a bug in the producing code, so retrying is provably useless. `console.error` + epoch bump + `rollbackOptimistic` + `onError { kind: 'defective', willRetry: false }`. No retry.
3. **`environment`** — ambiguous transport/storage failure. **Bounded** retry: `MAX_CHANGE_SUBMIT_ATTEMPTS = 3` (initial + 2), existing backoff and mid-retry guards (closed / rebased-to-empty / confirmed-through-other-path) preserved. On exhaustion the ops are **kept** (never discarded — they may have landed or may yet land) and the doc's write path is **latched**; `onError { kind: 'environment', willRetry: false, attempt }`.

### Write latch + recovery API

- `_writeLatches: Map<docId, Error>`. While latched, `_handleDocChange` skips persistence and emits `onError { kind: 'environment', latched: true }`; ops stay applied optimistically in memory (text stays visible, nothing minted or sent).
- `retrySavingChanges(docId?)` clears the latch(es) and re-drives the doc's retained optimistic entries through the normal submit path in capture order (drains the per-doc queue to a fixed point with the latch still set first, so a change captured while latched can't process the same entry out from under the re-drive). Re-submits go through the same three-bucket logic and may re-latch.
- Read-only latch state: `isWriteLatched(docId)`, `writeLatchedDocs`. No UX policy in the library (patches scope boundary).

### Types + docs

- `ChangeErrorContext` (exported) extends the `onError` payload with `kind?: 'rejection' | 'defective' | 'environment'` and `latched?: boolean` (backward compatible).
- `src/net/error.ts`: `isDefectiveChangeError`, duck-typed on `name` without an `instanceof Error` gate (cross-realm/DOMException, per `isStorageError`). `DataError`/`ConstraintError` are deliberately disjoint from `STORAGE_FAULT_NAMES` (malformed record vs environment failing to save a well-formed one).
- `src/client/BaseDoc.ts`: `_getOptimisticEntries()` accessor for the re-drive.
- `docs/Patches.md`: failure-class + `retrySavingChanges` section.

### Tests

Defective change (Error subclass **and** non-Error shape) → rollback, epoch bump, no retry, `kind: 'defective'`; ambiguous → exactly 3 attempts with backoff then latch, ops retained; latched doc skips persistence but keeps ops + emits `latched: true`; `retrySavingChanges` clears latch and re-submits in order on recovery, re-latches on repeated failure; rejection path (now `kind: 'rejection'`); confirmed-through-other-path and rebased-to-empty guards within the bounded loop; `isDefectiveChangeError` classification. Full suite: 3341 passed / 4 skipped. `type:check`, `lint`, `format:check` all green.

Refs DAB-835, DAB-836.

### DIFF
--- a/src/client/Patches.ts
+++ b/src/client/Patches.ts
@@ -128,6 +128,16 @@ export class Patches {
    * {@link retrySavingChanges}, which re-drives the retained optimistic ops through the submit path.
    */
   private _writeLatches: Map<string, Error> = new Map();
+  /**
+   * Per optimistic-entry stable change id (entry ops array → id). A change re-driven by
+   * {@link retrySavingChanges} after its write path was latched re-mints under the SAME id its
+   * failed submit used, so the server's id-based commit dedup still collapses the ambiguous case
+   * where a latched submit had actually landed — a fresh id would double-commit it. Keyed by the
+   * entry's ops array (stable by reference across in-place rebases — the same ref
+   * {@link BaseDoc._getOptimisticEntries} hands back), so it self-clears via GC once the entry is
+   * confirmed and dropped; no explicit teardown needed.
+   */
+  private _changeStableIds: WeakMap<JSONPatchOp[], string> = new WeakMap();
   /** Doc IDs whose openDoc() body is currently in flight (between createDoc and this.docs.set). */
   private _openingDocs: Set<string> = new Set();
   /** Single-slot, latest-wins snapshot stash for docs in `_openingDocs`. Drained on open completion. */
@@ -525,6 +535,13 @@ export class Patches {
       await drain;
       if (this._changeQueues.get(docId) === drain) this._changeQueues.delete(docId);
     }
+    // Drop any write latch with the doc. Done AFTER the drain so changes still queued during
+    // teardown keep hitting the latch's skip-and-return (rather than persisting against a doc being
+    // closed); once drained, the latch's retained optimistic ops live only in this in-memory doc
+    // (never persisted), so they go with it. A later openDoc must then start unlatched and persist
+    // normally instead of inheriting a stale latch that would silently skip every save. (close()
+    // clears all latches; this is the per-doc closeDoc / untrackDocs / deleteDoc teardown path.)
+    this._writeLatches.delete(docId);
     if (untrack) {
       await this.untrackDocs([docId]);
     }
@@ -779,8 +796,15 @@ export class Patches {
     algorithm: ClientAlgorithm,
     metadata: Record<string, any>
   ): Promise<void> {
-    // Minted once per captured change; stable across every retry of this submit.
-    const stableId = createId(STABLE_CHANGE_ID_LENGTH);
+    // Stable across every retry of this submit AND across a later retrySavingChanges re-drive of
+    // the same optimistic entry: keyed by the entry's ops array so the id survives the latch,
+    // keeping the server's commit dedup effective if a latched submit had actually landed (see
+    // {@link _changeStableIds}). A brand-new entry (including one typed while latched) mints fresh.
+    let stableId = this._changeStableIds.get(ops);
+    if (!stableId) {
+      stableId = createId(STABLE_CHANGE_ID_LENGTH);
+      this._changeStableIds.set(ops, stableId);
+    }
     const baseDoc = doc as unknown as BaseDoc<T> | undefined;
     for (let attempt = 0; ; attempt++) {
       try {
@@ -806,6 +830,9 @@ export class Patches {
 
         // 3) Ambiguous / environment failure — not a verdict on the ops. Retry bounded, then latch.
         if (attempt + 1 >= MAX_CHANGE_SUBMIT_ATTEMPTS) {
+          // Closing: don't repopulate the latch map close() just cleared — the instance and its
+          // optimistic ops are being discarded, and close() already parked the retry loops.
+          if (this._closed) return;
           // Exhausted: do NOT roll back (the ops may have landed or may yet land — discarding
           // would delete un-persisted work). Latch the write path so no further change on this
           // doc is minted/sent until the app retries; the ops stay applied optimistically.
--- a/src/client/Patches.ts
+++ b/src/client/Patches.ts
@@ -1,7 +1,7 @@
 import { createId } from 'crypto-id';
 import { type Unsubscriber, signal } from 'easy-signal';
 import type { JSONPatchOp } from '../json-patch/types.js';
-import { isRejectionError } from '../net/error.js';
+import { isDefectiveChangeError, isRejectionError } from '../net/error.js';
 import type { Change, PatchesSnapshot, QuarantinedChange } from '../types.js';
 import { singleInvocation } from '../utils/concurrency.js';
 import type { BaseDoc } from './BaseDoc.js';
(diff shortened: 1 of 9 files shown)


src/client/Patches.ts
```diff
@@ -81,6 +121,13 @@ export class Patches {
    * so their processing is skipped when their captured epoch is stale.
    */
   private _changeEpochs: Map<string, number> = new Map();
+  /**
+   * Docs whose write path is latched (docId → the error that exhausted its submit retries).
+   * While latched, changes are NOT persisted — the ops stay applied optimistically in memory so
+   * the user's text stays visible, but nothing is minted or sent. Cleared by
+   * {@link retrySavingChanges}, which re-drives the retained optimistic ops through the submit path.
+   */
+  private _writeLatches: Map<string, Error> = new Map();
```
RyanHavoc, line 130: **Latch leak in `closeDoc` → a reopened doc silently stops persisting.** This comment says the latch is cleared by `retrySavingChanges` (and `close()` clears it at L658), but `closeDoc()` doesn't: it deletes from `docs` and `_changeQueues` yet leaves `_writeLatches` intact — https://github.com/dabblewriter/patches/blob/8d98d7253ae5415324be41564c461798ab307306/src/client/Patches.ts#L517-L527 . `untrackDocs`/`deleteDoc` route through `closeDoc` and inherit it.

So a doc that latches on an environment exhaustion, is then closed (refCount→0), and later reopened rebuilds a fresh doc from the store snapshot but keeps the stale latch. The first edit on the reopened doc hits the latch check in `_handleDocChange` (L717) and returns without persisting — every edit is memory-only and lost on reload, with no new latch signal, and `isWriteLatched(id)` is `true` immediately after `openDoc` (which never clears the latch). Suggest deleting the docId from `_writeLatches` in `closeDoc` teardown, plus a close→reopen-while-latched test.

src/client/Patches.ts
```diff
@@ -736,6 +835,66 @@ export class Patches {
...
+    }
+    if (this._closed || !this._writeLatches.has(docId)) return;
+    this._writeLatches.delete(docId);
+    // Re-drive the retained optimistic entries (snapshot the queue; the same inner op arrays the
+    // live queue holds, so confirmation shifts line up). Enqueued synchronously here so they take
+    // their FIFO slots before any concurrent change() can, preserving capture order.
+    const baseDoc = managed.doc as unknown as BaseDoc<any>;
+    const entries = typeof baseDoc._getOptimisticEntries === 'function' ? baseDoc._getOptimisticEntries() : [];
+    await Promise.all(
+      entries.map(entryOps => this._handleDocChange(docId, entryOps, managed.doc, managed.algorithm, managed.metadata))
```
RyanHavoc, line 921: **Re-drive mints a new `stableId`, defeating the server dedup backstop across the latch boundary.** The bounded retry reuses one `stableId` per submit (https://github.com/dabblewriter/patches/blob/8d98d7253ae5415324be41564c461798ab307306/src/client/Patches.ts#L782-L784) so an ambiguously-landed write is deduped by the server's id-based commit dedup. But `stableId` is a local var, not stored on the optimistic entry, so this re-drive sends each retained entry through a fresh `_processDocChange` → a *new* stableId.

`savePendingChanges` keys only by `[docId, rev]` (no id dedup), so if a submit persisted-and-flushed but reported failure — the "write may have landed" case this feature targets — the re-drive re-mints the same ops under a new id and the server commits them twice (self-doubling). Not reachable with the shipped `OTIndexedDBStore` (atomic txn — a failed save persists nothing), but real for a non-atomic / worker-backed store where the ack is lost after commit. Suggest stashing the stableId on the optimistic entry and reusing it on re-drive so the invariant survives the latch.

src/client/Patches.ts
```diff
@@ -707,20 +788,38 @@ export class Patches {
...
+        // 3) Ambiguous / environment failure — not a verdict on the ops. Retry bounded, then latch.
+        if (attempt + 1 >= MAX_CHANGE_SUBMIT_ATTEMPTS) {
+          // Exhausted: do NOT roll back (the ops may have landed or may yet land — discarding
+          // would delete un-persisted work). Latch the write path so no further change on this
+          // doc is minted/sent until the app retries; the ops stay applied optimistically.
+          console.error(
+            `Doc change for ${docId} failed ${MAX_CHANGE_SUBMIT_ATTEMPTS} times; latching write path:`,
+            err
+          );
+          this._writeLatches.set(docId, err as Error);
```
RyanHavoc, line 843: Minor: unlike the non-exhaustion retry path (which checks `this._closed` after the backoff at L826), this exhaustion branch sets `_writeLatches` unconditionally, so a submit that fails during `close()` repopulates the map `close()` just cleared at L658. Benign (the instance is being discarded), but a `if (this._closed) return;` before the `set` keeps it clean.

src/net/error.ts
```diff
@@ -241,6 +241,43 @@ export function toStorageError(err: unknown): unknown {
...
+ *   value that isn't valid for the store's keyPath).
+ * - `ConstraintError` — the record structurally violates a uniqueness/keyPath constraint.
+ *
+ * These are deliberately NOT in {@link STORAGE_FAULT_NAMES}: a storage fault is the environment
+ * failing to save a WELL-FORMED record (retry may succeed once pressure lifts), whereas each of
+ * these means the record ITSELF is malformed. That is a bug in the code that produced the change,
+ * never a transient condition — which is exactly why the storage-fault doc-comment lists
+ * `ConstraintError`/`DataError` as programming errors it excludes.
+ */
+const DEFECTIVE_CHANGE_ERROR_NAMES: ReadonlySet<string> = new Set(['DataCloneError', 'DataError', 'ConstraintError']);
```
RyanHavoc, line 259: Classifying `ConstraintError`/`DataError` as terminal `defective` is consistent with the existing convention (both documented as programming errors, excluded from `STORAGE_FAULT_NAMES`) and safe for the shipped store — `OTIndexedDBStore.savePendingChanges` uses `put` and re-stamps rev in-transaction, so a transient `ConstraintError` can't arise.

One behavior change worth a doc note: a custom `OTClientStore` using `add` or a unique secondary index could emit a transient `ConstraintError`, which this now rolls back and discards where the pre-PR infinite-retry path retained it. Non-blocking — just a line in the failure-class docs.

RyanHavoc:
Pushed 8056fd8 addressing the review findings.

1. (high — latch leak) `closeDoc` / `untrackDocs` / `deleteDoc` left the doc in `_writeLatches`, so a latched doc that was closed and reopened inherited the stale latch and silently stopped persisting every edit. `closeDoc` now drops the latch in teardown, after the queue drain (so changes still queued during teardown keep hitting the skip-and-return rather than persisting against a closing doc). Added a close→reopen test.

2. (medium — dedup across the latch) `retrySavingChanges` re-drove a latched entry through a fresh `_processDocChange`, minting a new `stableId` and defeating the server's id-based commit dedup — a latched submit that had actually landed would double-commit under the new id. The stable id is now keyed by the optimistic entry's ops array (a `WeakMap`, so it self-clears via GC once the entry is confirmed), so a re-drive reuses the original id while a brand-new entry (including one typed while latched) still mints fresh. Added a test asserting the re-driven change keeps its original id.

3. (low — close race) the exhaustion/latch branch now returns early on `this._closed` before `_writeLatches.set`, so a submit failing during `close()` doesn't repopulate the map `close()` just cleared.

4. (doc) noted in `docs/Patches.md` that a custom store which can raise a transient `ConstraintError` must wrap it as a storage fault (environment bucket) rather than let it be discarded as defective.

Full suite green (3343 passed / 4 skipped); `type:check`, `lint`, `format:check` all clean.

PR #131, opened 2026-07-28
fix(net): decouple send/commit from SSE push-stream liveness (DAB-831)

matalina:
Part of [DAB-831](https://linear.app/dabblewriter/issue/DAB-831/patches-decouple-the-sendcommit-path-from-sse-push-stream-liveness-so) (full plan, line-verified review evidence, and resolved correctness questions live there). Customer context: [DAB-765](https://linear.app/dabblewriter/issue/DAB-765/30-sync-blocked-on-corporate-and-personal-networks-sync-moved-to-a-new) — 3 reporters whose networks (or on-device security software) block the SSE stream while every fetch succeeds.

## Problem

For the SSE+REST transport, `state.connected` only means "the EventSource push stream is open" — but every send path gated on it. A hostile middlebox or an antivirus HTTPS scanner that buffers `text/event-stream` keeps the stream down forever while plain requests work fine, so users could not save even though every `commitChanges` POST would have landed. Worse, the only cold-start flush trigger (`syncAllKnownDocs`) runs solely on the `'connected'` event — so for the never-connects cohort the pending backlog never drained at all.

## Change

- **`PatchesConnection.sendRequiresStream`** (required capability): `true` on `PatchesWebSocket` (all traffic rides the socket), `false` on `PatchesREST` (commits/reads are independent fetches).
- **`_canSend()`** — `!offline && (connected || !sendRequiresStream)` — replaces `_isConnectedAndOnline()` and re-gates every send path: `_handleDocChange`, `syncDoc`, `flushDoc` (both gates), the branch-meta loops, server deletes, and newly-tracked-doc sync. `syncAllKnownDocs` and its `'connected'` trigger stay stream-bound (they own the re-subscribe). WebSocket behavior is unchanged.
- **Degraded-mode pass (`_syncAllDegraded`)**: while online with the stream down on REST, flush pending + `getChangesSince` catch-up for every tracked doc via `syncDoc` (inheriting the per-doc serial gate, reload-buffer guard, and retry ladders). Runs immediately on connect failure and on the online transition, repeats on a jittered ~24-30s timer, and cancels the moment the stream connects. Skips subscribe (the server rejects subscriptions with no stream).
- **Reclassification**: a failed `connect()` on REST no longer latches `syncStatus: 'error'` — a stream-only failure is not a write failure. The retry/reprobe ladders key on `_canSend`, so send failures keep retrying while degraded instead of deferring to a reconnect that may never come.

Effect on the DAB-765 cohorts: the **flapping** case (stream resets every ~90s) flushes keystrokes over POST during the down-windows; the **never-connects** case (client-side SSE buffering) drains its backlog and receives catch-up without a `'connected'` event ever firing.

## Testing

- 11 new specs: flush succeeds stream-down-but-online on REST; still blocks on WS; keystroke reaches flush through the outer gates; **never-connects backlog drain** (failed `connect()` → no error posture + pending flushes with no connected event); degraded pass arms jittered, pulls catch-up, re-arms, and hands off on reconnect; clean pass reports global `synced`; declines offline / never engages on WS; server delete goes out while degraded.
- `PatchesREST.spec`: `sendRequiresStream === false`; commit POSTs without the stream ever opening.
- Full gate: `type:check` / `lint` / `format:check` / 3,330 tests passing. (The 3 `tests/solid` suite failures are a pre-existing `@solid-refresh` environment issue — they fail identically on untouched `main`.)

## Review fixes (2026-07-28, commit aca1ca1)

1. **(Bug) `disconnect()`/`destroy()` now actually stop the degraded machinery.** New `_started` intent flag (set by `connect()`, cleared by `disconnect()`), consulted by the **send-independent leg** of `_canSend()` only — the WebSocket leg still keys on `connected` alone, so WS semantics are untouched. One flag fixes all three symptoms: the in-flight pass's `finally` can no longer re-arm after teardown (no stale-auth POSTs after logout), an explicit disconnect parks keystrokes again like the pre-PR gate did, and the online-transition trigger can't fire on a never-connected instance. Regression spec tears the instance down mid-pass and asserts no re-arm, no send, and that `connect()` re-arms.
2. **`_syncAllDegraded` has a `catch`** (mirrors `syncAllKnownDocs`): a branch-store read rejection no longer escapes the `void` call sites as an unhandled rejection. The hang-liveness dependency (an IDB stall keeps `finally` from re-arming until the next connection/online event) is documented in a comment.
3. **Conscious decision on the ~30s wipe cadence: skip it on REST stream drops.** The transport retries the stream forever with ~30s backoff and this branch fires on every failed attempt — wiping retry ladders there also wiped the error-surfacing dedup (a latched 403 re-emitted to console/telemetry every cycle: a sustained Sentry drip) and flapped global status unsynced↔synced on that cadence. The "reconnect will resync everything" rationale doesn't hold on a transport whose reconnect isn't coming, so REST drops now preserve retry state, dedup, and `syncStatus`; the wipe moved to the **connected transition** (a fresh surfacing session right before `syncAllKnownDocs` re-attempts every doc). WS drop behavior is byte-for-byte unchanged. This also removes the status oscillation the earlier body called out, so the dw3 companion has no transition to mis-key on.
4. **Conservative duck-typing polarity:** permissive branches require `sendRequiresStream === false`; an un-migrated JS implementer gets the old stream-bound behavior.
5. **Test hygiene:** `makeRestSync` builds a fresh connection object (no spies shared with the WS mock) and arms the intent via `connect()`; new specs cover the disconnect race and the no-wipe cadence (755 net tests).
6. **Docs:** "Degraded Mode" section in `docs/PatchesSync.md` + pointer in `docs/sse-rest.md`.

**Review finding 4 — now also fixed here (commit 45beab6). Fixes [DAB-865](https://linear.app/dabblewriter/issue/DAB-865/patches-resume-pass-never-subscribes-docs-tracked-while-the-stream-was):** the resume pass now tracks the ids this instance actually subscribed (`_subscribedIds`) and additionally subscribes tracked docs missing from that set — a doc tracked while the stream was down was never subscribed by anyone and silently missed live pushes until the next cold connect. A successor tab (empty set) re-subscribes everything on resume: one idempotent POST buys correctness, and the hand-off round-trip saving remains for docs the instance knows it subscribed. Spec covers both the surgical case and the successor case.

Also fixed the message nit: `flushDoc`'s `NetworkError`s now say "offline or not connected" instead of claiming "not connected" while merely offline.

**Release note:** publish as **0.26.0** — the new required member on the exported `PatchesConnection` interface is breaking for external implementers.

## Review notes

- The shared test mock defaults `sendRequiresStream: true` (WS-like) so existing specs keep their historical gating semantics; REST-behavior specs opt in with `false`.
- Global `syncStatus` on REST is preserved across stream drops (see review fix 3); it starts 'unsynced' and turns 'synced' on the first clean degraded pass.
- Companion PR in dabble-writer-3.0 (banner/indicator rewire) must deploy only after this ships in a published `@dabble/patches` and dw3's pin is bumped.

### DIFF
--- a/src/net/PatchesSync.ts
+++ b/src/net/PatchesSync.ts
@@ -96,6 +96,14 @@ function jitterReprobeDelay(delayMs: number): number {
 }
 
 /**
+ * Interval between degraded-mode catch-up passes while online with the push stream down
+ * on a send-independent transport (see `_syncAllDegraded`). Jittered via
+ * `jitterReprobeDelay` (~24-30s) so a fleet of stream-blocked clients doesn't sweep the
+ * server in lockstep. DAB-831.
+ */
+const DEGRADED_SYNC_INTERVAL_MS = 30_000;
+
+/**
  * Handles server connection, document subscriptions, and syncing logic between
  * the Patches instance and the server.
  *
@@ -170,7 +178,17 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
 
     // --- Event Listeners ---
     this._unsubs = [
-      onlineState.onOnlineChange(online => this.updateState({ online })),
+      onlineState.onOnlineChange(online => {
+        this.updateState({ online });
+        // Coming online with the stream down: on a send-independent transport don't
+        // wait for the stream to establish (a hostile network may hold it down forever
+        // — DAB-831); drain pending and pull catch-up over plain requests right away.
+        if (online && !this.state.connected && !this.connection.sendRequiresStream) {
+          void this._syncAllDegraded();
+        } else if (!online) {
+          this._clearDegradedSync();
+        }
+      }),
       this.connection.onStateChange(this._handleConnectionChange.bind(this)),
       this.connection.onChangesCommitted(this._receiveCommittedChanges.bind(this)),
       this.connection.onDocDeleted(docId => this._handleRemoteDocDeleted(docId)),
@@ -195,6 +213,8 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
   private _reloadBuffers = new Map<string, Change[]>();
   /** Pending per-doc slow re-probe timers for docs the retry ladder gave up on (see `_scheduleSyncReprobe`). */
   private _syncReprobeTimers = new Map<string, ReturnType<typeof globalThis.setTimeout>>();
+  /** Pending timer for the next degraded-mode catch-up pass (see `_syncAllDegraded`); null when idle. */
+  private _degradedSyncTimer: ReturnType<typeof globalThis.setTimeout> | null = null;
   /**
    * Docs whose current sync failure has already been surfaced (console + `onError`).
    * A background re-probe re-enters `syncDoc` every few minutes; without this a
@@ -285,7 +305,17 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
     } catch (err) {
       console.error('PatchesSync connection failed:', err);
       const error = err instanceof Error ? err : new Error(String(err));
-      this.updateState({ connected: false, syncStatus: 'error', syncError: error });
+      if (this.connection.sendRequiresStream) {
+        this.updateState({ connected: false, syncStatus: 'error', syncError: error });
+      } else {
+        // A stream-only failure isn't a sync failure on a send-independent transport:
+        // every commit and read still works over plain requests (DAB-831). Keep the
+        // error off the sync posture, start the degraded catch-up/flush pass now (the
+        // pending backlog must not wait for a `connected` event that may never come),
+        // and let the transport keep retrying the stream with its own backoff.
+        this.updateState({ connected: false });
+        void this._syncAllDegraded();
+      }
       this.onError.emit(error);
       throw err;
     }
@@ -298,6 +328,7 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
     this.connection.disconnect();
     this.updateState({ connected: false, syncStatus: 'unsynced' });
     this._clearAllSyncRetries();
+    this._clearDegradedSync();
     this._resetSyncingStatuses();
   }
 
@@ -335,7 +366,7 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
     const deletes = pendingBranches.filter(b => b.pendingOp === 'delete');
 
     for (const branch of creates) {
-      if (!this.state.connected) break;
+      if (!this._canSend()) break;
 
       try {
         // Create the branch on the server. The server is idempotent when metadata.id is provided.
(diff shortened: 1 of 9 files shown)


src/net/PatchesSync.ts
```diff
@@ -504,17 +565,23 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
...
+      // the round-trips a hand-off is meant to avoid.
+      const subscribeCandidates = resume
+        ? syncIds.filter(id => syncedEntries[id]?.hasPending || !this._subscribedIds.has(id))
+        : syncIds;
       if (subscribeCandidates.length > 0) {
         try {
           const subscribeIds = this._filterSubscribeIds(subscribeCandidates);
           if (subscribeIds.length) {
             await this.connection.subscribe(subscribeIds);
+            subscribeIds.forEach(id => this._subscribedIds.add(id));
```
RyanHavoc: `subscribe()` resolves with the ids the client is *now successfully subscribed to* — a possibly-partial subset of the request ([contract](https://github.com/dabblewriter/patches/blob/45beab6158f7762869e5e214f0087454c0bd2d95/src/net/protocol/types.ts#L165-L170)). `PatchesREST.subscribe` resolves with `requested.filter(id => registered.has(id))` without throwing when ids are denied or left unaccounted ([PatchesREST.ts#L479-L481](https://github.com/dabblewriter/patches/blob/45beab6158f7762869e5e214f0087454c0bd2d95/src/net/rest/PatchesREST.ts#L479-L481)). Here — and at the same pattern in the newly-tracked-docs handler ([PatchesSync.ts#L1438-L1441](https://github.com/dabblewriter/patches/blob/45beab6158f7762869e5e214f0087454c0bd2d95/src/net/PatchesSync.ts#L1438-L1441)) — the resolved value is discarded and every *requested* id is added to `_subscribedIds`.

A denied or unregistered doc then satisfies `_subscribedIds.has(id)` in the resume filter ([PatchesSync.ts#L576-L578](https://github.com/dabblewriter/patches/blob/45beab6158f7762869e5e214f0087454c0bd2d95/src/net/PatchesSync.ts#L576-L578)), so it is permanently skipped from re-subscribe — the silent-miss DAB-865 fixes, reintroduced for the partial-response case.

Suggest using the resolved value at both sites:

```ts
const subscribed = await this.connection.subscribe(subscribeIds);
subscribed.forEach(id => this._subscribedIds.add(id));
```

plus a spec where the mock resolves with a proper subset — the shared `connectionMock` and the inline mocks all echo the full request, so this path is currently untested.

matalina: Fixed in dad52c3 — both sites now record the resolved grant (`(await subscribe(...)) ?? []`) instead of the request, and a new spec covers a proper-subset response plus the follow-up resume re-attempting the ungranted doc.

RyanHavoc (approved):

PR #132, opened 2026-07-28
fix(ot): refuse changes the splitter cannot get under the store's limit

jacwright:
## TL;DR

A user whose single edit was too large to store could get silently stuck forever with nothing saving. The app now shrinks its batches and retries automatically when the server says a change is too big, and refuses outright only when an edit genuinely cannot be divided at all. When it does refuse, it says so loudly instead of quietly wedging the queue.

## The bug

`breakSingleChange` has four escape hatches for an operation it has no seam to cut on. Each one emitted the oversized op anyway behind a `console.warn` that nothing listens to. Once such a change is persisted to IndexedDB, every retry re-splits it to the same oversized bytes, the store refuses it again, and because the piece count never changes, `flushDoc`'s `replacePendingChanges` path never fires either. One real user went 13 days with zero saves this way.

There is a fifth defect in the same file: `splitLargeInsertText` derives a character budget from a *compressed byte* budget and then pushes chunks with no size re-check. High-entropy text blows straight through the budget: a 1M-character insert against a 300KB budget produced pieces of 1.9MB.

## What changed

**`maxUnsplittableBytes`, a second threshold distinct from the split target.** Splitting still targets `maxStorageBytes` exactly as before. An unsplittable op over that budget but at or under `maxUnsplittableBytes` is still emitted, as today. One past `maxUnsplittableBytes` throws `UnsplittableChangeError`.

The two are deliberately separate. `maxStorageBytes` is a conservative target, and with `compressedSizeUint8` it measures `nonOpsSize + compressedOps` while the backend's real cap applies to `compressedOps` alone. Refusing at the split target would newly reject changes that store fine today.

**`UnsplittableChangeError`** lives in `src/net/error.ts` and is named into `DEFECTIVE_CHANGE_ERROR_NAMES`, so `Patches._processDocChange` classifies it as `kind: 'defective'`: roll the change back, surface it, never retry. It carries `op`, `path`, `bytes`, `maxBytes`, `docId` and `changeId`.

**`splitLargeInsertText` re-measures.** Each chunk is measured with the caller's size calculator and bisected until it fits; the character budget is now only a seed. Cuts avoid surrogate pairs.

**Resplit on 413.** When a flush fails with a `StatusError` code 413 and `data.scope === 'change'` (the shape pup PR #223 sends), `PatchesSync` halves that doc's effective `maxStorageBytes` and re-flushes, floored at 100,000 (four halvings from a 900KB budget). The budget resets once a flush lands. At the floor it stops halving and the doc latches with the error rather than looping. It only halves a budget the consumer configured; with no `maxStorageBytes` nothing was splitting in the first place and inventing one would start re-cutting changes for a consumer that never asked.

**A flush-time throw is safe.** A client that already has an oversized change queued now hits the error inside `flushDoc`. That latches the doc and surfaces it, leaving the rest of the docs syncing; `_isRetryableSyncError` returns false for it, so the retry ladder does not burn ten attempts on an op that measures identically every time.

**Telemetry.** A new `onOversizedOp` signal (`easy-signal`, same pattern as the other signals in the package) fires for every op the splitter cannot get under budget, carrying op type, path, measured bytes, the budget, `docId`, `changeId`, a `reason`, and `emitted: true|false` distinguishing emitted-anyway from refused. The four bespoke `console.warn` calls collapse into one in the shared guard.

## Compatibility

`maxUnsplittableBytes` defaults to `Infinity`. Unset, every escape hatch emits exactly as it does today, so **no existing consumer changes behavior**. There is a test pinning this for each of the four hatches and for `breakChangesIntoBatches`. `breakChanges` gains an optional fourth parameter; its existing three-argument signature is unchanged.

## Tests

30 new tests across three files.

- Each of the four hatches: emitted with a report below the ceiling, `UnsplittableChangeError` above it, emitted-anyway when the ceiling is unset.
- `splitLargeInsertText`: high-entropy inserts (100K and 1M characters) produce pieces all measured under budget with `compressedSizeUint8`, text and leading retain preserved, attributes carried, no surrogate pair cut.
- Resplit on 413: halving happens and is monotonic, is bounded by the floor, resets after a successful flush, ignores a 413 scoped to the doc, and does not invent a budget where none was configured.
- `UnsplittableChangeError` is recognised by `isDefectiveChangeError`, including for a name-only copy that crossed a worker boundary, and reaches `Patches.onError` as `kind: 'defective'` with the change rolled back.
- A flush-time throw latches the doc and leaves other docs syncing.

Each source change was reverted in turn to confirm the relevant tests fail without it.

### DIFF
--- a/src/algorithms/ot/shared/changeBatching.ts
+++ b/src/algorithms/ot/shared/changeBatching.ts
@@ -1,5 +1,7 @@
+import { signal } from 'easy-signal';
 import { createChange } from '../../../data/change.js';
 import type { JSONPatchOp } from '../../../json-patch/types.js';
+import { UnsplittableChangeError } from '../../../net/error.js';
 import type { Change } from '../../../types.js';
 
 /**
@@ -29,6 +31,66 @@ export function getJSONByteSize(data: unknown): number {
   }
 }
 
+/** Why the splitter had no seam to cut an op on. */
+export type OversizedOpReason =
+  /** Op type has no splitter (not `@txt`, `replace` or `add`). */
+  | 'op-type'
+  /** A single delta op that isn't a string insert: an embed, or a retain that can't shrink. */
+  | 'delta-op'
+  /** A `replace`/`add` whose value is a string or array, not a structure to pull text out of. */
+  | 'value-not-object'
+  /** A `replace`/`add` object value with no text deltas to extract. */
+  | 'value-no-text'
+  /** A single-character insert that still measures over budget. */
+  | 'insert-chunk';
+
+/** An op the splitter could not break below the storage budget. */
+export interface OversizedOpReport {
+  /** The JSON Patch op type (`@txt`, `replace`, `add`, …). */
+  op: string;
+  path: string;
+  /** Measured size of the op, by the caller's size calculator. */
+  bytes: number;
+  /** The `maxStorageBytes` budget it overshot. */
+  maxBytes: number;
+  docId?: string;
+  changeId?: string;
+  /** True when the op went out anyway; false when it was refused with an `UnsplittableChangeError`. */
+  emitted: boolean;
+  reason: OversizedOpReason;
+}
+
+/**
+ * Fires for every op the splitter cannot break below `maxStorageBytes`, whether it went out anyway
+ * or was refused. Subscribe to route these to telemetry: an emitted one is a change riding above
+ * the intended budget (fine until the store disagrees), a refused one is user work that can never
+ * be saved.
+ */
+export const onOversizedOp = signal<(report: OversizedOpReport) => void>();
+
+/** Options for splitting changes, beyond the byte budget itself. */
+export interface ChangeSplitOptions {
+  /**
+   * Hard ceiling for an op the splitter cannot break apart. Below it an unsplittable op is emitted
+   * as before (and reported via {@link onOversizedOp}); above it the split fails with an
+   * `UnsplittableChangeError`. Defaults to `Infinity`, so every op is emitted, as it always was.
+   *
+   * Keep it well above `maxBytes`: that budget is a conservative split *target*, so refusing at it
+   * would reject changes the store accepts happily. Set this to what the store genuinely rejects.
+   */
+  maxUnsplittableBytes?: number;
+  /** Doc id, carried on {@link OversizedOpReport} for the consumer's telemetry. */
+  docId?: string;
+}
+
+/** Everything the recursive splitters need, threaded as one value. */
+interface SplitContext extends Required<Pick<ChangeSplitOptions, 'maxUnsplittableBytes'>> {
+  maxBytes: number;
+  sizeCalculator?: SizeCalculator;
+  docId?: string;
+  changeId?: string;
+}
+
 /**
  * Break changes into smaller changes so that each change's storage size never exceeds `maxBytes`.
  *
@@ -39,18 +101,62 @@ export function getJSONByteSize(data: unknown): number {
  * @param changes - The changes to break apart
  * @param maxBytes - Maximum storage size in bytes per change
  * @param sizeCalculator - Custom size calculator (e.g., for compressed size)
+ * @param options - Unsplittable-op ceiling and reporting context
+ * @throws {UnsplittableChangeError} When an op can't be split and exceeds `maxUnsplittableBytes`
  */
-export function breakChanges(changes: Change[], maxBytes: number, sizeCalculator?: SizeCalculator): Change[] {
+export function breakChanges(
+  changes: Change[],
+  maxBytes: number,
+  sizeCalculator?: SizeCalculator,
+  options?: ChangeSplitOptions
+): Change[] {
   const results: Change[] = [];
   // Splitting one change into N pieces occupies N revs, so every change after it shifts up
   let revShift = 0;
   for (const change of changes) {
     const shifted = revShift ? { ...change, rev: change.rev + revShift } : change;
-    const pieces = breakSingleChange(shifted, maxBytes, sizeCalculator);
+    const pieces = breakSingleChange(shifted, {
+      maxBytes,
+      sizeCalculator,
+      maxUnsplittableBytes: options?.maxUnsplittableBytes ?? Infinity,
+      docId: options?.docId,
+      changeId: change.id,
+    });
     revShift += pieces.length - 1;
     results.push(...pieces);
   }
   return results;
+}
+
+/**
+ * Report an op the splitter ran out of seams on, and refuse it when it is past the point the store
+ * will accept. Returning normally means "emit it anyway", exactly as before this guard existed.
+ */
+function guardUnsplittable(
+  ctx: SplitContext,
+  op: string,
+  path: string,
+  bytes: number,
+  reason: OversizedOpReason
+): void {
+  const emitted = bytes <= ctx.maxUnsplittableBytes;
+  onOversizedOp.emit({
+    op,
+    path,
+    bytes,
+    maxBytes: ctx.maxBytes,
+    docId: ctx.docId,
+    changeId: ctx.changeId,
+    emitted,
+    reason,
+  });
+  if (!emitted) {
+    throw new UnsplittableChangeError(op, path, bytes, ctx.maxUnsplittableBytes, {
+      docId: ctx.docId,
+      changeId: ctx.changeId,
+    });
+  }
+  console.warn(`Oversized op ${op} at "${path}" is ${bytes} bytes and cannot be split; including it anyway`);
 }
 
 /** Default wire batch size limit (1MB) */
@@ -59,7 +165,7 @@ const DEFAULT_MAX_PAYLOAD_BYTES = 1_000_000;
 /**
  * Options for breaking changes into batches.
  */
-export interface BreakChangesIntoBatchesOptions {
+export interface BreakChangesIntoBatchesOptions extends ChangeSplitOptions {
   /** Batch limit for wire (uncompressed JSON). Defaults to 1MB. */
   maxPayloadBytes?: number;
   /** Per-change storage limit. If exceeded, individual changes are split. */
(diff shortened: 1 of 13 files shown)


src/algorithms/ot/shared/changeBatching.ts
```diff
@@ -329,21 +422,56 @@ function breakTextOp(
...
+      guardUnsplittable(ctx, textOp.op, textOp.path, bytes, 'insert-chunk');
+    }
+    pieces.push(deltaOps);
+  };
+
+  const baseSize = getSizeForStorage({ ...origChange, ops: [{ ...textOp, value: '' }] }, ctx.sizeCalculator);
+  const seedLength = Math.max(1, ctx.maxBytes - baseSize - 20);
+  for (let i = 0; i < text.length; ) {
+    const end = Math.min(text.length, Math.max(i + 1, safeSplitIndex(text, i + seedLength)));
+    emit(text.slice(i, end), i === 0 ? retainPrefix : 0);
```
matalina: **High — pieces after the first compose at position 0, so a multi-chunk split garbles the document when replayed.**

`retainPrefix` goes onto the first piece only; every later piece is a bare `[{insert: chunk}]`. But these pieces are *sequential changes* (consecutive revs), and `@txt` apply is `doc.compose(delta)` (`src/json-patch/ops/text.ts`) — a delta with no leading retain inserts at **position 0**. So `[retain 42, insert "ABCDEF"]` split into `[retain 42, insert "ABC"]` + `[insert "DEF"]` puts `DEF` at the head of the document, not at 45; with 3+ chunks the tail chunks stack up reversed at the head. Correct piece *k* needs `retain = retainPrefix + Σ len(chunk₁..chunkₖ₋₁)` (UTF-16 code units). Server-side transform doesn't rescue it: transforming `[insert]@0` against an earlier sibling piece leaves it at 0.

The first-chunk-only prefix is inherited from the old code, but two things make it this PR's problem:

- **Before, this path mostly self-destructed**: high-entropy multi-chunk splits blew the budget (the 1.9MB defect) and the server rejected them — the wedge. Now the pieces are correctly sized, so they **commit**, and the garbling lands in committed history: every other device and every future load replays it, and a reload mid-flush re-applies the persisted garbled pending locally too.
- **The 413 resplit ladder makes it reachable**: at the 100KB floor, a single string insert only needs ~100KB compressed (a large-scene paste, or a per-scene `@txt` produced by `breakLargeValueOp` from a restore/import seed) to multi-chunk.

The new preservation tests pass over this because they verify by *concatenating* inserts rather than *applying* pieces.

Suggested fix: thread a running position through `emit` — the prefix for each piece after the first is `retainPrefix + charsEmittedSoFar` — and add a test that composes the pieces sequentially onto a document and asserts the result equals composing the original op. (The same accumulation flaw exists in `breakTextOp`'s general multi-piece flush path — `retainToPrefixCurrentPiece` tracks intra-piece retains, not the absolute position across flushed pieces — but that's pre-existing and reasonable as a follow-up; the insert path rewritten here is the one the 413 ladder productionizes.)

src/net/PatchesSync.ts
```diff
@@ -1479,10 +1510,39 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
...
+   * so a 413 that isn't really about size latches the doc instead of looping.
+   *
+   * Only halves a budget the consumer configured: with no `maxStorageBytes` nothing splits at all,
+   * and inventing one here would start re-cutting changes for a consumer that never asked.
+   */
+  private _tryResplitOversized(docId: string, failure: unknown): boolean {
+    if (!(failure instanceof StatusError) || failure.code !== 413 || failure.data?.scope !== 'change') return false;
+    if (!this.maxStorageBytes) return false;
+    const current = this._resplitBudgets.get(docId) ?? this.maxStorageBytes;
+    if (current <= MIN_RESPLIT_STORAGE_BYTES) return false;
```
matalina, line 1539: **Medium — at the floor, a 413 re-enters the *retryable* ladder and re-probes forever.**

413 isn't in `TERMINAL_STATUS_CODES`, so once this returns false at the floor the failure falls through eject (won't corroborate — the change applies fine locally) to the generic latch, where `_isRetryableSyncError` returns **true**: a 10-attempt backoff ladder, then the ~5-minute background reprobe indefinitely. Every attempt re-splits the doc's entire (growing) queue at 100KB with the compression size calculator and resends, to be 413'd again. Not a storm, but it's the same "retry a thing that can never succeed" shape this PR exists to kill — and for the most likely real trigger (an inline image embed no budget can shrink), the floor is the *expected* terminus, not an edge case. Suggest treating a 413 `scope:'change'` as non-retryable once the budget is already at the floor, mirroring the `isUnsplittableChangeError` carve-out in `_isRetryableSyncError`, so it latches definitively like the splitter-refused case.

Related, self-limiting but worth a thought: for a single unsplittable op the ladder burns 4 round-trips (900K→450K→225K→112.5K→floor) resending byte-identical pieces before latching. A cheap short-circuit — skip halving when the named change is a single op the splitter already reported as unsplittable — would get the affected user to a legible latch on attempt 1.

src/algorithms/ot/shared/changeBatching.ts
```diff
@@ -87,12 +193,12 @@ export function breakChangesIntoBatches(
...
 
   const maxPayloadBytes = opts.maxPayloadBytes ?? DEFAULT_MAX_PAYLOAD_BYTES;
-  const { maxStorageBytes, sizeCalculator } = opts;
+  const { maxStorageBytes, sizeCalculator, maxUnsplittableBytes, docId } = opts;
 
   // First, split individual changes if they exceed storage limit
   let processedChanges = changes;
   if (maxStorageBytes) {
-    processedChanges = breakChanges(changes, maxStorageBytes, sizeCalculator);
+    processedChanges = breakChanges(changes, maxStorageBytes, sizeCalculator, { maxUnsplittableBytes, docId });
```
matalina, line 216: **Low — the wire-payload pass below double-reports oversized ops with misleading context.**

The second `breakChanges(processedChanges, maxPayloadBytes)` call (line ~220) passes no options: an op that survives this storage pass (or when `maxStorageBytes` is unset) but exceeds 1MB raw JSON fires `onOversizedOp` again — with `maxBytes` = the wire budget, `docId: undefined`, `emitted: true`. A consumer routing the signal to Sentry gets two reports for one op, one missing the doc and naming the wrong budget. Note the two passes measure differently (compressed storage vs raw wire), so "survived storage, tripped wire" is a real combination, not a can't-happen. Passing `{ docId }` through — and arguably tagging or suppressing the wire-pass report — would keep the telemetry clean.

src/algorithms/ot/shared/changeBatching.ts
```diff
@@ -329,21 +422,56 @@ function breakTextOp(
...
+  const measure = (deltaOps: any[]) =>
+    getSizeForStorage({ ...origChange, ops: [{ ...textOp, value: deltaOps }] }, ctx.sizeCalculator);
+
+  const emit = (chunk: string, prefix: number): void => {
+    const deltaOps: any[] = prefix > 0 ? [{ retain: prefix }] : [];
+    deltaOps.push({ insert: chunk, attributes: attributes ? { ...attributes } : undefined });
+    const bytes = measure(deltaOps);
+    if (bytes > ctx.maxBytes) {
+      if (chunk.length > 1) {
+        const mid = Math.max(1, safeSplitIndex(chunk, Math.ceil(chunk.length / 2)));
```
matalina: **Nit — the never-cut-a-surrogate-pair invariant isn't fully held.**

`Math.max(1, safeSplitIndex(...))` defeats the guard when a chunk *starts* with an astral pair and the safe index came back 0: `mid` is forced to 1, cutting the pair into two lone-surrogate pieces. The seed loop's `Math.max(i + 1, ...)` fallback does the same when `seedLength` is 1. Both are only reachable at pathological budgets (the sane-budget astral test rightly passes), but the `safeSplitIndex` comment claims an invariant the code doesn't hold — either handle the 2-unit-pair case (emit it whole and let `guardUnsplittable` rule on it) or soften the comment.

src/net/PatchesSync.ts
```diff
@@ -1479,10 +1510,39 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
...
+    const next = Math.max(MIN_RESPLIT_STORAGE_BYTES, Math.floor(current / 2));
+    this._resplitBudgets.set(docId, next);
+    this._clearSyncRetry(docId);
+    this._surfacedSyncErrors.delete(docId);
+    console.warn(
+      `Server refused change ${failure.data.changeId} on doc ${docId} as too large; re-splitting at ${next} bytes`
```
matalina: **Nit** — `_tryEjectPoisonChange` requires `typeof data.changeId === 'string'` before using it; this guard only checks `scope === 'change'`, so a 413 without a `changeId` logs `undefined` here. Harmless, but cheap to align with the eject guard.

matalina (commented):
Reviewed the full diff plus the surrounding splitter/sync code on main. The design here is sound and well-documented: the split *target* vs refuse *ceiling* separation is the right call, the ceiling defaulting to `Infinity` keeps every existing consumer byte-identical (and the tests pin that per hatch), the 413 ladder is bounded/monotonic/reset-on-success, and `UnsplittableChangeError` slots cleanly into the existing defective-change taxonomy with cross-realm name matching. The `splitLargeInsertText` rewrite genuinely fixes the compressed-budget-as-character-count defect. Test quality is high overall.

Five findings, left inline — one I'd want addressed before this ships, the rest are judgment calls or nits:

1. **High** — multi-chunk insert pieces compose at the wrong position; a drained backlog can commit *scrambled* text server-side (`changeBatching.ts`, the `emit` loop).
2. **Medium** — at the resplit floor, a 413 re-enters the retryable ladder and re-probes every ~5 min forever (`PatchesSync.ts`, `_tryResplitOversized`).
3. **Low** — the wire-payload split double-reports oversized ops via `onOversizedOp` with no `docId` and the wrong budget (`changeBatching.ts`, `breakChangesIntoBatches`).
4. **Nit** — the never-cut-a-surrogate-pair invariant isn't fully held at pathological budgets.
5. **Nit** — `_tryResplitOversized` logs `failure.data.changeId` without validating it exists.

Two notes without a clean inline anchor:

- **Flush-time refusal is all-or-nothing**: `breakChangesIntoBatches` throws before anything is sent, so clean changes *ahead of* a mid-queue poison change can't commit either. In practice the oversized change is usually the queue head, and the ceiling defaults off — fine, but worth remembering when dw3 turns the ceiling on.
- **Perf**: the bisection measures every candidate chunk with LZ compression — O(n log n) over a multi-MB insert, on whatever thread sync runs on. Rare path, bounded, acceptable; seeding from a measured bytes-per-char ratio would cut passes if this ever shows up in traces.

On test coverage: thresholds, defaults, classification, latch isolation, and the ladder arithmetic are all well covered (the 4-attempt/5-attempt expectations check out against the halving math). The one gap is the one that matters: no test **applies** split pieces to a document — the new preservation tests verify by concatenating inserts, which is exactly what hides finding 1. An apply-based round-trip assertion (split, compose pieces sequentially onto a doc, assert final text equals composing the original op) would have caught it and would pin it forever.

Comment-only review per Alicia — not approving or requesting changes. Everything except finding 1 is shippable as-is; finding 1 is small to fix (cumulative retain in `emit`) and I'd fold it into this PR since the 413 ladder is what makes the path reachable.

PR #134, opened 2026-07-30
Fix JSONPatch.transform/compose aliasing the caller's ops; add co-author convergence fuzz harness

matalina:
Part of the DAB-858 / DAB-876 root-cause work. Two independent pieces, split into two commits for review.

## 1. `JSONPatch.transform()` / no-arg `compose()` aliasing fix

`transformPatch` returns its input ops array **untouched** when no op needed transforming (disjoint paths — the common case), and `composePatch` does the same via `mapAndFilterOps`. Both `JSONPatch.transform()` and no-arg `compose()` handed that aliased array straight to the constructor, which stores it by reference — so the returned patch's builder methods (`this.ops.push`) appended into the **caller's** patch or raw ops array:

```ts
const server = new JSONPatch().replace('/title', 'x');
const local  = new JSONPatch().replace('/body', 'y');   // disjoint — transform is a no-op
const rebased = server.transform(local);                 // rebased.ops === local.ops (pre-fix)
rebased.add('/note', 'z');                               // local silently grows an op it never authored
```

Same aliasing class as the OTDoc optimistic-ops fix (#127), found by auditing every transform-result consumer in `src/` for that class. This was the only remaining site; it has **zero internal callers** (public-API surface only), so it is latent rather than a live data-loss path. Fix is the #127 pattern: copy at the consumer. Six regression tests in `tests/json-patch/transform-aliasing.spec.ts`, including a paths-collide case proving the copy doesn't suppress real transformation.

## 2. Co-author convergence fuzz harness (`tests/integration/coauthor-convergence.spec.ts`)

Two real OT clients (`OTAlgorithm` + `OTInMemoryStore` + `OTDoc`) committing interleaved edits through a real `OTServer`, with the delivery layer simulating production's failure surface:

- **sender-excluded** broadcasts (pup's fan-out since DAB-773 — a client never sees its own commit event),
- **deferred deliveries coalesced** around the recipient's own interleaved commits (the interior-gap shape 0.24.2's guards exist for),
- **dropped** broadcasts, recovered via the `getChangesSince` path (mirroring `PatchesSync.syncDoc`).

Seeded and deterministic (mulberry32; failures print seed + round + the full delivery trace). Every 5 rounds it asserts full convergence — client A == client B == server head, and each open doc == its own store at the same rev, the equal-rev invariant dw3's `equal_rev_content_heal` backstop repairs in production. An `afterAll` gate asserts the hostile paths actually fired across seeds, so a future change can't silently neuter the fuzz.

It **passes on current main**, which is a result in itself: it pins this layer as correct under JSON ops and narrows the DAB-876 hunt to `@dabble/delta` text-op transforms and the app-side fan/IDB layers. Extending it to delta text bodies is the planned next step, and it becomes the regression gate for any fan-out redesign.

## Testing

Full suite green locally: 3,342 passed / 4 skipped, `tests/solid/**` excluded — those fail at **collection** on my Windows setup (vite-plugin-solid's `file:///@solid-refresh` virtual module vs Node's ESM loader) on an untouched checkout too; unrelated to this change and expected to pass in CI.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@dabble/patches",
-  "version": "0.25.2",
+  "version": "0.25.4",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@dabble/patches",
-      "version": "0.25.2",
+      "version": "0.25.4",
       "license": "MIT",
       "dependencies": {
         "@dabble/delta": "^1.2.5",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dabble/patches",
-  "version": "0.25.3",
+  "version": "0.25.4",
   "description": "Immutable JSON Patch implementation based on RFC 6902 supporting operational transformation and last-writer-wins",
   "author": "Jacob Wright <jacwright@gmail.com>",
   "bugs": {
--- a/src/json-patch/JSONPatch.ts
+++ b/src/json-patch/JSONPatch.ts
@@ -183,8 +183,13 @@ export class JSONPatch {
    */
   transform(patch: JSONPatch | JSONPatchOp[], obj?: any): this {
     const JSONPatch = this.constructor as any;
+    // transformPatch hands back its input ops array UNTOUCHED when no op needed transforming
+    // (disjoint paths — the common case), so the result can BE the caller's array. Copy before
+    // wrapping: the returned patch's builder methods push into `this.ops`, and without the copy
+    // they would append into the source patch/array — the same aliasing class as the OTDoc
+    // optimistic-ops fix (#127).
     return new JSONPatch(
-      transformPatch(obj, this.ops, Array.isArray(patch) ? patch : patch.ops, this.custom),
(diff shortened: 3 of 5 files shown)


RyanHavoc:
This changes runtime behavior of the public `transform()`/`compose()` API but leaves `package.json` at [0.25.3](https://github.com/dabblewriter/patches/blob/94b1c09c662e7b2cb34661699f4496eabcc26b5d/package.json#L2-L4) — the fix can't reach dw3 (which pins exact versions) until a bump lands. Precedent for doing it in-branch: #127 (the OTDoc aliasing fix the new comments cite) bumped 0.25.1 -> 0.25.2 in the same commit, and #131 shipped without one and needed the dedicated follow-up #133. Suggest 0.25.4 here.

matalina:
Bumped to 0.25.4 in 4cff08c per the #127 precedent — the same commit also syncs the package-lock root version fields, which had drifted (still 0.25.2 while package.json said 0.25.3), and formats the convergence spec the `format:check` gate was failing on. The earlier red runs were spec-only typecheck errors (fixed in 94b1c09) and then that prettier miss; `type:check` / `lint` / `format:check` / the specs all pass locally, so CI should be green end-to-end on this push.

PR #135, opened 2026-07-31
fix(ot): merge a branch in bounded windows so it stops scaling with branch age (DAB-930)

matalina:
Draft — not to be merged with the current batch. Pairs with dabblewriter/pup#226; see [DAB-896](https://linear.app/dabblewriter/issue/DAB-896/root-cause-review-copy-merge-hard-fails-413-once-the-branch-passes).

## Why

`mergeBranch` read and committed a branch's whole unmerged history as one batch: one unbounded read, one OT transform against the source's concurrent changes, one write batch, and a response carrying every committed change. All four scale with how long the branch has been open — not with the size of the edit.

A real customer branch reached **12,000 changes in 27 hours** of a two-person editorial pass. Pup's unbounded-read cap refused the read outright (fixed in pup#226), but that only moves the cliff: the commit side is still one 12,000-change batch, ~24,000 Firestore writes and a ~20MB response.

## What

Commit at most `maxChangesPerMerge` (default **2,000**) branch changes per window, advancing `lastMergedRev` after each.

A window is exactly what a *repeat* merge already does — same pinned merge base, same `batchId`, same id-dedup — so this introduces no new commit shape. It's the documented "branch stays open, subsequent merges pick up new changes" path, run to completion inside one call.

Termination: a window that doesn't advance the watermark breaks the loop; `MAX_MERGE_WINDOWS` (500) is a backstop for a store silently losing watermark writes.

The return value is still every committed change, so the client contract is unchanged.

## Two changes worth reviewing deliberately

1. **Version copies are bounded to the window** (`endRev <= lastBranchRev`) so a version is never copied ahead of the committed change log. One straddling a window boundary is picked up by the window covering its `endRev`.
2. **The version rev-mapping anchor moved** from the pinned merge base to `max(base, current source tip)`. The base is shared by every window, so anchoring there maps window 2+ onto revs window 1 already occupies. The two coincide on the common path — a single window merging into a source that hasn't moved since the branch point — so this only changes the mapping where it was already wrong. It *is* a behaviour change on the single-window path when the source has moved.

## Trade-off, now documented on `mergeBranch`

A failure part-way leaves earlier windows committed and the watermark pointing at them, so the next attempt resumes rather than redoing. That is a real change from all-or-nothing, and it is the point: a branch too big to merge in one go still converges. The source is never left mid-window — each window's commit is atomic.

## Testing

Full suite green (3,338 passing), lint / format / type-check clean. 4 new integration tests:

- a branch larger than the window merges across several windows, every edit landing exactly once
- windowed and unwindowed merges produce identical source state and rev
- the watermark reaches the branch tip, so a follow-up merge is a no-op
- **resume**: a merge killed mid-way leaves the committed prefix and its watermark, then a retry completes it with no duplicate change ids

The 3 failing `tests/solid/*.spec.tsx` files are **pre-existing on `main`** (a `@solid-refresh` resolution error — the files don't load at all), verified on the base checkout.

## Shipping

Needs a publish plus a pin bump in pup before it reaches production. pup#226 stands alone and unblocks affected customers without it.

### DIFF
--- a/src/server/OTBranchManager.ts
+++ b/src/server/OTBranchManager.ts
@@ -65,7 +65,32 @@ export interface OTBranchManagerOptions {
    * duplicated text is meaningful, is policy that belongs to the consuming server.
    */
   contentDuplicationGuard?: ContentDuplicationGuardOptions;
+  /**
+   * How many branch changes a single merge commit may carry. A merge past this size is
+   * committed as several sequential windows instead of one batch (see
+   * {@link OTBranchManager.mergeBranch}). Defaults to {@link DEFAULT_MERGE_WINDOW_CHANGES}.
+   */
+  maxChangesPerMerge?: number;
 }
+
+/**
+ * Default window size for a chunked merge.
+ *
+ * A merge's cost is superlinear in batch size — every change is transformed against the
+ * source's concurrent changes, persisted, and returned — and an editorial pass on a shared
+ * branch reaches five figures of changes in a day or two (DAB-896: 12,000 changes in 27
+ * hours). Windowing keeps each commit's memory, write batch and response bounded regardless
+ * of how long a branch has been open, at the cost of more round trips on large merges.
+ */
+export const DEFAULT_MERGE_WINDOW_CHANGES = 2_000;
+
+/**
+ * Safety valve on the number of windows one `mergeBranch` call will commit. Termination is
+ * already guaranteed by the watermark advancing every window (a window that does not advance
+ * it breaks the loop), so this only bounds a store whose watermark writes are silently lost —
+ * a pathology that should surface as a stuck merge, not an unbounded write loop.
+ */
+const MAX_MERGE_WINDOWS = 500;
 
 /** Per-merge options for {@link OTBranchManager.mergeBranch}. */
 export interface MergeBranchOptions {
@@ -449,27 +474,77 @@ export class OTBranchManager implements BranchManager {
    *    `updateBranchIf` (see {@link advanceMergeWatermark}), so interleaved merges cannot
    *    regress the watermark; without the capability, legacy max-wins semantics apply.
    *
+   * ## Windowing
+   *
+   * The branch's unmerged changes are committed in windows of at most
+   * `maxChangesPerMerge` ({@link DEFAULT_MERGE_WINDOW_CHANGES}) rather than as one batch, and
+   * `lastMergedRev` advances after each. A window is exactly the work a *repeat* merge already
+   * does — same pinned merge base, same `batchId`, same id-dedup — so windowing adds no new
+   * commit shape, it only stops one merge's size scaling with how long the branch has been
+   * open. Consequences worth knowing:
+   *
+   * - A failure part-way leaves the earlier windows committed and the watermark pointing at
+   *   them, so the next attempt resumes rather than redoing. The source is never left
+   *   mid-window: each window's commit is atomic.
+   * - The duplication guard runs per window against the source's current head, so it still
+   *   sees a replayed seed (which lands in the first window) but judges each window on the
+   *   state the previous ones produced.
+   *
    * @param branchId - The ID of the branch document to merge.
    * @param options - Optional per-merge options (e.g. a content-duplication guard override).
-   * @returns The server commit change(s) applied to the source document.
+   * @returns The server commit change(s) applied to the source document, across all windows.
    * @throws Error if branch not found, already closed/merged, or merge fails.
    */
   async mergeBranch(branchId: string, options?: MergeBranchOptions): Promise<Change[]> {
-    // Load and validate branch
+    const windowSize = this.options.maxChangesPerMerge ?? DEFAULT_MERGE_WINDOW_CHANGES;
+    const committed: Change[] = [];
+    let previousRev: number | undefined;
+
+    for (let window = 0; window < MAX_MERGE_WINDOWS; window++) {
+      const result = await this.mergeBranchWindow(branchId, windowSize, options);
+      if (!result) break;
+
+      committed.push(...result.changes);
+      // The watermark must move every window, or the next read returns the same changes and
+      // the loop spins re-committing them. A store that silently drops the watermark write,
+      // or a concurrent merge that regressed it, stops here with the work done so far.
+      if (previousRev !== undefined && result.lastBranchRev <= previousRev) break;
+      previousRev = result.lastBranchRev;
+
+      if (!result.hasMore) break;
+    }
+
+    return committed;
+  }
+
+  /**
+   * Merge one window of a branch's unmerged changes; see {@link mergeBranch}. Returns
+   * `undefined` when the branch has nothing left to merge.
+   */
+  private async mergeBranchWindow(
+    branchId: string,
+    windowSize: number,
+    options?: MergeBranchOptions
+  ): Promise<{ changes: Change[]; lastBranchRev: number; hasMore: boolean } | undefined> {
+    // Load and validate branch. Re-read per window: the previous window advanced the
+    // watermark this window's cursor is derived from.
     const branch = await this.store.loadBranch(branchId);
     assertBranchExists(branch, branchId);
 
     const sourceDocId = branch.docId;
     const branchStartRevOnSource = await this.resolveMergeBase(branch);
 
-    // Get only unmerged changes: since lastMergedRev (if previously merged) or contentStartRev
+    // Get only unmerged changes: since lastMergedRev (if previously merged) or contentStartRev.
+    // Bounded by the window so the read, the commit and the response all stay proportional to
+    // the window rather than to the branch's whole history.
     const startAfter = branch.lastMergedRev ?? (branch.contentStartRev ?? 2) - 1;
-    const branchChanges = await this.store.listChanges(branchId, { startAfter });
+    const branchChanges = await this.store.listChanges(branchId, { startAfter, limit: windowSize });
     if (branchChanges.length === 0) {
-      return [];
+      return undefined;
     }
 
     const lastBranchRev = branchChanges[branchChanges.length - 1].rev;
+    const hasMore = branchChanges.length === windowSize;
 
     // Re-stamp branch changes for source document context:
     // - baseRev: where the branch diverged from source (for transformation)
@@ -493,16 +568,30 @@ export class OTBranchManager implements BranchManager {
     // query — repeat merges must not re-copy versions already on the source. This also skips
     // the branch's initial version (the branch point already exists in source history) and
     // offline-branch versions.
-    const branchVersions = await this.store.listVersions(branchId, {
-      origin: 'main',
-      orderBy: 'endRev',
-      startAfter,
-    });
+    //
+    // Bounded to this window: a version covering branch revs we have not committed yet would
+    // be copied ahead of the source's change log, which builders can only defer on. One that
+    // straddles the window boundary is simply picked up by the window that covers its endRev.
+    const branchVersions = (
+      await this.store.listVersions(branchId, {
+        origin: 'main',
+        orderBy: 'endRev',
+        startAfter,
+      })
+    ).filter(v => v.endRev <= lastBranchRev);
 
-    // Map branch-local revs onto the claimed source revs of the commit below. Copied versions
-    // must live in the source's rev-space: a branch-local endRev past the source tip would
-    // become the source's version watermark and leave real source revs un-versioned.
-    const toSourceRev = (rev: number) => branchStartRevOnSource + Math.max(0, rev - startAfter);
+    // Map branch-local revs onto the source revs this window's changes will occupy. Copied
+    // versions must live in the source's rev-space: a branch-local endRev past the source tip
+    // would become the source's version watermark and leave real source revs un-versioned.
+    //
+    // The anchor is the source's current tip, not the merge base: the base is pinned for the
+    // life of the branch (and shared by every window), so anchoring there would map the second
+    // and later windows onto revs the first window already occupies. They coincide on the
+    // common path — a single window merging into a source that has not moved since the branch
+    // point — so this only changes the mapping where it was already wrong.
+    const sourceRevBeforeCommit = await this.store.getCurrentRev(sourceDocId);
+    const versionAnchor = Math.max(branchStartRevOnSource, sourceRevBeforeCommit);
+    const toSourceRev = (rev: number) => versionAnchor + Math.max(0, rev - startAfter);
 
     // Copy versions with a deterministic identity: the source copy keeps the branch version's
     // id (version ids are namespaced per doc, so there is no collision on the source). A
@@ -544,7 +633,7 @@ export class OTBranchManager implements BranchManager {
     // merged (never the branch tip — a concurrent branch edit past our snapshot must remain
     // mergeable). CAS-guarded against concurrent merges when the store supports it.
     await advanceMergeWatermark(this.store, branchId, branch.lastMergedRev, lastBranchRev);
-    return committedMergeChanges;
+    return { changes: committedMergeChanges, lastBranchRev, hasMore };
   }
 
   /**
--- a/tests/server/OTBranchManager.integration.spec.ts
+++ b/tests/server/OTBranchManager.integration.spec.ts
@@ -352,6 +352,104 @@ describe('OTBranchManager integration', () => {
     warn.mockRestore();
   });
 
+  // A branch under active editorial review accumulates changes for as long as it stays open —
+  // five figures within a couple of days on a shared book (DAB-896). Merging that in one batch
+  // scales the read, the transform, the write batch and the response with the branch's whole
+  // history, so the merge is committed in windows instead.
+  describe('windowed merge', () => {
+    beforeEach(() => {
+      // Copied versions build against a source with no versions of its own; that warns.
+      vi.spyOn(console, 'warn').mockImplementation(() => {});
+    });
+    afterEach(() => {
+      vi.mocked(console.warn).mockRestore();
+    });
+
+    /** Source at rev 2 with `count` single-op branch edits waiting to merge. */
+    async function branchWithEdits(count: number, managerOptions?: OTBranchManagerOptions) {
+      const ctx = setup(managerOptions);
+      await ctx.server.commitChanges('doc1', [rootChange('s1', { src1: 1 })]);
+      await ctx.server.commitChanges('doc1', [change('s2', 1, '/src2', 2)]);
+      const branchId = await ctx.manager.createBranch('doc1', 2);
+      for (let i = 1; i <= count; i++) {
+        await ctx.server.commitChanges(branchId, [change(`e${i}`, i, `/edit${i}`, i)]);
+      }
+      return { ...ctx, branchId };
+    }
+
+    const expectedState = (count: number) => {
+      const state: Record<string, number> = { src1: 1, src2: 2 };
+      for (let i = 1; i <= count; i++) state[`edit${i}`] = i;
+      return state;
+    };
+
+    it('merges a branch larger than the window across several windows', async () => {
+      const { store, server, manager, branchId } = await branchWithEdits(7, { maxChangesPerMerge: 2 });
+      const listChanges = vi.spyOn(store, 'listChanges');
+
+      const committed = await manager.mergeBranch(branchId);
+
+      // 7 edits at 2 per window: four windows of work, plus the read that finds nothing left.
+      const branchReads = listChanges.mock.calls.filter(([docId]) => docId === branchId);
+      expect(branchReads.length).toBeGreaterThan(1);
+      expect(branchReads.every(([, options]) => options?.limit === 2)).toBe(true);
+
+      // Every edit lands exactly once, and the caller still sees the whole merge.
+      const { state } = await coldLoad(server, 'doc1');
+      expect(state).toEqual(expectedState(7));
+      expect(committed).toHaveLength(7);
+      const ids = await changeIds(store, 'doc1');
+      expect(new Set(ids).size).toBe(ids.length);
+    });
+
+    it('produces the same source state as an unwindowed merge', async () => {
+      const windowed = await branchWithEdits(7, { maxChangesPerMerge: 2 });
+      const single = await branchWithEdits(7, { maxChangesPerMerge: 1_000 });
+
+      await windowed.manager.mergeBranch(windowed.branchId);
+      await single.manager.mergeBranch(single.branchId);
+
+      const a = await coldLoad(windowed.server, 'doc1');
+      const b = await coldLoad(single.server, 'doc1');
+      expect(a.state).toEqual(b.state);
+      expect(a.rev).toBe(b.rev);
+    });
+
+    it('advances the watermark to the branch tip so a follow-up merge is a no-op', async () => {
+      const { store, manager, branchId } = await branchWithEdits(5, { maxChangesPerMerge: 2 });
+
+      await manager.mergeBranch(branchId);
+
+      expect((await store.loadBranch(branchId))?.lastMergedRev).toBe(6);
+      await expect(manager.mergeBranch(branchId)).resolves.toEqual([]);
+    });
+
+    // The point of advancing the watermark per window: a merge that dies part-way is resumable
+    // rather than all-or-nothing, so a branch too big to merge in one go still converges.
+    it('resumes from the last committed window after a mid-merge failure', async () => {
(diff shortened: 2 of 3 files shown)


src/server/OTBranchManager.ts
```diff
@@ -449,27 +474,77 @@ export class OTBranchManager implements BranchManager {
...
    * @throws Error if branch not found, already closed/merged, or merge fails.
    */
   async mergeBranch(branchId: string, options?: MergeBranchOptions): Promise<Change[]> {
-    // Load and validate branch
+    const windowSize = this.options.maxChangesPerMerge ?? DEFAULT_MERGE_WINDOW_CHANGES;
+    const committed: Change[] = [];
+    let previousRev: number | undefined;
+
+    for (let window = 0; window < MAX_MERGE_WINDOWS; window++) {
+      const result = await this.mergeBranchWindow(branchId, windowSize, options);
```
jacwright, line 504: Each window's commit re-reads every change the earlier windows committed, so the read this PR bounds comes back inside `commitChanges`.

`changesToCommit` keeps `baseRev = branchStartRevOnSource` for every window (`resolveMergeBase` returns `mergeBaseRev ?? branchedAtRev`, both pinned for the life of the branch). `commitChanges` then does an unbounded `store.listChanges(docId, { startAfter: baseRev })` (algorithms/ot/server/commitChanges.ts:240) — and on pup that store is the plain one, not the merge view: `PupOTBranchManager(store.mergeReadView(), pupServer.otServer)` and `otServer = new OTServer(store)`. So it carries the default `maxUnboundedReadChanges: 10_000`.

Window k therefore reads (source's own edits since the branch point) + 2,000·(k−1). For DAB-896's 12,000-change branch: window 6 reads 10,000 + however many changes the author made to the manuscript after creating the review copy. One is enough to cross the cap.

What the user gets:
- windows 1–5 committed, watermark at branch rev ~10,000, 10,000 changes now in the manuscript;
- window 6 throws the doc-scoped 413 out of `commitChanges`, so `wrapMergeCommit` re-wraps it as `Error("Merge failed: Status 413: …")`. `data` is dropped, so pup's `onError` never matches `isOversizedHistory` and nothing reaches Sentry — but `parseStatusError` still recovers 413 from the message, so dw3 classifies it `too-large` and shows "Nothing was changed, and retrying will hit the same limit" on a manuscript that just took 10,000 changes;
- the retry resumes at window 6 and reads the same 10,000+, so it fails identically forever.

The branch ends up permanently stuck at a partial merge, which is a worse place to be than today's clean up-front refusal. Same arithmetic for the other reported branch (10,016 changes): window 6, 10,000 committed, 16 stranded.

Re-pinning `baseRev` to the source tip at the start of each window fixes both this and the version-anchor mismatch below — the two are the same underlying issue, that the pinned base stops describing the frame after window 1. If the base has to stay pinned for the id-dedup window, then the commit path needs a bounded read before this can ship.

src/server/OTBranchManager.ts
```diff
@@ -449,27 +474,77 @@ export class OTBranchManager implements BranchManager {
...
     const sourceDocId = branch.docId;
     const branchStartRevOnSource = await this.resolveMergeBase(branch);
 
-    // Get only unmerged changes: since lastMergedRev (if previously merged) or contentStartRev
+    // Get only unmerged changes: since lastMergedRev (if previously merged) or contentStartRev.
+    // Bounded by the window so the read, the commit and the response all stay proportional to
+    // the window rather than to the branch's whole history.
     const startAfter = branch.lastMergedRev ?? (branch.contentStartRev ?? 2) - 1;
-    const branchChanges = await this.store.listChanges(branchId, { startAfter });
+    const branchChanges = await this.store.listChanges(branchId, { startAfter, limit: windowSize });
```
jacwright, line 541: Passing a `limit` turns off pup's byte guard on this read entirely, not just the count cap.

pup's `readPaged` only evaluates the caps on unbounded reads: `if (limit === undefined) this.assertWithinUnboundedReadCaps(docId, changes.length, opsBytes, budget)` (FirestoreOTStore.ts:356). With `limit: windowSize` set, neither `maxUnboundedReadBytes` nor the new `MERGE_READ_BUDGET.maxBytes` from pup#226 is ever consulted here.

So the bound on this read becomes 2,000 documents of arbitrary size. dw3 splits its flushes at ~900KB compressed, so a branch carrying large pastes at, say, 60KB of stored ops per change puts 120MB into one window — refused today by the 64MB byte cap, read straight into the heap after this lands, and 50–150× on decompression against `--max-old-space-size=2560` (env.prod.yaml). That is the Jul 6–8 shape the caps were added for.

It is also the exact inversion of pup#226's argument: that PR concludes the document count is the wrong bound for this read and switches it to bytes; this one bounds it by document count and disables the byte check. A byte-aware window (accumulate until either 2,000 changes or a stored-ops budget) keeps both properties.

src/server/OTBranchManager.ts
```diff
@@ -449,27 +474,77 @@ export class OTBranchManager implements BranchManager {
...
+   * ## Windowing
+   *
+   * The branch's unmerged changes are committed in windows of at most
+   * `maxChangesPerMerge` ({@link DEFAULT_MERGE_WINDOW_CHANGES}) rather than as one batch, and
+   * `lastMergedRev` advances after each. A window is exactly the work a *repeat* merge already
+   * does — same pinned merge base, same `batchId`, same id-dedup — so windowing adds no new
+   * commit shape, it only stops one merge's size scaling with how long the branch has been
+   * open. Consequences worth knowing:
+   *
+   * - A failure part-way leaves the earlier windows committed and the watermark pointing at
```
jacwright, line 486: This trade-off is documented here but silently invalidates two contracts downstream that both still say the opposite.

1. pup's `PupOTBranchManager.mergeBranch` docblock: "A refused merge is retry-safe by construction: the guard runs before the version copies and the commit, so refusal leaves no side effects on the source document." `checkContentDuplication` now runs per window against the source's live head, so a refusal raised in window k>1 leaves windows 1..k−1 committed. For the guard specifically that means the doubling those windows introduced is permanently in the manuscript — the DAB-760 class that survives "reject all tracked changes" — while pup still answers 409 as though nothing happened.

2. dw3's `branches_merge_content_duplication_error` and the new `branches_merge_too_large_error` both open with "Nothing was changed." True today; false for any refusal past window 1.

Even with the read issue above fixed, partial commits need to be visible to the caller — either a distinct response (merged-so-far count / a `partial: true` on the refusal) or the copy on both sides updated to stop asserting a rollback that no longer happens. Worth deciding before the publish, since pup and dw3 both have to move with it.

jacwright (changes requested):
Requesting changes. The windowing is the right idea and the resume test is the right test, but as written a 12,000-change branch still cannot merge — it now fails part-way instead of up front, which is worse than the status quo. Details inline.

On the design question: the window bounds the commit but not the frame every window is committed against. `baseRev` is the pinned merge base for all windows, so `commitChanges`' own `listChanges(source, { startAfter: baseRev })` grows by one window on each pass and the last window reads roughly the whole branch back out. Two ways out: advance the commit frame per window (re-pin `baseRev` to the source tip each window, which also removes the version-anchor special case), or keep the pinned base and give the commit path a bounded read. The first is closer to what the windows already mean.

Ordering across the three: pup#226 first (it stands alone and unblocks today), dw3#1306 any time, this one last behind a publish + pin bump. Worth keeping that order even after the above is fixed — dw3's `too-large` copy says "Nothing was changed", and a partial-commit merge makes that false, so dw3 should be out ahead of it, not behind.

matalina:
All three land. I checked each against the code rather than taking them on trust, and they hold:

**1. The commit frame is the real bug.** `commitChanges` does an unbounded `listChanges(docId, { startAfter: baseRev })` (commitChanges.ts:91), `baseRev` is pinned for every window, and on pup that read goes through the plain store — `otServer = new OTServer(store)` (PupServer.ts:104), not the merge view. So window k reads back everything windows 1..k-1 committed. For a 12,000-change branch: five windows land, window 6 reads 10,000+ and throws, and the retry fails identically. A branch permanently stuck at a partial merge is worse than today's clean refusal. This PR does not ship in this shape.

**2. The `limit` disables the byte guard.** Confirmed: `if (limit === undefined) this.assertWithinUnboundedReadCaps(...)`. So the windowed read is bounded by document count only — the precise inversion of pup#226's argument, in the same read. Noted on that PR too.

**3. Partial commits contradict two live contracts** — pup's "refusal leaves no side effects" docblock and dw3's "Nothing was changed" copy on both merge-failure strings.

On the design question, I do not think I should just re-pin `baseRev` per window without a decision, because that is what the pin exists to prevent: `resolveMergeBase`'s own rationale is that a retry recomputing a *higher* base lets the changes already committed below it escape id-dedup and apply twice. With windows the exposure is narrow — only a crash between a window's commit and its watermark write — and on pup the `_changeIds` markers would catch it at the store, but patches is generic and other backends have no such backstop. That trade wants your call, not mine.

The alternative shape I would put next to yours: leave the base pinned and make the commit read exclude the merge's own batch at query level (`withoutBatchId` already exists on the store options and `commitChanges` filters own-batch in memory immediately after). That keeps the dedup window intact and stops the read growing, but it moves re-send dedup onto `seenChangeIds`, and it touches every commit path rather than just merges — so it is a bigger blast radius for a smaller semantic change.

Either way the byte-aware window (2) and a partial-merge signal (3) come along with it.

Nothing here blocks the other two. pup#226 alone bounds the read that is actually breaking customers today — for the reported branches the commit-side read is only the source's own ~40 changes since the branch point, so a single-shot merge goes through. Ordering as you had it: pup#226, then dw3#1306, this one last and only once the frame question is settled.

matalina:
Tracking note: this work now has its own ticket — [DAB-930](https://linear.app/dabblewriter/issue/DAB-930/branch-merge-commits-the-whole-branch-in-one-batch-the-commit-side) — rather than riding on DAB-896, which is the customer-facing 413 and its pup fix. DAB-930 carries all three findings above and the merge-base decision this is blocked on. Retitled to match; DAB-896 now points here for the deferred half.

src/server/OTBranchManager.ts
matalina: Confirmed, and it is the blocker. Traced it the same way you did:

- `commitChanges` does `store.listChanges(docId, { startAfter: baseRev })` with no `limit` (`algorithms/ot/server/commitChanges.ts:91`).
- `resolveMergeBase` returns `mergeBaseRev ?? branchedAtRev`, both pinned for the life of the branch — so every window commits against the same frame.
- On pup that read is on the plain store: `otServer = new OTServer(store)` (`PupServer.ts:104`), not the merge view pup#226 installs on the branch manager. Default `maxUnboundedReadChanges: 10_000` applies.

So your arithmetic holds: window *k* reads (source's own edits since the branch point) + 2,000·(*k*−1). For the DAB-896 branch that is a hard stop in window 6 with five windows already committed, and the retry reproduces it exactly. A branch stranded at a partial merge is strictly worse than the clean up-front refusal it replaces, so this does not ship as written. #135 is parked and the work is tracked in DAB-930.

**The decision I do not think I should make alone — A or B?**

**A. Re-pin `baseRev` to the source tip per window.** Your preference, and I agree it is what the windows already mean; it also deletes the version-anchor special case this PR had to add. What stops me is that it inverts the property `resolveMergeBase`'s own docblock is built on: *"a retry that recomputed `min(branchedAtRev, tip)` would use a higher base, and the changes already committed below it would escape deduplication and be applied twice."* Windowing narrows the exposure to a crash between a window's commit and its watermark write — but that is exactly the window this code already documents as its crash case, so it is not hypothetical. On pup the `_changeIds` markers (DAB-607) catch it in `saveChanges`; patches ships to any backend and `OTStoreBackend` promises no such thing. So A trades a library-level guarantee for a pup-level one.

**B. Keep the base pinned, bound the commit read.** e.g. push the own-batch exclusion into the query — `withoutBatchId` already exists on the store options, and `commitChanges` filters own-batch in memory immediately after the read anyway, so the transform set is unchanged. The read then returns only the source's own concurrent changes and never grows with committed windows. Cost: re-send dedup has to move to `seenChangeIds`, and it touches every commit path rather than just merges.

My read is B is more conservative and A is cleaner, and the deciding question is whether you are willing to rest merge idempotency on the store's id markers rather than on the base pin. That is your call on this library — tell me which and I will build it, with the byte-aware window and the partial-merge signal from the other two threads coming along with it.

src/server/OTBranchManager.ts
matalina: Confirmed, and it is the sharper version of the point — I wrote both sides of this inversion in the same afternoon and did not notice.

`readPaged` gates on the read being unbounded: `if (limit === undefined) this.assertWithinUnboundedReadCaps(docId, changes.length, opsBytes, budget)`. Passing `limit: windowSize` therefore skips it entirely, so neither the 64MB default nor pup#226's 8MB `MERGE_READ_BUDGET.maxBytes` is consulted on the branch-log read any more. The bound becomes 2,000 documents of arbitrary size — and your worked example is the right shape: dw3 splits flushes at ~900KB compressed, so a branch carrying large pastes puts a window's worth of that straight into the heap, against `--max-old-space-size=2560`.

Agreed on the fix: accumulate until *either* the change count or a stored-ops budget is reached, so a window is bounded on both axes and the byte guard survives. That keeps this consistent with pup#226 rather than quietly undoing it.

One wrinkle worth naming for whoever builds it: patches sees decompressed ops, not the stored bytes pup's cap counts (`CompressedStoreBackend` decompresses on the way out), so a window budget expressed here is in a different unit than `maxUnboundedReadBytes`. Either the store grows a way to report stored size per change, or the window budgets on serialized-ops size and pup keeps its own byte cap as the backstop. Leaning toward the latter as the smaller change, but it needs deciding alongside the frame question above.

src/server/OTBranchManager.ts
matalina: Confirmed on both contracts, and the second one has moved since you wrote this.

1. `PupOTBranchManager.mergeBranch`: *"A refused merge is retry-safe by construction: the guard runs before the version copies and the commit, so refusal leaves no side effects on the source document."* With per-window guarding that is false for any refusal past window 1 — and for the duplication guard specifically it is the bad case, because the doubling the earlier windows introduced is the DAB-760 shape that survives "reject all tracked changes", while pup still answers 409 as though nothing happened.

2. dw3's `branches_merge_content_duplication_error` and `branches_merge_too_large_error` both open with "Nothing was changed." That copy is no longer just written — #1306 merged on 08-03 and pup#226 deployed to production on 08-04, so those strings are what a customer sees today, in twelve languages. Anything that makes a merge commit part-way now has to land *behind* a copy change, not alongside one.

So the partial-merge signal is a prerequisite for windowing rather than a follow-up, and it spans three repos: patches needs a response shape that distinguishes "refused, nothing committed" from "committed N, then refused"; pup needs to carry it through `wrapMergeCommit` (which today re-wraps the error as a plain `Error` and drops `data` entirely, so even the existing `scope` is lost on that path); dw3 needs `classifyMergeError` plus new strings for the partial case.

Noted on DAB-930 as part of the scope. It is also the strongest argument for keeping this parked rather than iterating on it — the sequencing constraint is real and the copy is already live.

src/server/OTBranchManager.ts
jacwright: Not casting a vote on a call that's Jacob's, but I dug into the premise underneath A-vs-B and I think one load-bearing fact in it is wrong, which changes which option is conservative.

**`OTStoreBackend` does specify the id guard.** The argument against A is _"patches ships to any backend and `OTStoreBackend` promises no such thing"_. It does — in `saveChanges`'s own docblock (`src/server/types.ts:88-105`):

> Implementations **should** also enforce [docId, change.id] uniqueness atomically … and throw `DuplicateChangeIdsError` … **This is the authoritative duplicate guard**: the commit path's read-side id dedup **cannot see a committed copy at a rev at or before the incoming `baseRev`** (a retry the client rebased onto a newer tip), nor arbitrate two simultaneous sends of the same change (DAB-607).

So it's a stated library-level contract citing the same ticket, not a pup-level detail. `should` rather than a type-level `must`, so it isn't compiler-enforced — but the interface already designates it authoritative, and already says in as many words that the read-side dedup A would weaken **cannot cover the case you're worried about**. That's the same failure mode from both directions: a committed copy at or below `baseRev`. A doesn't open a new class of exposure so much as lean on the guard the library already says is the one that closes it.

**B's mechanism doesn't exist on this path.** `seenChangeIds` is declared on `LWWStoreBackend` (`src/server/types.ts:212`), not on `OTStoreBackend` — it's the LWW compaction story, where there's no change log to dedup against. Moving OT re-send dedup onto it means adding new optional surface to `OTStoreBackend`, which lands in exactly the place the objection to A was aimed: optional, so any backend not implementing it silently loses dedup. B trades a documented-authoritative guard for one that would have to be invented, with the same optionality.

**Both halves check out on pup.** `FirestoreOTStore.ts:156` records `_changeIds/{id}` markers in the same batch as the changes and imports `DuplicateChangeIdsError` from `@dabble/patches/server`, so the authoritative guard is really implemented on the OT path, not just the LWW one.

One nuance in your favour, since I don't want to flatten it: the `resolveMergeBase` passage you quote is real, but its reasoning is scoped to **clamped** branches — the pin exists because a clamped base is derived from the tip and would otherwise move between attempts. On the healthy path the base is `branchedAtRev`, which is inherently stable. A would make the base move on _every_ branch, so the concern genuinely widens rather than being addressed by the existing text. It's just that the widened case is the one `saveChanges` already owns.

So on the facts I'd lean **A**, and I think the thing that stopped you is narrower than it looked. But the judgement it turns on is still Jacob's: whether an interface `should` is a strong enough promise for a library that ships to arbitrary backends. If the answer is no, the honest version of B isn't `seenChangeIds` — it's promoting the id guard to a `must` on `OTStoreBackend` and bounding the commit read, which gets you A's frame _and_ a library-level guarantee.

Two smaller things while I was in here:

- Your citation for the unbounded read is `commitChanges.ts:91`; it's **line 240**. Line 91 is unrelated. Worth fixing since this comment is what DAB-930 will be worked from.
- The in-memory batch filter at 240 isn't incidental to B — the comment directly above it says the read is unbounded _precisely so_ "the idempotency check must see previously committed changes from this same batch". Pushing `withoutBatchId` into the query removes the thing that read exists for, so B's "the transform set is unchanged" is true while its dedup properties are not. Your cost note covers this, but it reads as smaller than it is.

Agreed with everything in the other two threads, and agreed that the partial-merge signal is a prerequisite rather than a follow-up now that #1306 and pup#226 are live in twelve languages.

jacwright:
The merge-base question is settled, but not in favor of either option as framed — while re-verifying the A/B premises I found a second, independent defect in the windowed design, and it eliminates B outright. Full analysis below since this doubles as the DAB-930 spec.

## Windows 2+ transform in the wrong frame

A window's commit transforms its changes against the source's concurrent changes **as stored** — but stored ops are expressed in the frame they were committed in, and the branch's earlier changes move where those ops land in the branch's frame. The transform walk handles this: it advances each foreign change through the raw ops of the queue as it walks (`transformIncomingChanges.ts` — the `committedOps` accumulator; this is patches#66, the lost-words fix). The catch: that advance consumes the **raw pre-transform ops of every earlier change in the batch**, which only the queue supplies. A client resend always carries its full queue, so the flow this PR modeled windows on gets the advance for free. A window carries only its slice, so the walk meets later windows with foreign ops it never advanced — and their offsets are wrong by exactly whatever the earlier windows inserted or deleted.

Runnable repro on this PR's head, using the integration spec's own helpers — source body `"Hello world"`, branch edits `b1` = insert `"XXXXX "` at 0 and `b2` = insert `"!"` at 8 (between "He" and "llo"), committed separately; concurrent source edit `s2` = retain 5, delete 6 (deletes `" world"`):

```ts
it("windowed and one-shot merges agree on overlapping text edits", async () => {
  const oneShot = await mergeWithWindow(1000); // "XXXXX He!llo"
  const windowed = await mergeWithWindow(1); // "XXXXX! Hello"
  expect(windowed).toBe(oneShot); // FAILS on fc84e5a
});
```

The one-shot merge transforms `b2` against `s2` advanced through `b1` (retain 11, delete 6) and keeps the `!` where it was typed. Window 2's queue has no `b1`, so `b2` meets the raw `retain 5, delete 6`, its insert at 8 falls inside the deleted span, and the `!` collapses to position 5 — into text the author never touched. Silent corruption, no error, every existing test green.

Control experiment: with patches#66's advance reverted, one-shot and windowed _agree_ — both produce the corrupted `"XXXXX! Hello"`. So the walk is correct and #66 is load-bearing; the defect is that windowing starves the walk of the raw queue prefix the advance requires.

|                      | one-shot         | windowed         |
| -------------------- | ---------------- | ---------------- |
| current code         | `XXXXX He!llo` ✓ | `XXXXX! Hello` ✗ |
| #66 advance reverted | `XXXXX! Hello` ✗ | `XXXXX! Hello` ✗ |

Two corollaries:

- **The same bug is latent on `main` today** in the repeat-merge path ("branch stays open, subsequent merges pick up new changes") — tail-only queue, pinned base, identical shape. It has never fired because repeat merges don't happen in practice and it needs overlapping concurrent edits. "A window is exactly what a repeat merge already does" is true, and is the problem: windowing makes that path the common case, and a 27-hour editorial branch with ~40 concurrent source edits on the same scenes is the overlapping case by construction.
- **Option B is dead.** Excluding own-batch rows from the commit read at query level doesn't touch this — the rows B excludes were never the issue; the missing _raw_ queue prefix is, and the committed copies couldn't substitute for it anyway (they're post-transform). Any pinned-base design needs the full raw queue in every window, which is O(branch) per window and defeats the windowing. The equivalence test didn't catch any of this because its edits are `add` ops on disjoint paths with zero source changes after the branch point, so every window fast-forwards and the walk never runs.

## The fix: carry the frame instead of reconstructing it

The walk already computes the advanced foreign ops at every step — it just discards them when the call returns. The correct windowed merge keeps them:

1. `transformIncomingChanges` returns the advanced foreign ops alongside the transformed queue.
2. `mergeBranch` threads that state (the "bridge") through the window loop: lift each window's slice through the bridge, commit it at the **current source tip** (a fast-forward — the commit-side read stops growing entirely, which also resolves the original blocker), advance the bridge through the slice's raw ops, fold in any mid-merge source arrivals from the commit result, repeat.
3. The bridge persists alongside the watermark (small — it's the source's concurrent ops), so a crashed merge resumes bounded and a future repeat merge starts correct; a missing bridge falls back to rebuilding it from the branch's raw log.
4. With the base advancing, retry idempotency rests entirely on the store's write-time id guard, so `OTStoreBackend.saveChanges`'s id-uniqueness contract is promoted from _should_ to _must_ (it is already described as the authoritative guard, and `commitChanges` already resolves `DuplicateChangeIdsError` as a resend).
5. From the earlier threads, unchanged: byte-budgeted windows (count _and_ ops-size, so pup's byte-guard argument survives), and the duplication guard hoisted ahead of window 1 — refusals decide before anything commits, which restores pup's "refusal leaves no side effects" contract and dw3's "Nothing was changed" copy for the 409 path. Partial states then only arise from transient faults, which resume to completion; the partial-merge signal narrows to that crash window (tracked on DAB-930).

**Scope boundary — the other axis.** This bounds the _branch_ side. The mirror problem — a source that has taken thousands of changes since the branch point — is unfixed by this design and by this PR alike: the bridge is O(source-divergence), just as today's one-shot commit read is. That case already fails cleanly today (the merge read budget from pup#226 refuses it), and it keeps that clean refusal here. Bounding the foreign axis is the transpose of this fix — stream the source log past the branch queue with the same walk — and is tracked as a DAB-930 follow-up rather than folded into this change. Note the real-world asymmetry: branches accumulate thousands of changes over weeks (DAB-896's 12,000), while their sources took ~40 concurrent changes, so the branch axis is the one hurting customers.

Since the fix replaces the commit path's architecture rather than adjusting this diff, I'm superseding this PR with a fresh one rather than iterating here — reusing the window loop, watermark CAS, resume semantics, version-copy windowing, and most of the tests from this branch, which all carry over. The frame repro above joins the suite as the failing-first test, plus a repeat-merge-after-source-edit case that fails on `main` today. PR link to follow shortly; review welcome there.

jacwright:
Superseded by #140, which carries this PR's window loop, watermark CAS semantics, resume behavior, and tests forward on top of the merge-frame design from the analysis above (the windowed transform needs the source's concurrent ops advanced through the branch's earlier raw ops, which windows otherwise starve — see https://github.com/dabblewriter/patches/pull/135#issuecomment-5209127033). Thanks Alicia — the bounded-window structure and resume semantics here were the right call and they live on in #140; would love your review over there.

PR #136, opened 2026-08-05
Skip a committed change that can never apply instead of latching the doc forever (DAB-946)

jacwright:
## TL;DR

Today a single broken edit buried in a document's history can wedge that document for a writer on every device they own — the app tries to catch up, fails, reloads, tries again, and loops forever while the doc shows as stuck. This makes the client give up on that one broken edit after it has proven itself unfixable, apply everything around it, and get back to normal syncing. The writer keeps working; the one damaged edit is reported to our error tracking so we can find and repair the affected documents.

## Merge order

## The bug

A committed change in a doc's history can fail strict apply (DAB-946: an index-drifted op committed during a client queue-integrity bug, plus a stateless server that transforms but never state-validates). Every client that must replay across it latches:

1. `_applyServerChangesToDoc` throws `ApplyChangesError`
2. `syncDoc` catches it and reloads the authoritative snapshot
3. the next flush's `catchupChanges` (from a stale baseRev) or `GET _changes?since` re-delivers the same committed change
4. goto 1, every 30-60s, forever, on every device

The server already resolves this exact case: pup passes `{ reconstruction }` to `buildVersionState`, so server version state is computed with `applyChangesForReconstruction`, which **skips the whole unappliable change**. That makes change-granular skipping on the client *converge with the truth the server already serves* — which is what answers the divergence objection on `applyChanges`. (Op-granular skipping would match pre-strict lenient semantics but diverge from today's server snapshots, so it is deliberately not used.)

The divergence warning still holds for the *first* failure — a drifted local replica looks identical from the client's side, and the reload fixes that losslessly. It inverts once authoritative state has been installed and the same change still fails.

## The fix

Two seams, one mechanism. A skipped change is replaced by an **ops-less copy of itself** (keeping id and rev), which is the exact shape of "applied, changed nothing": the batch stays rev-contiguous, the id still confirms any pending copy, and state lands exactly equal to what `applyChangesForReconstruction` computes over the same history. `_neuterPoison` is the single point that does this and reports it.

### Seam 1 — committed batches arriving over the wire (`_applyServerChangesToDoc`)

A **second-strike floor**:

- `_committedApplyFailures` — per-doc memo (changeId → failure count, whether a reload has been installed since, whether the skip was surfaced). Bounded at 50 entries/doc, dropped on untrack/remote-delete. Deliberately **not** cleared on a successful sync: the reload that answers each failure *is* a success, so clearing there would reset the count every lap and the floor would never be reached.
- `_reloadDocFromServer` marks the doc's entries `reloadedSince` right after `saveDoc` installs authoritative state.
- `_skipFlooredPoison` runs at the top of `_applyServerChangesToDoc` — the single choke point every committed batch reaches (flush commit-responses, SSE broadcasts, `getChangesSince` catch-up, the reload-buffer drain, `applyMergeChanges`), so the floor is wired in one place rather than threaded through each.

### Seam 2 — the reload's own materialization (`_loadRepairedSnapshot`)

The reload is the recovery of last resort, so a server snapshot whose own committed envelope won't materialize left **nothing to fall back on**: `loadDoc` threw, the doc latched, and every retry re-fetched identical bytes. Now the unappliable change is skipped and the repaired envelope is re-installed via `saveDoc`, so the store holds only appliable history.

No strike counting here: the evidence bar is met by construction, because the base state and the changes both came straight from the server this instant. A change that fails against them is unappliable against authoritative truth — precisely the condition `applyChangesForReconstruction` exists for. Bounded at 5 repairs per snapshot; beyond that the error propagates and the doc latches as before.

### Telemetry

`CommittedPoisonSkippedError` (`docId`, `changeId`, `rev`, `ApplyChangesError` as `cause`) is emitted through `onError(err, { docId })` once per `(docId, changeId)`, so the app's Sentry handler collects the affected documents for repair.

Kept separate from the `onChangeQuarantined` / `quarantinedChanges` machinery on purpose: that contract is *the user's own unsent work pulled from the outbox*, with consent affordances (`discardQuarantinedChange`, `Patches.ejectPendingChange`) and UX to match. A server-committed change is nobody's unsent edit and is global to the doc — every client sees the same one — so routing it through that signal would have the app offering "recover or discard your edit" for someone else's committed history.

## Deliberate deviation: skip *before* the algorithm, not by retrying after the failure

The obvious shape — catch the failure, re-run the batch with reconstruction semantics — does not work here. The stores persist committed changes **before** the doc-level apply throws (`OTInMemoryStore.applyServerChanges` pushes to `committed`; `OTIndexedDBStore` completes its write transaction at :413, and the snapshot-compaction `applyChanges` at :431 can throw too), and both then drop already-stored revs on a retry. A post-failure retry would fix the in-memory doc while leaving the raw poison in the local store, where every subsequent `getDoc`/`loadDoc` rebuild throws on it — trading an infinite sync latch for an infinite store-rebuild latch.

Pre-filtering means the neutered change is what gets persisted, so the fix is durable across reloads and sessions. The cost is one extra lap: the floor needs a *confirmed* second failure (blind-skipping on the second delivery would break "a transient failure the reload fixed is never skipped"), so the sequence is fail → reload → fail → reload → skip. Each lap is one catch-up round trip, so recovery is seconds instead of never.

## Known gap, deliberately out of scope

A store poisoned in the **persist-then-crash window** — the batch is persisted, the apply throws, and the process dies before the recovery reload's `getDoc` returns — leaves `openDoc` throwing on the next cold start with no network involved. Verified reproducible.

It is narrow (the reload heals the store durably, and `OTIndexedDBStore.saveDoc` clears *all* committed rows before reinstalling the envelope, so any completed reload fully repairs it) and it predates this PR. Closing it needs a **durable** per-`(docId, changeId)` skip set the store materialization can consult — a schema and `ClientAlgorithm`/`OTClientStore` interface addition on a published library. Worth doing deliberately, not smuggled in here.

## LWW needs nothing

Verified: LWW materializes committed ops with `{ partial: true }` everywhere (`LWWDoc.import`, `LWWInMemoryStore`, `LWWIndexedDBStore`) and never routes committed changes through the OT `applyChanges` helper, so it cannot throw `ApplyChangesError` on committed history. Its `{ strict: true }` uses are local optimistic ops (dropped on failure by design) and the `verifyPendingChange` ejection probe. The floor is OT-only and no LWW code changed.

## Tests

`tests/net/PatchesSyncCommittedPoison.spec.ts` — real `Patches`/`OTAlgorithm`/`OTInMemoryStore`/`OTDoc`, only the network faked:

- first failure reloads the authoritative snapshot and skips nothing
- a poison that keeps failing is skipped: the rest of the batch applies, state equals `applyChangesForReconstruction`'s, `committedRev` advances, doc lands `'synced'`, telemetry fires exactly once (and not again on re-delivery), and the store rebuild is clean
- the same poison arriving as commit-response catch-up (the actual production vector) is skipped and the pending flush completes
- a reload snapshot whose own committed tail cannot materialize is repaired and re-installed
- a failure the reload actually fixed keeps its content and is never skipped — one strike is not the floor

Full suite green: 3382 passed / 4 skipped, plus `lint`, `format:check`, `type:check`.

Also corrects the `applyChangesForReconstruction` docstring: change-granular skipping only reproduces pre-strict client behaviour for *single-op* changes, since lenient `applyPatch` dropped just the offending op and kept its siblings.

Docs updated in `docs/PatchesSync.md` and `docs/algorithms.md`. Not published to npm.


src/net/PatchesSync.ts
```diff
@@ -1311,18 +1342,149 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
...
+
+  /**
+   * Mark the doc's recorded apply failures as having survived an authoritative snapshot — the
+   * half of the floor that distinguishes "this replica drifted" (which the reload just fixed)
+   * from "the history is unappliable".
+   */
+  private _markPoisonReloaded(docId: string): void {
+    const memo = this._committedApplyFailures.get(docId);
+    if (!memo) return;
+    for (const entry of memo.values()) entry.reloadedSince = true;
```
matalina: **The skip floor can be reached with zero failures on top of authoritative state.**

The guard is `failures >= POISON_SKIP_FLOOR && reloadedSince`, but nothing ties those two together: `_markPoisonReloaded` flips `reloadedSince` without touching `failures`, so two strikes racked up *before* any reload plus one reload satisfies it. The PR body's invariant is fail → reload → fail → reload → skip; the code implements fail → fail → reload → skip.

The production sequence that produces two pre-reload strikes is the ordinary broadcast path, not an exotic race:

1. A broadcast carrying the change reaches `_receiveCommittedChanges` (:1252) → `_applyServerChangesToDoc` throws → `_recordCommittedApplyFailure` sets `failures=1` (:1348).
2. `_receiveCommittedChanges` catches `ApplyChangesError` and calls `syncDoc` itself (:1264-1266).
3. `syncDoc`'s `getChangesSince` re-delivers the same tail → throws again → `failures=2` (:1348). `reloadedSince` is still `false`.
4. `syncDoc` catches and reloads (:813-817). `saveDoc` + `_markPoisonReloaded` (:973-976) flip `reloadedSince=true`.

The floor is now met and the change has never once been tried against the freshly-installed snapshot. It gets neutered on its very first post-reload delivery — either `_drainReloadBuffer` at :997 *inside that same reload call*, or the next lap's catch-up.

This is the drifted-replica case the design explicitly promises to protect ("a drifted local replica looks identical from the client's side, and the reload fixes that losslessly"). The cost when it misfires is the divergence `applyChanges` refuses to allow: a real committed change silently applied as a no-op on this device only, permanently, plus a false `CommittedPoisonSkippedError` steering the DAB-946 repair sweep at an undamaged document.

```suggestion
    for (const entry of memo.values()) {
      // Pre-reload strikes must not count toward the floor — the reload may well have fixed
      // them. Clamp anything not already floored so at least one failure has to land ON TOP
      // of authoritative state before `_skipFlooredPoison` will neuter the change.
      if (!entry.surfaced) entry.failures = Math.min(entry.failures, POISON_SKIP_FLOOR - 1);
      entry.reloadedSince = true;
    }
```

`!entry.surfaced` keeps an already-decided skip decided, so a reload after the floor was legitimately reached doesn't send the doc back around the latch loop.

tests/net/PatchesSyncCommittedPoison.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+
+    expect(ctx.doc.state).toEqual({ title: 'x', note: 'kept' });
+    expect(ctx.doc.committedRev).toBe(4);
+    expect(ctx.skips).toHaveLength(1);
+    expect(ctx.skips[0]).toMatchObject({ docId: DOC, changeId: poison.id, rev: 3 });
+    // The repaired envelope was re-installed, so the store no longer holds unappliable history.
+    expect((await ctx.store.getDoc(DOC))?.state).toEqual({ title: 'x', note: 'kept' });
+  });
+
+  it('never skips a failure the reload actually fixed — one strike is not the floor', async () => {
```
matalina, line 159: **This test can't catch the pre-reload strike accumulation, because it only ever produces one strike.**

It drives `syncDoc` directly, so lap 1 records exactly one failure and `failures=1 < POISON_SKIP_FLOOR` carries the assertion — not the `reloadedSince` logic it's meant to pin. No test in the file goes through `_receiveCommittedChanges`, which is the actual broadcast vector *and* the one that calls `syncDoc` from its own catch, producing a second strike before any reload runs (see my note on `_markPoisonReloaded`).

Routing the same drift scenario through the broadcast path first should make this test fail today:

```ts
await ctx.sync['_receiveCommittedChanges'](DOC, [note, drifted]); // strike 1, then its catch syncs → strike 2 → reload
await ctx.sync['syncDoc'](DOC);                                    // re-delivery: neutered, though the reload fixed it

expect(ctx.skips).toEqual([]);                                     // fails: one skip reported
expect(ctx.doc.state).toEqual({ title: { a: { b: 1 } }, note: 'n' }); // fails: ops were dropped
```

Worth adding regardless of how the floor is fixed — it's the case that separates "two failures" from "two failures that mean something".

Two other untested bounds while you're here: `MAX_POISON_MEMO_ENTRIES` eviction and `MAX_SNAPSHOT_REPAIRS`. Both are load-bearing safety limits with no coverage.

src/net/PatchesSync.ts
```diff
@@ -1290,6 +1318,9 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
   protected async _applyServerChangesToDoc(docId: string, serverChanges: Change[]): Promise<void> {
     const doc = this.patches.getOpenDoc(docId);
     const algorithm = this._getAlgorithm(docId);
+    // The one choke point every committed batch reaches (flush echoes, broadcasts, catch-up,
+    // merges, the reload drain), so the poison floor is applied here and nowhere else.
+    const changes = this._skipFlooredPoison(docId, serverChanges);
```
matalina: **The skip is reported before the batch is known to have applied.**

`_skipFlooredPoison` runs here at the top, and `_neuterPoison` (:1441-1446) emits `CommittedPoisonSkippedError` and latches `surfaced = true` synchronously — before `algorithm.applyServerChanges` is even attempted at :1346.

If the batch is `[floored-poison, changeB]` and `changeB` throws (or the store write fails), nothing is persisted and the doc stays latched, but telemetry has already asserted the poison was skipped, and `surfaced` guarantees it will never be re-reported on any retry. A DAB-946 repair sweep reading those events treats the doc as resolved while it's still wedged and the raw poison is still in the store — the opposite of what the "for telemetry to enumerate affected documents for repair" contract in `CommittedPoisonSkippedError`'s docblock promises.

Emitting after `applyServerChanges` resolves would keep the report honest. That means `_skipFlooredPoison` returns the neutered batch plus the list of changes it neutered, and the emit happens in the success path — or, more cheaply, `_neuterPoison` stages the errors and `_applyServerChangesToDoc` flushes them once the apply succeeds.

src/net/PatchesSync.ts
```diff
@@ -1311,18 +1342,149 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
...
+    let memo = this._committedApplyFailures.get(docId);
+    if (!memo) this._committedApplyFailures.set(docId, (memo = new Map()));
+    const existing = memo.get(changeId);
+    if (existing) {
+      existing.error = error;
+      return existing;
+    }
+    const entry = { failures: 0, reloadedSince: false, surfaced: false, error };
+    memo.set(changeId, entry);
+    const oldest = memo.size > MAX_POISON_MEMO_ENTRIES ? memo.keys().next().value : undefined;
```
matalina: **Eviction can drop the very entry that is doing the work.**

`memo.keys().next().value` is the oldest *inserted* key, and `_poisonEntry` returns early for an existing entry without reinserting it — so insertion order is first-seen order, and the floored poison is normally first-seen (it's what started the trouble on that doc).

On a doc that accumulates more than 50 distinct failing changes — a badly damaged history, which is the population this PR exists for — the active floored entry is evicted first. `_skipFlooredPoison` (:1388) stops matching it, the doc re-latches, re-earns two strikes, and `_neuterPoison` emits a *second* `CommittedPoisonSkippedError` for the same `(docId, changeId)` — the duplication the `surfaced` flag and the error's docblock both state cannot happen.

Preferring a non-floored victim keeps the decided skips stable:

```suggestion
    const oldest =
      memo.size > MAX_POISON_MEMO_ENTRIES
        ? ([...memo].find(([, e]) => !e.surfaced)?.[0] ?? memo.keys().next().value)
        : undefined;
```

src/net/PatchesSync.ts
```diff
@@ -1311,18 +1342,149 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
...
+        return await algorithm.loadDoc(docId);
+      } catch (err) {
+        const changes = installed.changes;
+        const index =
+          err instanceof ApplyChangesError && repairs < MAX_SNAPSHOT_REPAIRS
+            ? (changes?.findIndex(change => change.id === err.changeId) ?? -1)
+            : -1;
+        if (index === -1) throw err;
+        changes[index] = this._neuterPoison(docId, changes[index], err as ApplyChangesError);
+        await algorithm.store.saveDoc(docId, installed);
```
matalina, line 1475: **One full committed-envelope rewrite per repair, on the recovery path.**

The loop probes with `loadDoc`, neuters one change, saves the whole envelope, and probes again. Per the PR body, `OTIndexedDBStore.saveDoc` "clears *all* committed rows before reinstalling the envelope" — so a snapshot with 5 unappliable changes costs 5 complete rewrites of a large project doc's committed tail, purely to re-probe.

That lands on exactly the devices already showing 4-8.5s IndexedDB write stalls (DAB-766), during recovery, when the doc is already stuck.

`applyChangesForReconstruction` already collects every skipped change in one pass via `onSkippedChange` — running it once over `installed.state` + `installed.changes` would identify the full set, then a single neuter-all + one `saveDoc` + one `loadDoc` finishes the job. It also makes `MAX_SNAPSHOT_REPAIRS` a straightforward count check instead of a loop bound.

src/net/PatchesSync.ts
```diff
@@ -1311,18 +1342,149 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
...
+      try {
+        return await algorithm.loadDoc(docId);
+      } catch (err) {
+        const changes = installed.changes;
+        const index =
+          err instanceof ApplyChangesError && repairs < MAX_SNAPSHOT_REPAIRS
+            ? (changes?.findIndex(change => change.id === err.changeId) ?? -1)
+            : -1;
+        if (index === -1) throw err;
+        changes[index] = this._neuterPoison(docId, changes[index], err as ApplyChangesError);
```
matalina: **The `changes` guard doesn't actually guard the write it precedes.**

`changes?.findIndex(...) ?? -1` defends against `installed.changes` being absent at runtime — which :961 explicitly anticipates ("older/LWW transports may omit it") even though `PatchesSnapshot` types it as required. But `changes[index]` on the last line indexes it unguarded. It's safe only incidentally: the undefined path happens to fall through to `-1` and the early return.

That makes the safety invisible to the next edit. Any change to the `index === -1` condition — e.g. also bailing on a non-`ApplyChangesError` — turns the write into a TypeError on a transport that omits `changes`.

```suggestion
        const changes = installed.changes;
        if (!changes) throw err;
        const index =
          err instanceof ApplyChangesError && repairs < MAX_SNAPSHOT_REPAIRS
            ? changes.findIndex(change => change.id === err.changeId)
            : -1;
        if (index === -1) throw err;
        changes[index] = this._neuterPoison(docId, changes[index], err as ApplyChangesError);
```

src/net/PatchesSync.ts
```diff
@@ -272,6 +285,18 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
...
+   * Per-doc memo of COMMITTED changes that failed strict apply this session (see
+   * `_skipFlooredPoison`): how many times each failed, whether an authoritative snapshot has
+   * been installed since, and whether the skip has been surfaced. Deliberately NOT cleared
+   * when the doc syncs cleanly — the reload that answers each failure IS a success, so
+   * clearing on success would reset the count on every lap and the floor would never be
+   * reached. Session-scoped, dropped with the doc.
+   */
+  private _committedApplyFailures = new Map<
+    string,
+    Map<string, { failures: number; reloadedSince: boolean; surfaced: boolean; error: ApplyChangesError }>
```
matalina: **`error` pins an `ApplyChangesError` graph for the session.**

The entry retains the whole error, whose `cause` is the underlying patch error carrying the offending op — including its `value`, which for a manuscript change can be a large text payload. With up to `MAX_POISON_MEMO_ENTRIES` (50) per doc, deliberately never cleared on success, and released only on untrack/remote-delete, a long-lived tab's retained heap grows with every distinct apply failure across every tracked doc.

`_neuterPoison` only consumes it as `cause` for the emitted `CommittedPoisonSkippedError`, and `_poisonEntry` overwrites it on each failure anyway (so it isn't a history). Keeping just the message — or a small `{ changeId, rev, message }` — would drop the retained graph without changing what telemetry sees.

tests/net/PatchesSyncCommittedPoison.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+  const doc = (await patches.openDoc<TestDoc>(DOC)) as OTDoc<TestDoc>;
+  sync['updateState']({ connected: true });
+
+  const skips: CommittedPoisonSkippedError[] = [];
+  sync.onError(err => err instanceof CommittedPoisonSkippedError && skips.push(err));
+  return { store, algorithm, patches, connection, sync, doc, skips };
+}
+
+describe('PatchesSync committed-poison skip floor', () => {
+  beforeEach(() => {
```
matalina, line 64: Nit: this installs a `console.error` spy with no teardown. `vitest.config.ts` doesn't set `restoreMocks`, and the sibling `tests/net/PatchesSync.spec.ts` pairs each `vi.spyOn` with an `afterEach` restore — worth matching, so the suite states its own cleanup rather than relying on per-file isolation staying on.

```ts
afterEach(() => vi.restoreAllMocks());
```

(`afterEach` needs adding to the `vitest` import on line 20.)

src/net/PatchesSync.ts
jacwright: Fixed in af830ff. `_markPoisonReloaded` now clamps every non-floored entry to `POISON_SKIP_FLOOR - 1` before setting `reloadedSince`, so at least one failure has to land on top of the installed snapshot.

One change from your suggestion: the guard is `isFlooredPoison(entry)` (`failures >= FLOOR && reloadedSince`) rather than `!entry.surfaced`. With the `surfaced` latch moved to after the apply (your other note), floored no longer implies surfaced, and clamping on `!surfaced` would clamp the entry that earned the floor on its post-reload strike — the doc would then never converge (it re-fails, re-reloads, re-clamps forever). The floored predicate is now the single definition of "decided", shared by the skip, the clamp, and the eviction guard.

Also worth recording: the sequence is real but the mechanism is not the refetch. `store.applyServerChanges` persists the committed rows before `otDoc.applyChanges` throws, so `getChangesSince` after that first strike returns nothing new. What produces the second strike is an ordinary re-delivery of the same commit (SSE broadcast + HTTP ack, re-broadcast, catch-up overlap) against the same untouched doc state. Same conclusion, and the new test drives that vector.

tests/net/PatchesSyncCommittedPoison.spec.ts
jacwright: Added in af830ff: "never counts two pre-reload strikes as the floor (the broadcast vector)". It drives `_receiveCommittedChanges` twice with the same commit, then reloads, then syncs — and asserts the drift applied in full with no skip. It fails on the old code (one skip, ops dropped) and passes now.

The batch it drives is a re-delivery rather than your `syncDoc`-from-the-catch step: the store persists the committed rows before the doc apply throws, so the refetch comes back empty and only a second delivery lands the second strike.

Both other bounds are covered too: "keeps the floored entry when the memo evicts past its bound" (`MAX_POISON_MEMO_ENTRIES`, 100 later failures on the same doc) and "gives up on a snapshot needing more repairs than the bound allows" (`MAX_SNAPSHOT_REPAIRS`). All four new tests were checked to fail against the pre-fix source.

src/net/PatchesSync.ts
jacwright: Fixed in af830ff. `_neuterPoison` now stages a `CommittedPoisonSkippedError` instead of emitting it, and the new `_reportSkippedPoison` emits and latches `surfaced` — called after `applyServerChanges` resolves in `_applyServerChangesToDoc`, and after the successful `loadDoc` in `_loadRepairedSnapshot`. A batch that throws downstream reports nothing and stays re-reportable.

New test: "reports the skip only once the batch carrying it has applied".

src/net/PatchesSync.ts
jacwright: Fixed in af830ff, close to your suggestion: eviction now prefers the oldest non-floored entry and only falls back to plain insertion order when every entry is decided. Predicate is `isFlooredPoison` rather than `!surfaced`, since `surfaced` is now set at report time and a staged-but-unreported skip must be protected too.

Test: "keeps the floored entry when the memo evicts past its bound".

src/net/PatchesSync.ts
jacwright: Fixed in af830ff, taking your suggestion — `if (!changes) throw err;` up front, plain `findIndex`, no `?? -1`. The guard now covers the write.

src/net/PatchesSync.ts
jacwright: Fixed in af830ff. `_poisonEntry` stores `trimApplyError(error)`: a copy of the `ApplyChangesError` with the same name, message, stack, and changeId/rev/index, but with the `cause` chain reduced to its message — so the patch error and its op (with the text payload in `value`) are not retained. That trimmed copy is also what goes out as the emitted error's `cause`, so telemetry sees the same thing and no consumer that holds onto it pins the graph either.

tests/net/PatchesSyncCommittedPoison.spec.ts
jacwright: Added in af830ff — `afterEach(() => vi.restoreAllMocks())` plus the import.

src/net/PatchesSync.ts
jacwright: Not batched — kept the probe loop, with the reasoning now written into the docblock (af830ff).

The blocker is that `loadDoc` is the only oracle for what the store will actually replay, and the two shipped stores disagree: `OTIndexedDBStore.getDoc` reads committed rows from `snapshot.rev + 1` and drops anything at or below the snapshot rev, while `OTInMemoryStore.getDoc` replays every row it was handed. A single `applyChangesForReconstruction` pass over `installed.state` + `installed.changes` would therefore identify a skip set from a model the store may not share — and each entry in that set gets persisted as an ops-less change, so over-neutering silently drops ops the store would have applied. On a recovery path that already exists because the doc cannot be opened, that is a worse failure than the writes.

The cost is real and bounded: at most `MAX_SNAPSHOT_REPAIRS` (5) envelope rewrites, only for a snapshot whose own committed tail will not materialize. If we want it cheaper, the honest fix is an algorithm-level "materialize this snapshot without persisting it" probe so the store stays the oracle — happy to file that separately if you think the recovery path warrants it.

matalina: Withdrawing the ask — you are right, and my finding rested on a wrong assumption.

I took `applyChangesForReconstruction` over `installed.state` + `installed.changes` to be an equivalent model of what the store will replay. Your answer shows it is not, and the failure direction is the bad one: over-neutering persists ops-less changes and silently drops ops the store would have applied, on a path that only exists because the doc cannot be opened. Five bounded envelope rewrites is much the cheaper mistake. Keep the loop.

The reason deserves better than a PR thread, though. `OTIndexedDBStore.getDoc` reads committed rows from `snapshot.rev + 1` and drops anything at or below the snapshot rev; `OTInMemoryStore.getDoc` replays every row it was handed. **Two shipped stores materialize the same envelope differently.**

That is a latent trap well past this recovery path — it is exactly the shape where a test against the in-memory store is green while IndexedDB behaves differently in production, and the in-memory store is most of our test surface. Worth its own ticket, and I would rank it above the snapshot-probe optimization you offered to file.

Not blocking this PR.

matalina (approved):
Approving on content — I checked the fix commit rather than taking it on trust.

`_markPoisonReloaded` now clamps every non-floored entry to `POISON_SKIP_FLOOR - 1` before setting `reloadedSince`, so the floor genuinely requires a strike on top of an installed snapshot. The broadcast-vector test fails on the old code, which is the part I cared about.

Your choice of `isFlooredPoison(entry)` over the `!entry.surfaced` I suggested is right and my version was wrong: with the report staged until after the apply, `surfaced` is false while an entry is legitimately floored, so clamping on it would reset the entry that just earned the floor and the doc would never converge. Good catch. The predicate now being the single definition of "decided" — shared by the skip, the clamp and the eviction — is better than what I asked for.

Also verified: the report is staged and only emitted once `applyServerChanges` resolves, eviction prefers a non-floored victim, the `changes` guard covers its write, and `trimApplyError` drops the op payload from the retained graph.

Withdrew my remaining finding (the snapshot-probe batching) in the thread — your reasoning about the two stores disagreeing on what a snapshot replays is correct, and the optimisation would have been a real regression. I do think that divergence deserves its own ticket.

**Sequencing note, not a blocker:** this is third in the recommended release order (#145 → #137 → #136). It heals the read side for poison already committed, so it wants both generators closed before it lands rather than healing against an active source. Approval here is on the code, not a signal to merge ahead of the other two.

matalina:
Follow-up to my approval — walking back the sequencing caveat in it.

I wrote that this is third in the recommended release order and that the approval was not a signal to merge ahead of the other two. That was over-cautious. Since all three ship in one `@dabble/patches` publish, merge order into `main` does not change what users get, and this PR is safe in any position.

Merge it whenever suits. The one real constraint is release-level and does not involve this PR: do not publish #137 without #145. Detail in the [thread on #145](https://github.com/dabblewriter/patches/pull/145#discussion_r3770980940).

jacwright:
Filed the snapshot-replay divergence as DAB-1034. Merge order noted: this lands third, after #145 and #137.

PR #137, opened 2026-08-05
fix(ot): flush the durable pending queue, not the open doc's mirror

jacwright:
## TL;DR

When someone types fast in Dabble while the server is throttling them, some of their edits could be quietly skipped when the app uploaded its work, and others could be uploaded out of order. Skipped edits stayed on the device but never reached the server, and because later edits were built on top of them, the document on the server ended up in a state that doesn't match what the writer saw. This fixes the upload step so it always sends everything that's been saved locally, oldest first.

Investigation-first: the failing tests came before the fix, and they fail on `main`.

## Merge order

## The mechanism

A document's unsent changes live in two places:

- **the store** (IndexedDB) — the durable queue, and the only thing that assigns each change its `rev` (revision number), stamped inside the write transaction
- **the open document** — an in-memory mirror of that queue

`OTAlgorithm.getPendingToSend` (the function that decides what goes on the wire) read the **mirror**, then topped it up with store rows whose `rev` was **greater than the mirror's last rev**:

```ts
const pending = otDoc.getPendingChanges();
const tail = pending[pending.length - 1]?.rev ?? otDoc.committedRev;
const delta = await this.store.getPendingChanges(docId, { startAfterRev: tail });
```

The comment above it assumed another context's change always lands "past the doc's tail". It doesn't. Revs come from the **store's** tail, so ordering between contexts is decided by who writes first, not by whose mirror is ahead:

1. another context (second tab / worker) mints and takes rev N+1
2. this document mints next, and the store stamps it N+2 — its mirror's tail is now N+2
3. the flush asks the store for rows `> N+2` and gets nothing

The row at N+1 is **durable, unsent, and invisible to every subsequent flush**. It only re-enters the queue when a commit echo rebuilds the queue from the store (`_collectPending`, which does read the store as ground truth). By then:

- changes minted *after* it have already committed at *lower* revs — **reordering**, and
- it gets rebased against those changes; if they touched the same region, `rebaseChanges` transforms it away to nothing and drops it — **the persisted change is never committed at all**, while later committed changes still reference its effects. That is the divergence signature in DAB-946.

The two halves of the queue were using **different authorities**: the receive path (`_collectPending`) treats the store as ground truth and the mirror as the supplement; the send path did the opposite.

The failing conditions are exactly the reported ones: an opaque commit failure (a 429'd CORS preflight surfaces as a status-less `TypeError`, so there is no HTTP status to act on) widens the window in which the queue is assembled from more than one context before anything is flushed.

## The fix

`getPendingToSend` now reads the store and merges doc-only changes through the same `_collectPending` helper the receive path uses — one contract on both sides. Net −5 lines.

The `startAfterRev` ranged read is gone from the send path, which also removes the back-compat hazard its old id-filter existed to paper over (a custom store ignoring `startAfterRev` would return the whole queue and fold the mirror in twice).

## Tests

`tests/net/PatchesSyncQueueOrdering.spec.ts` — all three fail on `main`:

| test | on `main` | expected |
| --- | --- | --- |
| flushes every durable pending change | sends `['mine']`, strands the older change | `['foreign', 'mine']` |
| never commits an older pending change after a newer one | `['second', 'first']` | `['first', 'second']` |
| withholds nothing when a flush fails opaquely and mints keep arriving | `['b','b','d','a','c']` | `['a','b','a','b','c','d']` |

The third drives the production shape end to end: a commit dies without a status, edits keep arriving during the retry window, and the queue must still drain in mint order with nothing withheld.

Checks: `npm test` (3380 passed, 4 skipped) · `type:check` · `lint` · `format:check`.

## Scope / not in this PR

Three adjacent defects surfaced while investigating. Flagging them rather than folding them in:

1. **`@blockable` on `PatchesSync._receiveCommittedChanges` is inert.** `simplified-concurrency` only defers a `blockable` call while a `blockWhile` call is live, and nothing in the codebase calls `blocking`/`blockWhile` any more. The stated guarantee — "while a doc is sending changes to the server, it isn't receiving changes from the server" — does not hold today.
2. **`_withConsistentBaseRev` relabels stragglers without transforming them.** It raises a change's `baseRev` to the queue max, so the server never transforms its ops against the changes in `(trueBase, maxBase]` and commits them in the wrong frame. It fires routinely under rapid mints, not just after the torn reload its comment describes.
3. **`serialGate` has a re-entrancy window.** `run()` invokes the target before `inFlight.set(key, promise)`, and `syncDoc` synchronously notifies `docStates`/`syncStatus` subscribers before its first `await`; a subscriber that calls back into `syncDoc` gets a second concurrent pass. It also returns the *in-flight* promise to a queued caller, so awaiting `syncDoc` does not await the work that caller triggered.


src/client/OTAlgorithm.ts
```diff
@@ -130,26 +130,21 @@ export class OTAlgorithm implements ClientAlgorithm {
...
-      const have = new Set(pending.map(c => c.id));
-      batch = [...pending, ...delta.filter(c => !have.has(c.id))];
-    } else {
-      batch = await this.store.getPendingChanges(docId);
-    }
-    if (batch.length === 0) return null;
-    return this._withConsistentBaseRev(docId, batch);
+    const otDoc = doc as OTDoc<any> | undefined;
+    const committedRev = otDoc ? otDoc.committedRev : await this.store.getCommittedRev(docId);
+    const { pending } = await this._collectPending(docId, otDoc, committedRev);
```
matalina: **A doc-only pending change below the store tail is now withheld from the wire.**

`_collectPending` folds the mirror in only for `c.rev > latestRev`, where `latestRev` is the store's tail (:576-578). The old send path started from the mirror and carried *all* of it. So a change that lives in the open doc but not in the store, at or below the store tail, used to be flushed and now never is.

Reachable, and the `_collectPending` docstring already names the case ("a torn store write — a change persisted only to the doc"):

- `handleDocChange` writes the store (:118) *before* mirroring into the doc (:121), so a doc-only row means the store write reported success without persisting — the WebKit silent-transaction stall this codebase already tracks — or a later store-side removal (`reconcilePending` / `ejectPendingChange` / `replacePendingChanges`) the mirror hasn't followed.
- Doc has a change at rev 11 the store lacks; a foreign context then mints and the store tail moves to 12. `latestRev = 12` → `docOnly` is empty → the rev-11 change is gone from every future batch.

The user-visible shape is bad: `hasPending` reads the store, so nothing reports as pending, while the editor still shows the content. Silent loss with no indicator.

The rev guard earns its keep on the *receive* path — it stops a rebased-away lower-frame copy resurrecting (P3, fuzz seed 1000319). On the send path the same guard means "never transmitted", which is a materially different trade. Unifying the contract is right; inheriting this particular clause into the send path looks incidental rather than decided, and the PR body doesn't mention it. Worth either an explicit carve-out for doc-only rows on the send side, or a comment on `_collectPending` saying the send path accepts the drop and why.

src/client/OTAlgorithm.ts
```diff
@@ -130,26 +130,21 @@ export class OTAlgorithm implements ClientAlgorithm {
...
+    if (pending.length === 0) return null;
+    return this._withConsistentBaseRev(docId, pending);
```
matalina, line 198: **This widens the batch feeding `_withConsistentBaseRev`, which the PR body itself flags as broken.**

The batch now includes every store row, foreign-context mints below the mirror tail included. Those carry the *minting* context's `baseRev`: two tabs where one has received a commit the other hasn't mint at different `committedRev`s, so a mixed-baseRev batch stops being the torn-reload rarity `_withConsistentBaseRev`'s docstring (:535-544) describes and becomes the ordinary multi-tab case.

`_withConsistentBaseRev` then bumps the stragglers' `baseRev` to the max **without transforming their ops** (:554) — scope item #2 in your own PR description, and precisely what #145 fixes. So #137 landing alone would increase DAB-951 poison generation in production, on the same multi-tab population this PR is trying to protect.

Nothing in the repo enforces the "ship all three in one release" note from Slack. Worth making the dependency explicit in the PR description (or stacking #137 on #145) so a future bisect/backport doesn't take this one on its own.

Minor knock-on: the `console.warn` at :551 will label these "(torn-reload straggler)", which will be wrong for most of what it now fires on.

src/client/OTAlgorithm.ts
matalina: **Full pending-queue scan per flush batch.**

`_collectPending` does an unranged `store.getPendingChanges(docId)` (:572), and `PatchesSync.flushDoc` calls `getPendingToSend` once per batch inside the send loop (`PatchesSync.ts:1170`). Draining a K-batch queue therefore re-reads the entire remaining queue K times, where the old doc path read the mirror from memory and asked the store only for `rev > tail`.

That's O(K²) row reads on reconnect with a large offline queue — DAB-810's 369k-word local-only project is the extreme, and it lands on the same devices already showing 4-8.5s IndexedDB stalls (DAB-766).

Reading the store *is* the right call; re-reading all of it per batch isn't required to get there. Either hoist one store read out of the flush loop and drop the rows just committed, or give `_collectPending` an optional floor that the send path passes as "the rev I already committed this pass" (distinct from the old mirror-tail floor, which is what was wrong).

src/client/OTAlgorithm.ts
```diff
@@ -130,26 +130,21 @@ export class OTAlgorithm implements ClientAlgorithm {
...
-      // twice — a duplicate commit the server's id dedup would otherwise have to catch.
-      const have = new Set(pending.map(c => c.id));
-      batch = [...pending, ...delta.filter(c => !have.has(c.id))];
-    } else {
-      batch = await this.store.getPendingChanges(docId);
-    }
-    if (batch.length === 0) return null;
-    return this._withConsistentBaseRev(docId, batch);
+    const otDoc = doc as OTDoc<any> | undefined;
+    const committedRev = otDoc ? otDoc.committedRev : await this.store.getCommittedRev(docId);
```
matalina: Dead store round-trip on the `!doc` branch: the awaited `getCommittedRev` result can't reach the return value. `_collectPending` uses `committedRev` only as the `latestRev` fallback *inside* `if (doc)` (:576) and as the `tailRev` seed (:581), and `getPendingToSend` destructures `{ pending }` only.

`getOpenDoc` returning null is a normal call shape here (`PatchesSync.spec.ts:2675` asserts it), and `syncAllKnownDocs` hits it per doc — so this is an extra IndexedDB round-trip per doc on every reconnect sweep, for a value that is discarded.

```suggestion
    const committedRev = otDoc?.committedRev ?? 0;
```

src/client/OTAlgorithm.ts
```diff
@@ -130,26 +130,21 @@ export class OTAlgorithm implements ClientAlgorithm {
     return pending.length > 0;
   }
 
+  /**
```
matalina, line 156: Two contract docs outside this diff still state the rule this PR inverts, and both are worth pulling in:

**`src/client/ClientAlgorithm.ts:94-96`** — the interface contract, i.e. what custom-algorithm and custom-store implementors build against:

> When `doc` (the open doc) is passed, OT trusts its in-memory pending as the source of truth and only reads the store for foreign-tab mint deltas — no state materialization.

That is now exactly backwards. Since it's the published interface, leaving it means the repo asserts both rules in two places.

**`tests/client/OTAlgorithm-getPendingToSend.spec.ts:66-70`** — the describe block still reads "trusts the doc for pending ... only does a ranged store read past the doc's tail to fold in a foreign tab's mint (R3a)", and the last test's rationale at :125 ("the id filter must keep the doc's own pending from folding in twice") defends the filter this PR deletes.

I traced all four of those tests against the new implementation and they still pass — the store is empty or agrees in each case — so nothing goes red. That's the problem: the contract text just quietly becomes wrong with no signal. Re-pointing both docstrings at `_collectPending` (and dropping the R3a label) would keep the "one contract on both sides" claim true in the places people actually read it.

tests/net/PatchesSyncQueueOrdering.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+  let store: OTInMemoryStore;
+  let algorithm: OTAlgorithm;
+  let patches: Patches;
+  let conn: ReturnType<typeof makeConnection>;
+  let sync: PatchesSync;
+  let doc: any;
+
+  /** A second context over the same store, exactly as another tab would run. */
+  async function foreignMint(label: string): Promise<void> {
+    const foreignDoc = algorithm.createDoc<Doc>(DOC_ID, (await algorithm.loadDoc(DOC_ID)) as any) as OTDoc<Doc>;
```
matalina: The "second context" shares the `algorithm` instance under test, so it also shares `_docLocks` — and `_withDocLock`'s docstring is explicit that this is not the production topology:

> Cross-context (multi-tab) safety is the store's, not this lock's — foreign tabs run their own instance. (`OTAlgorithm.ts:519`)

So every `foreignMint` is fully serialized against this doc's mints. The tests reproduce the *rev-ordering consequence* of the bug faithfully, but never the concurrency that produces it: no interleaving between a foreign `savePendingChanges` and this doc's `getPendingToSend`, and the R2 conflict-replace retry loop in `applyServerChanges` (:228-244) never fires.

```suggestion
    const foreignAlgorithm = new OTAlgorithm(store);
    const foreignDoc = foreignAlgorithm.createDoc<Doc>(DOC_ID, (await foreignAlgorithm.loadDoc(DOC_ID)) as any) as OTDoc<Doc>;
```

(and `foreignAlgorithm.handleDocChange(...)` on the next line). Same store, separate lock — which is the topology the fix is actually about.

tests/net/PatchesSyncQueueOrdering.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+    doc.change((patch: any) => patch.replace('/items', []));
+    await settle(1);
+    await sync['syncDoc'](DOC_ID);
+    conn.commitChanges.mockClear();
+    // Mint offline for the rest of each test, so the queue is assembled before any flush
+    // rather than being drained one change at a time by the per-change auto-sync.
+    sync['updateState']({ connected: false });
+  });
+
+  it('flushes every durable pending change, not just the ones the open doc mirrors', async () => {
```
matalina, line 105: Coverage gap: all three tests put every change in the store first, so the `docOnly` branch of `_collectPending` is never exercised from the send side — even though the send path now depends on it, and its `rev > latestRev` guard is what decides whether a doc-only change ever reaches the wire (see my note on `OTAlgorithm.ts:145`).

A fourth test would settle which contract is intended:

```ts
it('...a change the store lost but the doc still holds', async () => {
  doc.change((patch: any) => patch.add('/items/-', 'orphan'));
  await settle();
  // Simulate the store write that reported success without persisting.
  const rows = await store.getPendingChanges(DOC_ID);
  await store.replacePendingChanges(DOC_ID, rows, []);
  await foreignMint('later');            // store tail now above the orphan's rev
  ...
});
```

Whichever way it asserts, it turns an inherited receive-path clause into a decided send-path contract.

tests/net/PatchesSyncQueueOrdering.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+import { makeConnection } from './connectionMock.js';
+
+interface Doc {
+  items?: string[];
+}
+
+const DOC_ID = 'doc1';
+
+/** Drain timers + microtasks so a mint's store write and its onChange dispatch have landed. */
+async function settle(turns = 3): Promise<void> {
```
matalina: Nit: `turns = 3` is a magic number over real timers, and callers pick 1 or 3 with no stated rule — it encodes how many macrotask turns `OTInMemoryStore` happens to need today.

If a store write ever takes one more turn (swapping in the IndexedDB store, or an added await inside `handleDocChange`), the assertions read a half-built queue and fail in a way that looks like a product regression rather than a test-timing one.

`foreignMint` already returns the `handleDocChange` promise, so awaiting that covers the foreign side; for the doc side `vi.waitFor` on the queue length states the actual precondition:

```ts
await vi.waitFor(async () => expect(await store.getPendingChanges(DOC_ID)).toHaveLength(2));
```

src/client/OTAlgorithm.ts
jacwright: Escalating rather than deciding it here. You are right that inheriting the \`rev > latestRev\` guard onto the send path was incidental, not chosen, and that "never transmitted" is a materially different trade from "does not resurrect". Taking the contract question (store-authoritative vs an explicit doc-only carve-out on the send side) to Jacob; no send-path semantics changed in d442c65.

Everything else in the review is addressed there.

src/client/OTAlgorithm.ts
jacwright: Agreed, and the dependency is now in the code rather than in Slack. d442c65 cross-references #145 from both \`getPendingToSend\` and \`_withConsistentBaseRev\`, stating that a store-backed batch makes mixed baseRevs ordinary and that the relabel has to become a real transform in the same release. #145 does that; the two ship together.

Also dropped "(torn-reload straggler)" from the warning — as you note, it would be wrong for most of what it now fires on.

src/client/OTAlgorithm.ts
jacwright: Fixed, and it turned out cheaper than either option you sketched: the per-batch read was vestigial. \`batches\` is split once before the loop, so the \`getPendingToSend\` at PatchesSync.ts:1170 never fed the next batch despite its comment — its only consumer was \`_stableDocStatus(docId, pending)\` in the catch, which reads \`pending.length > 0\` and nothing else. That boolean is already true from the flush-start read (an empty queue returns before the loop).

d442c65 deletes it. Draining a K-batch queue is back to one queue read per flush.

src/client/OTAlgorithm.ts
jacwright: Fixed in d442c65, taking your suggestion. Left a line on why \`0\` is safe: the value only reaches the doc-merge floor, and that branch requires a doc.

src/client/OTAlgorithm.ts
jacwright: Both updated in d442c65. \`ClientAlgorithm.getPendingToSend\` now states the store-as-truth / doc-as-supplement rule, and the open-doc describe block points at \`_collectPending\` with the R3a label dropped.

The two tests underneath were leaning on the same stale premise, so they moved with it. The last one is the bigger change: "a custom store ignores startAfterRev" is moot now that the send path passes no range, so it now puts the doc's copies at revs *above* the store's — a re-stamp the mirror has not followed. That fails on rev-merge alone and passes only because of the id filter, so the guarantee it claims to pin is now the one it actually pins.

tests/net/PatchesSyncQueueOrdering.spec.ts
jacwright: Done in d442c65 — one \`foreignAlgorithm = new OTAlgorithm(store)\` held for the describe, so successive foreign mints serialize with each other (one tab) but not with this doc's (separate \`_docLocks\`), which is the real topology.

Verified it still catches the bug rather than just passing differently: with \`src/client/OTAlgorithm.ts\` and \`src/net/PatchesSync.ts\` checked out from origin/main, the three regression tests still fail on the new instance split.

tests/net/PatchesSyncQueueOrdering.spec.ts
jacwright: Added in d442c65, written to pin current behavior without deciding the contract — so it is coverage, not a verdict.

It exercises the \`docOnly\` branch from the send side with the doc-only change ABOVE the store tail (mint, drop the row via \`store.dropPendingChanges\`, flush), which is the half of the merge rule both paths already agree on. It passes on main and on this branch.

The at-or-below case your sketch reaches with \`foreignMint('later')\` is exactly the contract question, so it stays out until that is decided — see the reply on OTAlgorithm.ts:145. Once it is, that test is a two-line extension of this one.

(\`store.replacePendingChanges\` does not exist on \`OTClientStore\`; \`dropPendingChanges\` is the equivalent.)

tests/net/PatchesSyncQueueOrdering.spec.ts
jacwright: Fixed in d442c65. \`settle(turns)\` is gone; \`waitForQueue(n)\` states the precondition via \`vi.waitFor\` on the store queue length, as you suggested, and \`settle()\` is now a single yield used only where no queue state is being asserted.

Two things fell out of it, both cases where the old turn count was hiding the real precondition rather than expressing it:

- \`beforeEach\` cannot assert queue length 1 for the setup change — it is minted while connected, so the auto-sync may already have drained it. It now waits for the commit to have happened and the queue to be empty, which is what the tests actually need.
- The third test's "all four queued before the next flush" only held by turn-count luck: with a real wait the auto-sync drains c/d as they arrive. It now mints the retry-window edits offline (the same trick \`beforeEach\` already uses) so the precondition holds by construction.

Suite: 3518 passed, 4 skipped. That file also went from ~4.0s to ~0.3s.

src/client/OTAlgorithm.ts
matalina: Answering the escalation: **store-authoritative** — keep the current behaviour — but make the drop visible.

The reasoning. If the store does not have it, it is not durable, and transmitting content the durable queue never accepted lets the server hold something no local rebuild reproduces. That is a worse divergence than not sending. Store-as-truth is also what makes the rest of this PR coherent: a carve-out on the send side re-opens the split-authority problem the change closes.

What should not survive is the silence. Today a doc-only row at or below the store tail means the editor shows content, `hasPending` reads the store and reports nothing, and it never syncs — no indicator, no telemetry, nothing to diagnose from. A doc holding a change the store lacks is a store-integrity failure, and the known producer is the WebKit silent-transaction stall, so it should read as one:

- surface it through `onError` when `_collectPending` drops a doc-only row on the send side, or
- reflect it in `hasPending` so the doc stops claiming it is fully synced.

Either is fine. The point is that the condition stops being invisible, because right now it is indistinguishable from working correctly.

And please write the decision into `_collectPending` rather than leaving it inherited. The rev guard earns its keep on the receive path for a specific reason (P3, fuzz seed 1000319); the send path now shares the clause for a different one, and the next reader should not have to derive which.

With that settled, the doc-only test you held back can go in — asserting the at-or-below case is dropped rather than sent.

matalina:
**Merge-order constraint** — flagging at top level because it is not visible from the diff.

This should not land before #145.

It widens the flush batch to include foreign-context rows below the mirror tail, which pushes more mixed-baseRev changes through `_withConsistentBaseRev` — and on `main` that still re-stamps a straggler baseRev without transforming its ops (DAB-951). So this PR **alone** increases poison generation in production, on the same multi-tab population whose reordering it is fixing.

#145 replaces that relabel with a real transform.

Recommended sequence for the three-PR release is **#145 → #137 → #136**, with `OTAlgorithm-frameDebt.spec.ts` and `PatchesSyncQueueOrdering.spec.ts` run together on the merged tree before cutting — a check only possible once both #145 and this are in, since this PR is where the queue-ordering spec arrives.

Reasoning in the thread on #145.

matalina:
Adding a concrete instance to the constraint above — **this PR and #145 cancel each other in `flushDoc`**, and it will probably auto-merge clean.

`d442c65` here deletes the per-batch `getPendingToSend` at the bottom of the batch loop (correctly — it was vestigial). #145 `b87ef19` builds its follow-up progress check on exactly that read: `headBefore` captured before the loop, compared against `pending[0].id` after it.

Merged, `pending` at that comparison is still the flush-start array, so `drained` is always `false` and the follow-up pass only fires on `reloadedMidFlush` — silently reverting #145 seam 3, so a deferred baseRev run sits unsynced until the next mint or reconnect.

The hunks are adjacent, not overlapping, so there is likely no conflict marker. Suggested merged form and full detail in the [thread on #145](https://github.com/dabblewriter/patches/pull/145#discussion_r3770980940).

src/client/OTAlgorithm.ts
jacwright: Implemented as decided: behavior kept, silence removed.

The send path stays store-authoritative — a doc-only row at or below the store tail is still never transmitted. What is new is that the withholding reports itself: `_collectPending` now returns those rows as `withheld`, and `getPendingToSend` emits an `UnstoredPendingError` (docId + changeIds) on a new algorithm-level `onError`. `PatchesSync` forwards it to its own `onError` in the same shape it already forwards `connection.onError`, so it reaches app telemetry with no consumer change. Emitted before the empty-queue return, since a withheld row is exactly the case where everything else reads as fully synced. At-least-once, documented on the error: key on docId + changeId to dedupe.

I took `onError` rather than `hasPending` for two reasons. `hasPending(docId)` takes no doc, so making it doc-aware changes the `ClientAlgorithm` contract and every call site. And it would latch true for a row that can never be sent, which turns a diagnosable integrity failure into a permanently unsynced indicator with no recovery path and repeated flush attempts behind it. The error names the condition instead of just colouring it.

The decision is written into `_collectPending` rather than left inherited: the rev guard now states its two separate reasons, receive (a copy the store already rebased away, so folding it back resurrects it — P3, fuzz seed 1000319) and send (not durable, so transmitting it would leave the server holding content no local rebuild reproduces, and carving out the send path re-opens the split authority this PR closes).

The doc-only test I held back is in, asserting both halves: the row does not reach the wire, and the drop surfaces. Verified it fails in each direction rather than trusting it — with the emit removed the report assertion fails; with the rev guard relaxed so the row is merged in, `expected ['later', 'orphan'] to deeply equal ['later']`.

Checked for false positives too: across `tests/client` and `tests/net` the only emission in the whole suite is this test's own. Full suite 3519 passed, 4 skipped; type, lint and format clean.

jacwright:
Reconciled in `b230f0d` on this branch, against a tree that already has #145 rather than against this branch alone.

The deletion stands (the per-batch read never fed the next batch), but one read comes back after the loop, assigning into `pending` so #145's `drained` comparison reads the queue the flush left behind rather than the flush-start array. Both wins survive: K reads collapse to one, and the follow-up gate keeps a sound progress signal.

Verified on a local merge of the two branches: `PatchesSyncFollowUpFlush.spec.ts`, `OTAlgorithm-frameDebt.spec.ts` and `PatchesSyncQueueOrdering.spec.ts` all green, full combined suite 3535 passed / 4 skipped, and the merged tree fails `still re-arms while the queue is actually draining` if the post-loop read is removed. So the cancellation you found is real, and `PatchesSyncFollowUpFlush.spec.ts` does catch it once both are in.

Merge order is now recorded in all three PR descriptions (#145 -> #137 -> #136), including the combined-tree spec run before cutting. Detail in the [thread on #145](https://github.com/dabblewriter/patches/pull/145#discussion_r3777319288).

src/client/OTAlgorithm.ts
```diff
@@ -130,26 +167,43 @@ export class OTAlgorithm implements ClientAlgorithm {
...
-    } else {
-      batch = await this.store.getPendingChanges(docId);
+    const otDoc = doc as OTDoc<any> | undefined;
+    // Only the doc-merge floor needs it, and that branch needs a doc; without one the store read
+    // it would take is discarded, so don't pay for it (the tailRev seed is unused here).
+    const committedRev = otDoc?.committedRev ?? 0;
+    const { pending, withheld } = await this._collectPending(docId, otDoc, committedRev);
+    // Before the empty-queue return: a withheld row is exactly the case where the rest of the
+    // queue can be empty and every other signal reads as fully synced.
+    if (withheld.length > 0) {
```
matalina: The contract implementation is right — store-authoritative, decision written into `_collectPending`, both reasons stated separately. Your argument for `onError` over `hasPending` is better than mine: a doc-aware `hasPending` would change the interface at every call site, and latching true on a row that can never be sent turns a diagnosable failure into a permanent indicator with flushes stacking behind it. Agreed.

**The emit has no dedupe, and the condition never clears.** A row the store lost stays lost, so this fires on every `getPendingToSend` for the life of the doc. There are four call sites in `PatchesSync` — `:747` (syncDoc), `:1052` (flushDoc), `:1175` (the new post-loop read), `:1538` (remote delete) — and a normal sync cycle hits two of them, so one lost row emits **twice per sync, on every sync, indefinitely**.

That is the flood shape this repo has hit before: the 3.0.56 out-of-band repair path put ~6M events/day into telemetry for want of a latch. And both neighbouring cases here already solved it — `PatchesSync._surfacedSyncErrors` (`:268`, a per-doc Set reporting once) and #136 per-entry `surfaced` flag, added during this same review batch for exactly this reason.

The docblock acknowledges at-least-once and pushes dedupe to consumers, but the consumer is app telemetry, which is where the volume costs money and buries the signal. A `Set` keyed on `docId + changeId`, cleared on untrack, matches what the rest of the class already does:

```ts
private _reportedUnstored = new Set<string>();
// ...
const fresh = withheld.filter(c => !this._reportedUnstored.has(`${docId}:${c.id}`));
if (fresh.length > 0) {
  fresh.forEach(c => this._reportedUnstored.add(`${docId}:${c.id}`));
  this.onError.emit(new UnstoredPendingError(docId, fresh.map(c => c.id)), { docId });
}
```

**Separately: this also fires while a remotely-deleted doc is being torn down.** `_handleRemoteDocDeleted` reads pending at `PatchesSync.ts:1538` *before* it closes the doc at `:1541`, so the `if (doc)` branch is still live and any withheld row emits. A collaborator losing access, or a project deleted from another device, would surface a store-integrity error for rows that are about to be discarded either way — and it will read to whoever triages it as data loss at the exact moment a doc vanished. Worth suppressing on the teardown path, or moving that read after the close.

src/net/PatchesSync.ts
```diff
@@ -1165,11 +1170,14 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
             if (fullSnapshot) this._applySnapshotPreservingPending(docId, fullSnapshot, changeBatch);
           }
         }
-
-        // Fetch remaining pending for next batch or check completion
-        pending = (await algorithm.getPendingToSend(docId, this.patches.getOpenDoc(docId) as PatchesDoc<any>)) ?? [];
       }
 
+      // One read of the queue as it now stands, replacing the K per-batch reads this PR removed:
```
matalina, line 1255: This resolves the #145 conflict exactly as needed — verified the placement is after the batch loop closes and before `_resplitBudgets.delete`, so #145 `drained` gate will compare against the queue the flush left. Pre-empting it on this side means #145 merges clean. Good call.

One thing though: the call needs exactly two scalars — `pending[0].id` for the gate and `pending.length > 0` for `_stableDocStatus` — but `getPendingToSend` is not a read. It also:

- emits `UnstoredPendingError` (the second emit per flush; the first is at `:747` before the flush starts — see my note on `OTAlgorithm.ts:196`), and
- runs `_withConsistentBaseRev`, which `console.warn`s on any mixed-baseRev queue. Once #145 lands, mixed-baseRev is the ordinary multi-tab case rather than the torn-reload rarity that warning was written for — so every flush of a mixed queue now warns twice.

My suggestion was the shape, not this specific call. A plain store read gives both scalars with no slicing, no warn, no emit:

```suggestion
      // One read of the queue as it now stands, replacing the K per-batch reads this PR removed:
      // `batches` is split once above, so those never fed the next batch, but a follow-up pass
      // does have to compare against the queue the flush left behind rather than the array it
      // started from (#145 gates its deferred-frame follow-up on the head row having moved).
      // Deliberately the raw store read, not `getPendingToSend`: only the head id and emptiness
      // are wanted here, and the send-path builder also warns on mixed baseRev and reports
      // withheld rows, neither of which should fire twice per flush.
      pending = await algorithm.store.getPendingChanges(docId);
```

If reaching through `.store` from `PatchesSync` is unwanted, a `peekPendingHead(docId)` on `ClientAlgorithm` states the intent better than reusing the send-path builder as a status probe.

tests/net/PatchesSyncQueueOrdering.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+    const reported: UnstoredPendingError[] = [];
+    sync.onError(err => {
+      if (err instanceof UnstoredPendingError) reported.push(err);
+    });
+
+    sync['updateState']({ connected: true });
+    await sync['syncDoc'](DOC_ID);
+
+    expect(sentLabels(conn)).toEqual(['later']);
+    expect(reported.map(e => e.changeIds)).toContainEqual([orphan.id]);
```
matalina: The test drives the right scenario and I like that you verified it fails in both directions. One gap: `toContainEqual` passes for one report or twenty, so the emit-count behaviour is invisible here — and that is the property actually at risk.

This single `syncDoc` reaches `getPendingToSend` twice (`PatchesSync.ts:747` and the new `:1175`), so `reported` most likely already has two entries and the assertion passes regardless. Which means the duplication `b230f0d` introduced shipped without the suite noticing.

```suggestion
    expect(reported).toHaveLength(1);
    expect(reported[0].changeIds).toEqual([orphan.id]);
    expect(reported[0].docId).toBe(DOC_ID);
```

That should fail today, which is the point. Worth a second `syncDoc` after it asserting the count still does not grow, since "reports once per doc, not once per flush" is the contract that keeps this out of the flood category.

src/client/OTAlgorithm.ts
```diff
@@ -573,19 +644,22 @@ export class OTAlgorithm implements ClientAlgorithm {
...
     if (doc) {
       const inMemory = doc.getPendingChanges();
       const latestRev = storePending[storePending.length - 1]?.rev ?? committedRev;
       const storedIds = new Set(storePending.map(c => c.id));
-      const docOnly = inMemory.filter(c => c.rev > latestRev && !storedIds.has(c.id));
-      if (docOnly.length > 0) pending = [...storePending, ...docOnly];
+      const docOnly = inMemory.filter(c => !storedIds.has(c.id));
+      const merged = docOnly.filter(c => c.rev > latestRev);
+      if (merged.length > 0) pending = [...storePending, ...merged];
+      withheld = docOnly.filter(c => c.rev <= latestRev);
```
matalina, line 707: Precision point on the classification, not a bug I can demonstrate.

`withheld` is "doc-only and at or below `latestRev`", but `latestRev` falls back to `committedRev` when the store queue is empty — a different quantity. With store rows present, "at or below" genuinely means the store moved past this row. With an empty store queue it means "at or below the doc own committed rev", and such a row is not a lost durable row at all, it is a stale mirror entry for something already committed.

I could not construct a reachable case: `OTDoc.applyChanges` re-sequences pending strictly above the new `committedRev`, and `dropResolvedPending` re-syncs the open doc when it drops rows, so both windows look closed.

Raising it because the error now asserts a specific cause — "a store write that reported success without persisting" — and that claim only holds under the first reading. Worth either splitting the two cases or noting on `withheld` that the empty-store fallback is believed unreachable and why, so a later change to the rev-sequencing invariant does not quietly turn this into a misreported cause.

src/client/OTAlgorithm.ts
```diff
@@ -36,6 +37,35 @@ function firstGapIndex(changes: Change[]): number {
...
+ * putting it on the wire would let the server hold content no local rebuild reproduces. The
+ * silence is not — the editor still shows that content, and `hasPending` reads the store, so
+ * without this the doc reports itself fully synced while a change sits unsendable. A doc holding
+ * a change the store lacks is a store-integrity failure (the known producer is a store write that
+ * reported success without persisting), and this reports it as one.
+ *
+ * At-least-once: emitted on every flush that observes the row, so consumers should key on
+ * `docId` + `changeIds`.
+ */
+export class UnstoredPendingError extends Error {
```
matalina: Placement: `CommittedPoisonSkippedError` from #136 lives in `src/net/error.ts` with the rest of the error surface, and this one lives here. Both are types an app is expected to `instanceof` off the same `PatchesSync.onError` signal, so after the release a consumer discriminating between them writes two different import paths for no reason they can infer.

Not broken — `src/client/index.ts:12` does `export * from './OTAlgorithm.js'`, so it is reachable from `@dabble/patches/client`. I checked before raising it.

Moving it next to its sibling costs one import and keeps the error surface discoverable in one place. Low priority, but the two PRs land together so it is the cheap moment to do it.

src/net/PatchesSync.ts
```diff
@@ -217,6 +217,11 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
       // malformed server-pushed event the transport had to drop) so they reach the
       // app's telemetry instead of vanishing.
       ...(this.connection.onError ? [this.connection.onError(error => this.onError.emit(error))] : []),
+      // Same for algorithm-level failures no caller can observe — OT withholding a pending change
+      // the store never persisted is reported here and nowhere else.
+      ...Object.values(patches.algorithms).flatMap(algorithm =>
+        algorithm?.onError ? [algorithm.onError((error, context) => this.onError.emit(error, context))] : []
```
matalina, line 274: Worth a comment rather than a change: this builds the forwarding subscriptions once, from `Object.values(patches.algorithms)`, in the constructor list — so an algorithm registered after `PatchesSync` is constructed would have its `onError` silently unforwarded.

In practice `Patches` takes `{ algorithms: { ot, lww } }` at construction, so the map is complete before this runs and it is correct today. The reason to note it is that the guarantee lives in a different class, and the failure mode if it ever stops holding is silence — an algorithm reporting integrity failures nobody receives, which is the exact condition this commit exists to remove.

One line saying `patches.algorithms` is fixed at construction would stop a future lazy-registration change from quietly reopening it.

src/client/OTAlgorithm.ts
jacwright: Latched in 36abe25. `_reportedUnstored` is a `Map<docId, Set<changeId>>` rather than one composite-keyed Set so the untrack clear is a delete instead of a prefix scan; cleared in `untrackDocs` and `confirmDeleteDoc`, so a row reports once per doc lifetime.

Teardown suppressed too: `_handleRemoteDocDeleted` now calls `getPendingToSend(docId)` with no doc, so the mirror is never classified on the way out. Same payload as moving the read after the close would produce (post-close `getOpenDoc` is undefined), but it doesn't depend on statement order surviving a later edit.

src/net/PatchesSync.ts
jacwright: Went with `peekPendingHead` (36abe25). The raw store read doesn't typecheck here: `algorithm` is a `ClientAlgorithm`, whose `store` is a `PatchesStore` — `getPendingChanges` lives on `OTClientStore` — and `flushDoc` runs for LWW docs too, where a durable change queue doesn't exist to read. So it's an optional method on `ClientAlgorithm` alongside `dropResolvedPending`/`replacePendingChanges`, OT-only, returning the head row or null; when an algorithm doesn't implement it the flush keeps the batch it sent (only `_stableDocStatus` reads it on that path). No warn, no emit, and #145 gets its head id.

tests/net/PatchesSyncQueueOrdering.spec.ts
jacwright: Strengthened as suggested, but the strict form does **not** fail on the pre-fix code — I ran it first. That single `syncDoc` reports once: the commit echo re-syncs the open doc from the store before `:1175` reads it, so the mirror no longer holds the orphan and the second read classifies nothing. Same `OTDoc.applyChanges` re-sequencing your `_collectPending` note cites.

The repeat needs the condition to survive the flush, so the count assertion moved to its own test where the commit dies status-lessly and the queue can't drain: three sync attempts give three reports pre-fix (`expected [ …(3) ] to have a length of 1`) and one after the latch. Both directions verified in 36abe25.

src/client/OTAlgorithm.ts
jacwright: Took the second option in 36abe25: the `_collectPending` docblock now states that the `latestRev` fallback to `committedRev` makes a doc-only row at or below it a stale mirror entry rather than a lost durable row, cites `OTDoc.applyChanges` re-sequencing pending strictly above the new committedRev and `dropResolvedPending`'s re-sync as why no producer is known, and says to split the cases rather than let the error keep asserting a cause it can't prove if that invariant changes.

src/client/OTAlgorithm.ts
jacwright: Moved to `src/net/error.ts` in 36abe25, at the same anchor #136 uses (right after `isUnsplittableChangeError`), so the two land adjacent — expect a trivial add/add conflict for whichever of us merges second. `src/client/index.ts` re-exports it next to `UnsplittableChangeError`, the pattern already there, so `@dabble/patches` and `/client` still resolve it and `/net` now does too.

src/net/PatchesSync.ts
jacwright: Added in 36abe25.

src/net/PatchesSync.ts
```diff
@@ -1533,9 +1547,13 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
...
-    // Get pending changes before cleanup so app can handle them
-    const pendingChanges =
-      (await algorithm.getPendingToSend(docId, this.patches.getOpenDoc(docId) as PatchesDoc<any>)) ?? [];
+    // Get pending changes before cleanup so app can handle them. Deliberately without the open
+    // doc: its mirror is discarded by the close below either way, and classifying it here would
+    // report a row the store lost (`UnstoredPendingError`) at the exact moment the doc
+    // legitimately vanished — a store-integrity alarm for a collaborator losing access or a
+    // delete from another device. Passing no doc reads the durable queue, which is all this
+    // payload can meaningfully carry.
+    const pendingChanges = (await algorithm.getPendingToSend(docId)) ?? [];
```
matalina: The alarm suppression is right, and doing it by argument rather than by statement order is the better of the two options I offered. But dropping the doc also drops content from the payload.

`pendingChanges` here is the **shelving hook** — `_emitRemoteDocDeleted` hands it to `onRemoteDocDeleted`, and the comment below says it is awaited so *an app that shelves those changes has landed the write before we proceed*.

Before this commit, `getPendingToSend(docId, openDoc)` returned the send batch, which includes doc-only rows above the store tail (`merged` in `_collectPending`). Passing no doc reads only the durable queue, so those rows are no longer in the payload.

That inverts the priority: the content most at risk is the content the store never accepted, which exists nowhere but the open doc being closed two lines further down. A collaborator losing access mid-edit now shelves strictly less than they did before this commit.

Suppressing the report rather than the doc keeps both properties:

```ts
// in OTAlgorithm
async getPendingToSend(docId: string, doc?: PatchesDoc<any>, options?: { report?: boolean }): Promise<Change[] | null>
// …and here:
const pendingChanges = (await algorithm.getPendingToSend(docId, this.patches.getOpenDoc(docId) as PatchesDoc<any>, { report: false })) ?? [];
```

Or read `_collectPending` directly on this path and hand the app `[...pending, ...withheld]` — arguably the most honest payload, since on a doc that is about to vanish the distinction between durable and not stops mattering and the user just wants their words back.

src/client/OTAlgorithm.ts
```diff
@@ -130,26 +148,49 @@ export class OTAlgorithm implements ClientAlgorithm {
...
+      if (fresh.length > 0) {
+        fresh.forEach(id => reported.add(id));
+        this._reportedUnstored.set(docId, reported);
+        this.onError.emit(new UnstoredPendingError(docId, fresh), { docId });
+      }
     }
-    if (batch.length === 0) return null;
-    return this._withConsistentBaseRev(docId, batch);
+    if (pending.length === 0) return null;
+    return this._withConsistentBaseRev(docId, pending);
```
matalina, line 198: Good correction on the typing — you are right that `ClientAlgorithm.store` is a `PatchesStore` and that `flushDoc` runs for LWW, so my raw-store-read suggestion would not have compiled. The optional-method shape alongside `dropResolvedPending` is the right home for it.

One thing about the implementation: the interface promises *a plain read* for callers that need only the head row, but this materializes the entire pending queue and discards all but `[0]`. `OTClientStore.getPendingChanges` takes only `{ startAfterRev }` — there is no limit.

Still a big win over K per-batch reads. But on the population where it matters it is not a peek: a user coming back online with a large offline queue pays a full IndexedDB scan of the pending table to learn one id, per flush, on the devices already showing multi-second write stalls (DAB-766).

Worth closing now rather than later, because every future caller will reasonably read `peek` as O(1):

```ts
// OTClientStore
getPendingChanges(docId: string, options?: { startAfterRev?: number; limit?: number }): Promise<Change[]>;

// OTAlgorithm
async peekPendingHead(docId: string): Promise<Change | null> {
  const [head] = await this.store.getPendingChanges(docId, { limit: 1 });
  return head ?? null;
}
```

src/net/PatchesSync.ts
```diff
@@ -1165,11 +1172,18 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
...
 
+      // One read of the queue as it now stands, replacing the K per-batch reads this PR removed:
+      // `batches` is split once above, so those never fed the next batch, but a follow-up pass
+      // does have to compare against the queue the flush left behind rather than the array it
+      // started from (#145 gates its deferred-frame follow-up on the head row having moved).
+      // A peek, not `getPendingToSend`: only the head row and emptiness are wanted here, and the
+      // send-path builder would warn on mixed baseRev and re-report rows the store lost as it
+      // went. An algorithm with no durable queue to peek keeps the batch it flushed.
+      const head = await algorithm.peekPendingHead?.(docId);
+      if (head !== undefined) pending = head ? [head] : [];
```
matalina, line 1263: Latent until #145 lands, but worth fixing before it does, because this is the line that makes its gate work.

`peekPendingHead` reads the **store**. #145 captures `headBefore` from the flush-start **send batch** (`pending[0].id`), and those are not always the same collection: `getPendingToSend` returns `_withConsistentBaseRev(_collectPending(...).pending)`, and when the store queue is empty that array is `[...[], ...merged]` — its `[0]` is a doc-only row the store never held.

So `headBefore` can be an id `peekPendingHead` cannot return, making `drained` true on any non-empty store queue whether or not anything drained — a foreign row arriving mid-flush reads as progress. The follow-up that fires is serial-gated and harmless in itself, but it defeats exactly the anti-spin property the gate was added for.

Cheapest fix is to have both ends read the store, which also removes the coupling to whatever `getPendingToSend` decides to put at the front:

```ts
// before the batch loop, in place of #145 `const headBefore = pending[0].id`
const headBefore = (await algorithm.peekPendingHead?.(docId))?.id;
```

Worth agreeing which PR carries it so it does not fall between the two — happy either way, but this one already owns `peekPendingHead`.

src/client/OTAlgorithm.ts
```diff
@@ -63,6 +72,15 @@ export class OTAlgorithm implements ClientAlgorithm {
...
   private readonly _docLocks = new Map<string, Promise<unknown>>();
 
+  /**
+   * Change ids already reported as unstored, per doc (see {@link UnstoredPendingError}). The store
+   * cannot recover a row it never took, so the condition never clears: without this the same row
+   * is reported on every read of the queue, for the life of the doc. Mirrors
+   * `PatchesSync._surfacedSyncErrors`. Cleared when the doc is untracked or deleted — the next
+   * tracking of that doc is a fresh lifetime.
+   */
+  private readonly _reportedUnstored = new Map<string, Set<string>>();
```
matalina, line 87: Latch looks right, and `Map<docId, Set<changeId>>` for an O(1) untrack clear rather than a composite-key prefix scan is a better shape than what I suggested. Clearing in both `untrackDocs` and `confirmDeleteDoc` covers the lifetimes.

Small note: this has no per-doc bound, while the memo it cites as its model caps at `MAX_POISON_MEMO_ENTRIES` (50). A doc whose store keeps losing rows adds an id per loss and never evicts while it stays tracked.

Much milder than the case #136 bounded — short id strings, not retained `ApplyChangesError` graphs carrying op payloads — so this is a note, not a defect. Raising it only because the docblock says it mirrors `_surfacedSyncErrors` and the nearest analogue in the same release chose a cap; worth either matching it or saying why unbounded is fine here, so the difference reads as a decision.

jacwright:
Merged main (with #145's final form) into this branch: c4760b1. The only conflicts were this branch's pre-#145 docblock and warn text on _withConsistentBaseRev — resolved to main's flush-one-frame version wholesale. The interlock #145 depends on survives: the post-loop head peek this branch kept feeds main's headBefore/drained follow-up gate in PatchesSync.flushDoc. Full gate green: 3537 tests, type/lint/format clean.

matalina (changes requested):
Reviewed the merge (`c4760b1`) and re-checked the four open threads against the merged tree. One of them blocks; the other three don't, and I should have said so explicitly the first time — all my reviews here have been plain comments, so nothing on this PR has actually marked what stands between it and an approval. That's on me, and it's presumably why this looks ready from your side.

**The merge is good.** I checked what the resolution dropped rather than what it added:

- `OTAlgorithm-getPendingToSend.spec.ts` is the risky one — both PRs rewrote it — and the resolution is a genuine three-way merge, not a wholesale take. #145's three slicing tests came in, *and* this branch's docstring rewrite survived on top (store-as-truth, pointing at `_collectPending`, R3a label dropped). The only test that disappeared is `re-stamps a stale baseRev in the pending queue to committedRev`, which asserts the relabel #145 removed — correct to drop.
- Every branch contribution survives: `UnstoredPendingError`, `_reportedUnstored`, `peekPendingHead`, `withheld`, `_collectPending`.
- `_withConsistentBaseRev` is main's flush-one-frame version, which is right — #145 owns that function now.

---

### Blocking

**[`PatchesSync.ts:1572`](https://github.com/dabblewriter/patches/pull/137#discussion_r3786902007) — the remote-delete payload.** Suppressing the teardown alarm by dropping the open doc also drops doc-only rows from `pendingChanges`, and that payload is the shelving hook: `_emitRemoteDocDeleted` hands it to `onRemoteDocDeleted`, awaited so an app that shelves has landed the write first.

So the content most at risk — what the store never accepted, which exists nowhere but the doc being closed two lines later — is exactly what no longer reaches the hook. A collaborator losing access mid-edit shelves less than they did before this PR.

Suppressing the *report* rather than the *doc* keeps both: an options argument on `getPendingToSend` (`{ report: false }`), or reading `_collectPending` here and handing over `[...pending, ...withheld]`.

### Not blocking — happy for these to be follow-ups

1. **[`headBefore` vs `peekPendingHead`](https://github.com/dabblewriter/patches/pull/137#discussion_r3786902158)** read different collections. Your merge note says the interlock survives, and mechanically it does — the peek does feed the gate. What doesn't hold is that both ends read the same thing: `headBefore` (`:1073`) comes from the send batch, which can begin with a doc-only row the store never held, so `drained` can read true when nothing drained. Latent before the merge, live now. Two lines: `const headBefore = (await algorithm.peekPendingHead?.(docId))?.id`.
2. **[`peekPendingHead`](https://github.com/dabblewriter/patches/pull/137#discussion_r3786902073)** materializes the whole queue to return `[0]` — fine at one read per flush, but a method called `peek` will be assumed O(1) by whoever touches it next.
3. **[`_reportedUnstored`](https://github.com/dabblewriter/patches/pull/137#discussion_r3786902236)** has no per-doc cap where the memo it cites as its model has one. Note, not a defect.

Fix the first one and I'll approve — the merge and the latch work are both solid, and this is the last thing holding the patches release.

src/net/PatchesSync.ts
jacwright: Fixed in d9a9000 with your first shape: the delete path passes the open doc with { report: false }, so the doc-only rows are back in the payload and no store-integrity alarm fires while the doc vanishes. The silent read also leaves the once-per-doc latch alone — a doc that lives on still reports on its next send. I kept rows the tail passed excluded (same rule as the wire) rather than handing out [...pending, ...withheld]; if you want those in the shelf too I'd surface _collectPending or add a dedicated method — say the word and it can ride this PR. Both delete shapes are pinned in PatchesSyncQueueOrdering.

src/client/OTAlgorithm.ts
jacwright: Done in d9a9000: getPendingChanges takes { limit } — the native getAll count in the IndexedDB store, a slice in memory — and peekPendingHead reads { limit: 1 }. Tests pin the count passthrough and the single-row read.

src/net/PatchesSync.ts
jacwright: Done in d9a9000, carried in this PR as you offered: headBefore now comes from peekPendingHead when the algorithm has one, batch-to-batch otherwise, so both ends of the drained compare always read the same collection. One observation from implementing it: I could not construct a scenario where the old capture actually misfires — a doc-only batch head requires an empty store queue, so any non-empty post-flush queue means a mid-flight arrival, and the follow-up that triggers is productive (it flushes that row) and terminates on the next pass under either baseline. The change is real value anyway — it removes the coupling to whatever getPendingToSend fronts — but if you can see a spin shape I missed I will pin it in a test.

src/client/OTAlgorithm.ts
jacwright: Comment added in d9a9000 making it a decision: entries are short id strings, growth requires the store to keep losing rows on one tracked doc, and evicting an id resumes the per-attempt alarm this latch exists to stop — lifecycle clearing (untrack/delete) is the bound, unlike the poison memo's retained error graphs.

src/net/PatchesSync.ts
matalina: Verified — the doc is passed with `{ report: false }`, and gating the emit on `report !== false` also leaves `_reportedUnstored` untouched, so a doc that survives still reports on its next send. That's the shape I wanted.

**On your question — yes, please, if it can ride this PR.**

Your reasoning for excluding `withheld` is coherent: same rule as the wire. But I think the wire rule and the shelf rule are answering different questions. The wire rule is about *durability and authority* — don't let the server hold what no local rebuild reproduces. The shelf is about *not silently destroying user-visible content*, and on a doc being torn down those come apart.

The sharper version: `pending` rows survive without the shelf — they're in the store, and if the doc weren't being deleted they'd be retried. `withheld` rows are by definition unsendable and exist only in a mirror that is discarded four lines later. So the one category where the shelf is the *sole remaining copy* is the one category it currently excludes.

A dedicated method reads better than widening `getPendingToSend`, since this isn't a send:

```ts
/** Everything the doc still holds, durable or not — for a doc being discarded. */
async collectUnsyncedForDiscard(docId: string, doc?: PatchesDoc<any>): Promise<Change[]>
```

Not blocking either way — the current state restores exactly what the payload carried before this PR, so nothing regressed. Happy for it to be a follow-up if you'd rather keep this PR's diff where it is.

src/net/PatchesSync.ts
matalina: You're right and I overstated it — thanks for pushing back rather than just applying it.

I went looking for the spin shape and can't construct one either. A doc-only batch head does require an empty store queue, so a non-empty post-flush queue means something arrived mid-flight, and the follow-up that fires *flushes that row* — productive work, and the next pass terminates under either baseline. Calling it "defeats the anti-spin property" was wrong; the gate misnames the reason but the behaviour was already correct.

So no test to pin — there's nothing to pin. The change stands on the coupling argument you gave: `headBefore` no longer depends on whatever `getPendingToSend` happens to front, which is the property that would have made this fragile as `_collectPending` evolves.

One note on the new line: with a peek available and an empty pre-flush store queue, `headBefore` is `undefined`, so `pending[0].id !== undefined` makes `drained` true for any non-empty post-flush queue. That's the mid-flight-arrival case again — productive, terminates — so it reads correct to me, just worth knowing it's load-bearing that `undefined` is a distinct value here rather than something to normalise to `null` later.

matalina (approved):
Approving — the blocker is fixed and I verified all four rather than reading the commit message.

- **Shelf payload** (`PatchesSync.ts:1579`): the delete path passes the open doc with `{ report: false }`, so doc-only rows are back in `pendingChanges`, and gating the emit on `report !== false` leaves `_reportedUnstored` untouched — a doc that survives still reports on its next send. Exactly the shape I asked for.
- **`headBefore`** (`:1079`): now `peekPendingHead` when the algorithm has one, batch-to-batch otherwise, so both ends of the `drained` compare read the same collection.
- **`peekPendingHead`** (`OTAlgorithm.ts:203`): `{ limit: 1 }`, with `limit` added to the `OTClientStore` contract and backed by the native `getAll` count rather than a memory slice.
- **`_reportedUnstored`**: documented as a decision, with a reason I hadn't considered — evicting an id *resumes* the per-attempt alarm the latch exists to stop, so a cap would reintroduce the flood it prevents. That's a better answer than matching the poison memo, and it's now written down.

Two things worth recording.

**You were right about `headBefore` and I was wrong.** I went looking for the spin shape you asked about and can't construct one either — a doc-only batch head requires an empty store queue, so a non-empty post-flush queue means a mid-flight arrival, and the follow-up that fires flushes that row and terminates. "Defeats the anti-spin property" overstated it. The change stands on the coupling argument you gave, not on a bug.

**One open question, not blocking:** whether `withheld` rows belong in the shelf payload too — detail in that thread. Short version: the wire rule and the shelf rule answer different questions, and `withheld` rows are the only category where the shelf is the sole remaining copy. Happy for it to ride this PR or follow; the current state restores exactly what the payload carried before, so nothing regressed either way.

Also worth saying: my three earlier rounds all went in as plain comments, so nothing marked what was blocking until yesterday. That's why this sat — the delay was mine, not yours.

This is the last of the three patches PRs. #136 is approved and clean, #145 is in.

src/net/PatchesSync.ts
jacwright: Agreed on the category argument — withheld rows are the one class whose sole remaining copy is the mirror being discarded, so the wire rule doesn't transfer. Since the approval landed and #136/#137 merged as a pair, I'm doing it as the follow-up you offered: `collectUnsyncedForDiscard(docId, doc?)` as a dedicated method (not a widening of `getPendingToSend`), used by `_handleRemoteDocDeleted`. PR shortly.

src/net/PatchesSync.ts
jacwright: Thanks for closing the loop. Good point on `undefined` being load-bearing in the `drained` compare — adding a one-line comment saying it must not be normalised to `null`, riding the follow-up PR rather than reopening this one.

src/net/PatchesSync.ts
jacwright: Opened #148 with the dedicated method, the LWW both-halves read, and the `undefined` comment.

PR #139, opened 2026-08-06
fix(ot): estimate split points instead of compressing at every boundary (DAB-931)

matalina:
Fixes the crash behind DAB-931: whole-project **Duplicate** and **Time Machine full-project restore** take the browser down on large projects.

> **Rebased onto `main` (0.26.0).** The first version of this PR also carried a retain fix and a DAB-760 guard rework. Both had already landed in `c3c5018` — I had branched off a stale local `main` and never fetched. They are gone from the diff; what remains is the estimator alone, re-applied on top of `main`'s `pieceStartPos` implementation. See the review thread for the full accounting.

## Root cause

`breakSingleChange` measured storage size by **actually compressing the accumulated payload once per op** — and again per delta op inside an oversized text op — making a split `O(ops × payload)` LZ-string passes. Both operations emit a single op carrying the entire project body, so both land straight in it.

Under the threshold nothing splits and one compression happens; over it, the splitter re-compresses megabytes repeatedly **on the main thread**. That is the "Page unresponsive" / tab crash users report. It is a cliff, not a slope.

## The fix

Measure the change **once**, derive how many storage bytes one JSON byte costs, and answer the hot "does this still fit?" questions from per-op JSON sizes computed once each (memoised in a `WeakMap`).

The estimate only chooses *where* to cut. Two things keep being wrong from mattering:

- **`verifyPieces`** re-measures every emitted piece with the real calculator and re-splits it under a tightened budget if the estimate ran optimistic (max 3 passes, 0.8× per pass). A bad ratio costs an extra split, never a change over the store's cap.
- **Where a decision has a side effect** rather than just moving a boundary — `guardUnsplittable` reports and refuses on a *measured* size — the estimate only decides whether to look closer, and the real calculator decides what to do. One real measurement per oversized op is bounded; it was measuring on *every* op that made this quadratic.

With no `sizeCalculator` the ratio is exactly 1 and this reproduces `getJSONByteSize` byte for byte, so the uncompressed path keeps its previous split boundaries exactly.

## Measured — same fixture, same machine

A 4.2 MB whole-document change (400 docs of prose), split against `maxStorageBytes: 900_000` + `compressedSizeUint8`:

| | `origin/main` | this branch |
| -- | -- | -- |
| one split | **65.6 s** — blew the test's own 60 s ceiling | **3.9 s** |

Taken by checking `main`'s `changeBatching.ts` into this worktree and running the new timing test against it, so both sides measure the identical fixture.

## Two changes beyond a straight port

Worth review attention, since neither is purely mechanical:

**1. `resplit` suppresses duplicate oversized-op telemetry.** Caught by `main`'s own test (`defaults to Infinity through breakChangesIntoBatches`): a `verifyPieces` re-split walks ops the first pass already reported, so it emitted a second `onOversizedOp` for the same op. Now suppressed on re-split passes. **Refusal is unaffected** — `maxUnsplittableBytes` does not move between passes, so the emit/refuse decision is identical either way; only the duplicate report is dropped.

**2. `splitLargeInsertText` seeds its chunk length through the estimator.** It previously treated a possibly-compressed byte budget as a character count. Under compression that seed is several times too short and emits far more pieces than the budget needs. It remains only a seed — every chunk is still measured and halved, and surrogate-pair safety is untouched.

## Tests

New `tests/algorithms/ot/shared/changeBatchingEstimator.spec.ts` (7 tests) pinning the estimator's own contract:

- every emitted piece lands within the storage cap
- `verifyPieces` re-splits rather than emitting, driven by a deliberately optimistic calculator
- the document survives replay of the pieces
- revs stay contiguous across re-splits, and the first piece keeps the original id
- the split is deterministic
- the uncompressed path keeps its boundaries
- a multi-megabyte document splits without the quadratic cost

Positioning and replay-order correctness are already covered by `changeBatchingApply.spec.ts` from `c3c5018`; I dropped my duplicates of those rather than re-land them.

Full suite: **3392 passed / 4 skipped**. `type:check`, `lint`, `format:check` clean. Three `tests/solid/*` suites fail to load on `@solid-refresh` — **pre-existing, fails identically on clean `main`**.

## Review guidance

- `PIECE_COMPRESSION_MARGIN` (1.1) is a deliberate over-estimate: a piece compresses slightly worse than the change it came from, because LZ builds a smaller dictionary over less input. Since `verifyPieces` is authoritative, being too optimistic costs a re-split pass — but being too *pessimistic* costs extra seams, which is invisible in tests (they pin piece counts as deterministic, not their magnitude) and shows up as revision-count inflation on a large split, a branch seed most of all. The comment says so; prefer nudging it down rather than up.
- The estimator is derived per change, not shared through `SplitContext` — `breakLargeValueOp` builds a different change and needs its own ratio.

## Not in this PR

1. **Guard rail (dw3).** A cheap `JSON.stringify().length` pre-flight probe with a progress modal. **I'd argue for this regardless of how fast the splitter gets**: measuring two live accounts, one user's project sits at **95% of the threshold** (856,974 B of 900,000) with no warning of any kind. Observed compression ratio across six real projects is ~0.28, so the threshold maps to roughly **3.2 MB uncompressed** for a probe constant.
2. **Per-doc ops instead of one giant op (dw3).**
3. **Chunked commit + yielding (dw3).**
4. **Server-side duplicate (pup/rest).**

With this landed, 2 and 3 are optimizations rather than necessities. 1 stands on its own.

## Shipping path

`main` is at 0.26.0; this needs a merge, then a **publish**. dw3 pins **0.25.3** and pup pins **0.25.2**, so both need pin bumps before any of this reaches a user.

### DIFF
--- a/src/algorithms/ot/shared/changeBatching.ts
+++ b/src/algorithms/ot/shared/changeBatching.ts
@@ -102,6 +102,14 @@ interface SplitContext extends Required<Pick<ChangeSplitOptions, 'maxUnsplittabl
   docId?: string;
   changeId?: string;
   budget: 'storage' | 'payload';
+  /**
+   * Set on a {@link verifyPieces} re-split. The first pass already reported every op it had no
+   * seam for, and a re-split walks those same ops again under a tighter budget — emitting again
+   * would double-count one op in the consumer's telemetry. Refusal is unaffected:
+   * `maxUnsplittableBytes` does not move between passes, so the emit/refuse decision is identical
+   * either way.
+   */
+  resplit?: boolean;
 }
 
 /**
@@ -154,24 +162,28 @@ function guardUnsplittable(
   reason: OversizedOpReason
 ): void {
   const emitted = bytes <= ctx.maxUnsplittableBytes;
-  onOversizedOp.emit({
-    op,
-    path,
-    bytes,
-    maxBytes: ctx.maxBytes,
-    budget: ctx.budget,
-    docId: ctx.docId,
-    changeId: ctx.changeId,
-    emitted,
-    reason,
-  });
+  if (!ctx.resplit) {
+    onOversizedOp.emit({
+      op,
+      path,
+      bytes,
+      maxBytes: ctx.maxBytes,
+      budget: ctx.budget,
+      docId: ctx.docId,
+      changeId: ctx.changeId,
+      emitted,
+      reason,
+    });
+  }
   if (!emitted) {
     throw new UnsplittableChangeError(op, path, bytes, ctx.maxUnsplittableBytes, {
       docId: ctx.docId,
       changeId: ctx.changeId,
     });
   }
-  console.warn(`Oversized op ${op} at "${path}" is ${bytes} bytes and cannot be split; including it anyway`);
+  if (!ctx.resplit) {
+    console.warn(`Oversized op ${op} at "${path}" is ${bytes} bytes and cannot be split; including it anyway`);
+  }
 }
 
 /** Default wire batch size limit (1MB) */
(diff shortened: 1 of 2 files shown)


jacwright (changes requested):
Requesting changes — but only on packaging, not on the idea. The estimator is the right fix, it's well built, and I verified the headline number. The problem is that changes (2) and (3) already landed on `main` five days ago, and the evidence offered for them being pre-existing doesn't hold.

## The estimator (1) is excellent and I confirmed the win

I ran your round-trip test on unmodified `origin/main` and on this branch:

|                                                                                | `origin/main`  | this branch  |
| ------------------------------------------------------------------------------ | -------------- | ------------ |
| `reproduces the original document when the split pieces are replayed in order` | **224,185 ms** | **3,748 ms** |

~60× on the same test, same machine. The cliff is real and this removes it. `verifyPieces` re-measuring with the real calculator is what makes an estimate safe to ship, and gating the whole thing behind ratio-1 when there's no `sizeCalculator` keeps the uncompressed path byte-identical. 52/52 pass here.

## (2) and (3) are already on `main`

> Verified pre-existing by running the new round-trip tests against **unmodified `origin/main`**, where they fail identically.

They don't fail. I copied this spec onto a clean `origin/main` worktree (`b59eae9`) and ran exactly those two tests: **both pass.** That's the 224s run above — slow, because `main` still has the quadratic measurement, but correct.

The reason is `c3c5018 fix(ot): position split @txt pieces with cumulative retains`, one of the **7 commits this branch is behind**. Its message is your (2) and (3), in that order:

> every piece after the first was emitted without a leading retain, so on replay its ops composed at the head of the document - a multi-chunk insert reassembled scrambled. Track the position each piece's ops act at … and prefix every later piece with that retain
>
> The DAB-760 merge-duplication guard pattern-matched the old retain-less shape … The guard now arms on retain-prefixed inserts and detects that adjacent shape alongside the legacy head-anchored one

On `main` today `breakTextOp` carries `pieceStartPos` / `withStartRetain` / `advanceOf` — the same mechanism as your `resumeAt` / `deltaOpOutputLength`. `retainToPrefixCurrentPiece` is gone, and `leadingInsertText` no longer exists anywhere in `src/`. `c3c5018` touched both files you touched, including 72 lines of `OTBranchManager.ts`, and shipped with its own tests (`changeBatchingApply.spec.ts` +169, `OTBranchManager.integration.spec.ts` +47). It went out in **0.26.0**, which is on npm now — this branch is based on 0.25.4, which is also why the shipping section says `main` is at 0.25.4.

So the bug was real when you branched. You found it independently and fixed it correctly. It just isn't yours to fix any more, and re-landing it will conflict in both files.

## What I'd do

Rebase onto `main` (0.26.0) and drop (2) and (3) entirely, then re-apply the estimator on top of `main`'s `pieceStartPos` implementation. That's real work — your `breakTextOp` rewrite is interleaved with the resume tracking — but it shrinks this PR to the part that's genuinely new, and removes the risk of the merge silently reverting `c3c5018`.

One thing to compare rather than assume while you're there: `main`'s guard keeps the **legacy head-anchored** detection deliberately, "which branches seeded by shipped splitters still need." Your `findDuplication` keeps a head-anchored path too (the `if (segment.original) break;` loop), so I don't think anything is lost — but that's the one place where picking your version over `main`'s could quietly narrow coverage for branches seeded by already-released splitters, so it's worth diffing rather than eyeballing.

## Scope — you stayed inside it

For the record, since this was the thing to check: this is step 1 as agreed. Cheap JSON-byte estimate for the split-point search, one real compression per emitted piece, measured before/after on real project sizes. **No worker, no progress indicator**, and server-side duplicate and per-doc ops are correctly listed under "Not in this PR" rather than built. Good.

Your argument for guard rail (1) in that list is the strongest thing in the section and I'd back it: a live account sitting at **95% of the threshold** (856,974 of 900,000) with no warning is a cliff a user walks off with no way to know. That stands on its own regardless of how fast the splitter is, and it survives this rebase untouched.

## Small

`PIECE_COMPRESSION_MARGIN` at 1.1 with a 0.8× tightening and max 3 passes is a sensible envelope, and since `verifyPieces` is authoritative a wrong guess costs a split rather than correctness. Worth a one-line note on the constant saying what happens if it's badly wrong in the _pessimistic_ direction — over-splitting a 7 MB change into many more pieces than needed is invisible in tests but shows up as revision-count inflation on a branch seed, which `splits deterministically` pins the count of but not the magnitude.

matalina:
You were right on all of it. Rebased onto `main` (0.26.0) and dropped (2) and (3) — force-pushed, now `c981557`, **2 files / +406−51** where it was 3 files / +487−97.

## How I got it wrong

I checked **local** `main` and never ran `git fetch`. Local was sitting at `5adafac`/0.25.4, so `main..HEAD` came back empty and I read that as "nothing upstream to worry about". `origin/main` was 7 commits ahead the whole time.

Worse, the "verified pre-existing against unmodified `origin/main`" line was not something I ran — I carried it over from the ticket's earlier write-up and repeated it as if it were my own check. That's the part I'd most like back: it's the sentence that made the retain fix look load-bearing, and it was doing the opposite of verification. Confirmed your version now — `c3c5018` is on `origin/main`, `pieceStartPos`/`withStartRetain`/`advanceOf` are all there, and `retainToPrefixCurrentPiece` and `leadingInsertText` are gone from `src/` entirely.

## The guard question — diffed, not eyeballed

You were right to flag it, and the answer is that picking mine would have **narrowed** coverage. `main`'s `findInsertedDuplication` keeps both shapes:

- **shape A** — head-anchored, the `if (segment.original) break;` loop, for branches seeded by already-shipped splitters
- **shape B** — position-correct, `followingOriginal.startsWith(segment.text)`: the surviving original resumes with the inserted run, verbatim

Mine had a head-anchored loop and a *generic neighbour-window* check — looser than B, which pins the exact relationship. And `main` additionally has `coalesceSegments`, so formatted spans arriving as several sub-`minLength` fragments still get caught; mine had nothing for that. So `main`'s is strictly broader, and dropping mine loses nothing.

## Two things in the rebase that aren't a straight port

Both are flagged in the body, but they're the parts I'd want a second pair of eyes on:

**`resplit`.** Your test caught this — `defaults to Infinity through breakChangesIntoBatches` went red. A `verifyPieces` re-split walks ops the first pass already reported, so `onOversizedOp` fired twice for one op. Now suppressed on re-split passes. Refusal is untouched: `maxUnsplittableBytes` doesn't move between passes, so emit-vs-refuse is identical either way and only the duplicate report is dropped.

**`splitLargeInsertText`'s seed** now goes through the estimator rather than treating a compressed byte budget as a character count. That seed was several times too short under compression and emitted far more pieces than the budget needed — same revision-count inflation you flagged, just from the other direction. Still only a seed; measure-and-halve and the surrogate-pair handling are untouched.

## `PIECE_COMPRESSION_MARGIN`

Noted in the comment, in the pessimistic direction as you asked — and your framing is exactly why it needed saying, since `splits deterministically` pins the piece *count* but not its magnitude, so over-splitting passes silently. The comment now says to prefer nudging it down rather than up, and names branch-seed revision inflation as what it costs.

## Numbers

I re-measured rather than reusing yours, since my fixture differs — checked `main`'s `changeBatching.ts` into the worktree and ran the new timing test against it, so both sides are the identical 4.2 MB fixture on the same machine:

| | `origin/main` | `c981557` |
| -- | -- | -- |
| one split | **65.6 s** (blew the test's own 60 s ceiling) | **3.9 s** |

Tests: dropped my round-trip and resume-offset duplicates — `changeBatchingApply.spec.ts` covers positioning properly and covers it better. New `changeBatchingEstimator.spec.ts` sticks to the estimator's own contract: per-piece cap, `verifyPieces` re-splitting under a deliberately optimistic calculator, replay, contiguous revs and id preservation, determinism, the uncompressed path staying byte-identical, and the non-quadratic timing above.

Full suite 3392 passed / 4 skipped; `type:check`, `lint`, `format:check` clean. The three `tests/solid/*` suites still fail to load on `@solid-refresh`, identical on clean `main`.

## On scope

Thanks for confirming step 1 landed where it was meant to. Guard rail noted as the one I'll keep pushing for — it survived the rebase untouched because it's dw3-side, and the 95%-of-threshold account is the argument. Happy to open that as its own ticket whenever you want it queued.

jacwright (approved):
Approving. The rebase is exactly right, and the PR is now the part that was genuinely new.

### The rebase did what it needed to

`0` commits behind main, merge base `b59eae9` (0.26.0), and `OTBranchManager.ts` is gone from the diff entirely. `breakTextOp` now carries **main's** `pieceStartPos` / `withStartRetain` / `advanceOf` rather than a second implementation of the same idea — your `resumeAt` / `deltaOpOutputLength` pair is gone, and the estimator is layered on top of `c3c5018` instead of racing it. That's the whole of what I asked for, and threading `opJSONSize({ retain: pieceStartPos })` into the estimate so the prefix is costed is the right way to join them.

722 tests pass across `tests/algorithms/ot/shared/`. The win survived: the multi-megabyte non-quadratic test runs in **6.2s** here, against the **224s** the equivalent takes on `main`.

### One thing to fix: the `verifyPieces` test doesn't test `verifyPieces`

`verifyPieces` is the entire reason an estimate is acceptable rather than a gamble — it's the argument both of us have leaned on. So I checked it the way I'd check any safety gate, by removing it:

```ts
function verifyPieces(...) {
  return pieces; // verification disabled
  …
}
```

**All 7 tests still pass**, including `re-splits a piece the estimate left oversized rather than emitting it`.

The reason is that the hostile calculator is _uniformly_ proportional:

```ts
const optimistic = (data: unknown) => Math.floor(getJSONByteSize(data) * 0.1);
```

`createSizeEstimator` derives `ratio = exactSize / jsonSize` from one real measurement, so against that calculator the ratio is exactly `0.1` — the estimate is perfect everywhere, then `PIECE_COMPRESSION_MARGIN` pads it a further 10%. No piece can land over budget, so the re-split branch is never entered. A ratio-based estimator cannot be fooled by a function that is a constant multiple of JSON size.

What would actually exercise it is a calculator whose cost is **non-uniform** across the split — cheap for the whole change, disproportionately expensive for a piece. That's exactly what real LZ does, which is why `PIECE_COMPRESSION_MARGIN` exists at all. Something like a calculator that returns `jsonSize * 0.1` for inputs above some size and `jsonSize * 0.9` below it would put first-pass boundaries genuinely over budget and force the re-split, and would then fail if `verifyPieces` were ever removed or broken.

Worth doing rather than leaving: the failure this guards against is a change committed above the store's cap, which pup answers with a terminal 413 — your own stated reason for building the gate.

### Small, carried over

`PIECE_COMPRESSION_MARGIN` still has no note on the pessimistic direction. Over-estimating splits a change into more pieces than needed, which is invisible in tests but shows up as revision-count inflation on a branch seed. `splits deterministically` pins the piece count's _stability_, not its _magnitude_.

matalina:
Both notes addressed — pushed `7640dfd`, test-only.

## The `verifyPieces` test now tests `verifyPieces`

You were right about why it couldn't: the ratio calibrates from a real measurement of the very change being split, so a calculator that is any constant multiple of JSON size is estimated perfectly everywhere. I reproduced your probe first — `return pieces;` at the top of `verifyPieces` — and all 7 stayed green, exactly as you found.

The new calculator is the step function you suggested: `0.1×` JSON size at whole-change scale, `0.5×` below half the change's JSON size. Calibration locks onto the cheap tier, first-pass pieces measure ~4.5× over budget, and the re-split has to run. Two properties were deliberate:

- **Convergence.** The expensive tier is uniform *within itself*, so the re-split recalibrates on the oversized piece and lands in one pass — the test doesn't depend on where the 3-pass / 0.8× envelope gives up.
- **It can't go silently vacuous again.** The calculator records any piece-sized measurement that comes back over budget, and the test asserts that list is non-empty. Nothing but `verifyPieces` measures piece-sized payloads, so with the gate removed the test fails on that assert (verified — your probe goes red now), and if a future fixture change ever made the first pass genuinely fit, it fails there too instead of passing while testing nothing.

Re-ran after restoring the gate: 7/7 on the spec, 722 across `tests/algorithms/ot/shared/`, `type:check` / `lint` / `format:check` clean.

## The margin note

One correction on the carried-over item: the pessimistic-direction note did land in `c981557` — it's the comment block on the constant, [changeBatching.ts#L290-L301](https://github.com/dabblewriter/patches/blob/c981557a8c3b6a0b663467cfcd8445857ecd4c9a/src/algorithms/ot/shared/changeBatching.ts#L290-L301), including the revision-count-inflation cost on a branch seed and "prefer nudging this down rather than up". I suspect it carried over from the first review's draft, so nothing further was pushed for it.

jacwright:
Verified. The new test does what the old one couldn't — I disabled `verifyPieces` (`return pieces` at the top) and got exactly one failure, this one. Under the previous uniform calculator all seven passed with the gate removed, so the difference is real and the gate is now genuinely pinned.

The `stepped` calculator is the right model too: cheap at whole-change scale, 5× that rate for piece-sized payloads. That's the direction real LZ actually errs — a large change compresses better than the pieces cut from it — which is the same reason `PIECE_COMPRESSION_MARGIN` exists. And asserting `oversizedMeasurements.length > 0` before checking the pieces is what stops the test quietly degenerating into a tautology if the estimator ever gets good enough to avoid the path.

Nothing further from me on this one.

PR #140, opened 2026-08-06
DAB-930: Merge branches in bounded windows, carrying the OT frame between them

jacwright:
## TL;DR

Merging a review copy back into a manuscript used to be all-or-nothing: a branch that grew too big (one customer's reached 12,000 changes) simply could not merge, and the obvious fix — merging in chunks — would have silently scrambled text whenever the author kept writing while the review copy was open. This change makes big merges work in bounded chunks _and_ keeps every character exactly where its author put it, no matter how the chunks fall. It also makes a refused merge truly leave no trace, matching what the app already tells users.

Supersedes #135 (DAB-930); the window loop, watermark CAS semantics, resume behavior, and a number of tests carry over from @matalina's work there.

## The two defects this solves

**Unbounded merge (the #135 goal).** `mergeBranch` read and committed a branch's whole unmerged history at once — reads, transform, write batch and response all scaling with branch age. pup#226 bounds the read and refuses; the merge itself still couldn't complete.

**The frame bug (found reviewing #135, latent on `main`).** Splitting the commit into windows breaks OT correctness: committed foreign ops are stored in the frame they committed in, and the branch's earlier changes move where those ops land in the branch's frame. The transform walk handles this by advancing each foreign change through the queue's raw ops (`transformIncomingChanges`, patches#66) — but that advance needs the raw ops of _every earlier change in the batch_, which only a full queue supplies. A tail-only window transforms against raw foreign ops and lands its text at shifted offsets — silently, with every then-existing test green. The same shape has been latent in the repeat-merge path on `main` since before #66; windowing just makes it the common case. Repro (now a permanent test): branch inserts a 6-char prefix then types `!` inside a word; source concurrently deletes a word; windowed merge moved the `!` six characters from where it was typed.

## The design: carry the frame

The walk already computes the foreign ops' advanced forms and discarded them. Now it returns them (`transformIncomingChangesWithFrame`), and the merge carries them between windows as the **merge frame** — the source's concurrent changes re-expressed in the branch's frame:

1. **Catch-up** — each window folds source changes committed since the frame's position, in log order: foreign rows append; rows from this branch's own merge batch advance the frame through their _raw_ branch ops (read from the branch log). Crash resume, stale/missing frames, and concurrent merges all reduce to this one path.
2. **Lift + commit** — the window's slice is lifted through the frame and committed at the source tip with `baseRev` = the frame's source position: a fast-forward, so the commit-side read never grows (this also removes the growing `commitChanges` read that was #135's hard blocker). Mid-commit foreign arrivals come back in the commit result and fold into the frame in order.
3. **Persist** — the frame rides in the same CAS write as `lastMergedRev` (`Branch.mergeFrame`), so the pair never diverges; frames over 256KB aren't persisted and the next merge rebuilds via catch-up from the raw logs. A lost CAS drops the in-memory carry and the next window adopts the winner's state.

Consequences:

- **Windows are bounded on both axes** — change count (`maxChangesPerMerge`, default 2,000) and serialized ops bytes (`maxBytesPerMergeWindow`, default 4MB, surfaced to stores as a new `ListChangesOptions.maxBytes` read hint) — so pup#226's "bytes are the right bound" argument survives windowing.
- **The duplication guard is hoisted** ahead of window 1 and streams the branch log in pages: a refused merge leaves zero side effects, which keeps pup's "refusal leaves no side effects" docblock and dw3's "Nothing was changed" strings true. Partial state can only arise from a transient fault, and a retry resumes and completes it (returning the resumed prefix too, so the caller still sees the whole merge).
- **`OTStoreBackend.saveChanges`'s id-uniqueness contract is promoted from _should_ to _must_.** Windowed commits advance `baseRev` past earlier windows, so the read-side dedup cannot cover a re-send; the store's write-time guard (which the interface already called authoritative, and pup already implements via same-batch `_changeIds` markers) is now load-bearing. The in-memory test store enforces it too.
- Repeat merges ("branch stays open, merge again later") are now frame-correct — they were silently wrong on `main` whenever the source had moved with overlapping edits.

## Hardening from adversarial verification

A 17-agent adversarial review (4 attack dimensions, every finding independently verified with runnable repros) confirmed and drove these additional fixes:

- **Frame coverage accounting** (2 critical): the frame's `sourceRev` now advances over every source row the window accounted for — including mid-commit foreign arrivals and a concurrent merge's dedup'd echoes — never just its own committed rows. Without this, an all-noop window (both sides deleting the same scene) or a concurrent-merge race re-folded a foreign change into the frame twice, shifting later windows' ops. Foreign rows in a commit result also now fold through only the _remaining_ as-sent suffix, so rows already expressed post-slice append raw instead of double-advancing.
- **`MergeFrame.programs` persists as a JSON string** — the natural `JSONPatchOp[][]` is an array of arrays, which Firestore refuses as a field value; every windowed merge with concurrent source edits would have failed at the first watermark write in pup.
- **Dense results**: `mergeBranch` returns the full committed span it observed (folded foreign rows included, rev-dense, deduped) — `PatchesSync.applyMergeChanges` routes interior rev gaps to a degraded snapshot path, which windowed results with mid-merge arrivals would otherwise have triggered.
- **`MergePartialProgressError`**: a failure after the first committed window now throws a typed error carrying the committed changes and watermark, so consumers can stop presenting a partially-merged branch as "nothing was changed." (pup mapping + dw3 copy are DAB-930 follow-ups; refusals before window 1 throw unchanged.)
- **`MergeFrameAlignmentError`** (typed, permanent) for a pruned branch log, and reclassification of foreign rows wearing a forged `batchId` — id membership in the branch log is the authoritative own-row classifier, so a client-minted collision folds as foreign instead of wedging the branch forever.
- **Guard TOCTOU** closed (resend corpus re-collected after the head reconstruction) and the legacy no-CAS clamp path re-loads before pinning.
- `listBranches` strips `mergeFrame` — server working state (up to 256KB) must not ride every branch-list sync to every client.

## Review round 1 (@matalina)

- `MergePartialProgressError` and `MergeFrameAlignmentError` now export from the server barrel (`MergeFrame` already reached the root barrel).
- Partial progress is typed on **every** path: a crash between any window's commit and its watermark write — including the first window's — throws `MergePartialProgressError` carrying the crashed window's rows (classification keys on "did a commit land"), and the stalled-cursor and window-cap backstops throw it instead of returning a partial merge as success. Completion is an explicit `drained` signal from the empty-slice read, so a concurrent winner finishing the branch isn't mistaken for a stall.
- A repeat merge where only the source moved returns `[]` again (foreign rows alone are not this branch merging).
- The frame cap and window byte budget measure real UTF-8 bytes; `Branch.mergeFrame` admits `null` as a persisted state.
- New tests: first-window crash, foreign-only no-op repeat (twice), multi-run foreign fold inside one commit result, acks-but-drops watermark store, 600-change window-cap exhaustion.
- Egress audited: patches' RPC surface returns branch records only via `listBranches` (stripped); pup's `_branches` endpoint goes through the manager, and its one direct `store.listBranches` use is server-internal. Clearing the frame when a review copy closes is app-level and goes on the pup follow-up list.

## Scope boundary

This bounds the _branch_ axis. A source that has taken thousands of changes since the branch point (the mirror problem) is unfixed here: the frame is O(source-divergence), exactly as the one-shot commit read was. That case keeps today's clean refusal (pup#226's merge read budget) and its windowing — streaming the source log past the branch queue with the same walk — is a DAB-930 follow-up. In practice the asymmetry is heavily branch-side: DAB-896's branch had 12,000 changes against ~40 concurrent source edits.

## Testing

- The frame repro: windowed and one-shot merges agree on overlapping text edits (failed on the #135 approach; the pre-#66 walk fails it in both modes).
- Repeat merge after the source moved — fails on `main` today.
- Seeded fuzz: 12 branch text edits × 5 interleaved source edits, window sizes {1, 2, 3, 1000} — identical final states across 10 seeds.
- Mid-merge foreign arrival between windows; crash-resume with persisted frame; oversized-frame rebuild path; guard-refusal leaves the source untouched; byte-budgeted window splitting.
- Ported from #135: multi-window merge lands every edit exactly once; windowed ≡ unwindowed; watermark reaches tip; resume after a killed merge.
- Full suite, lint, format, type-check green.

## Shipping

Needs a publish plus a pin bump in pup before production. Order stays: pup#226 (deployed), dw3#1306 (merged — its "Nothing was changed" copy stays _true_ under this design), then this. The partial-merge signal for the transient-crash window (patches error shape → pup `wrapMergeCommit` passthrough → dw3 strings) is tracked on DAB-930 and is no longer a blocking prerequisite, since deterministic refusals now decide before anything commits.

### DIFF
--- a/src/server/OTBranchManager.ts
+++ b/src/server/OTBranchManager.ts
@@ -1,11 +1,20 @@
 import type { Op } from '@dabble/delta';
 import { findLatestMainVersion } from '../algorithms/ot/server/getSnapshotAtRevision.js';
 import { getStateAtRevision } from '../algorithms/ot/server/getStateAtRevision.js';
+import { transformIncomingChangesWithFrame } from '../algorithms/ot/server/transformIncomingChanges.js';
 import { breakChanges } from '../algorithms/ot/shared/changeBatching.js';
 import { createChange } from '../data/change.js';
 import { toOps } from '../json-patch/ops/text.js';
 import { createVersionMetadata } from '../data/version.js';
-import type { Branch, Change, CreateBranchMetadata, EditableBranchMetadata, ListBranchesOptions } from '../types.js';
+import type { JSONPatchOp } from '../json-patch/types.js';
+import type {
+  Branch,
+  Change,
+  CreateBranchMetadata,
+  EditableBranchMetadata,
+  ListBranchesOptions,
+  MergeFrame,
+} from '../types.js';
 import type { BranchManager } from './BranchManager.js';
 import {
   advanceMergeWatermark,
@@ -34,6 +43,28 @@ type OTBranchStore = OTStoreBackend & BranchingStoreBackend;
  * ordinary prose — the default sits well between the two and is configurable per manager.
  */
 const DEFAULT_DUPLICATION_MIN_LENGTH = 64;
+
+/** Default maximum branch changes committed per merge window. */
+export const DEFAULT_MERGE_WINDOW_CHANGES = 2_000;
+
+/** Default maximum serialized ops bytes per merge window — the window's second axis. */
+export const DEFAULT_MERGE_WINDOW_BYTES = 4 * 1024 * 1024;
+
+/**
+ * Backstop against a store that silently drops watermark writes (the loop's cursor check
+ * breaks first in any real scenario). Far above any real branch at the default window size.
+ */
+const MAX_MERGE_WINDOWS = 500;
+
+/**
+ * Frames larger than this are not persisted — branch records live in stores with document
+ * size limits. The merge still completes (the frame is carried in memory); only a resumed or
+ * repeat merge pays the rebuild from the raw logs.
+ */
+const MAX_PERSISTED_FRAME_BYTES = 256 * 1024;
+
+/** Page size for frame catch-up, alignment and guard reads. */
+const FRAME_PAGE_CHANGES = 500;
 
 /** Object-replacement character standing in for embeds — never matches real text. */
 const EMBED_PLACEHOLDER = '￼';
@@ -65,6 +96,19 @@ export interface OTBranchManagerOptions {
    * duplicated text is meaningful, is policy that belongs to the consuming server.
    */
   contentDuplicationGuard?: ContentDuplicationGuardOptions;
+  /**
+   * Maximum branch changes committed per merge window. Windows keep a merge's reads, commit
+   * batches, memory and CPU proportional to the window rather than to how long the branch has
+   * been open. Defaults to {@link DEFAULT_MERGE_WINDOW_CHANGES}.
+   */
+  maxChangesPerMerge?: number;
+  /**
+   * Maximum serialized ops bytes per merge window — the second axis of the window bound
+   * (`maxChangesPerMerge` caps document count; this caps memory for branches carrying large
+   * changes). Passed to the store as a read hint (`ListChangesOptions.maxBytes`) and enforced
+   * in memory after the read. Defaults to {@link DEFAULT_MERGE_WINDOW_BYTES}.
+   */
+  maxBytesPerMergeWindow?: number;
 }
 
 /** Per-merge options for {@link OTBranchManager.mergeBranch}. */
@@ -348,6 +392,56 @@ export class MergeContentDuplicationError extends Error {
 }
 
 /**
+ * Thrown when a windowed merge fails AFTER at least one window committed. The committed
+ * prefix is real and visible on the source; the watermark points at it, so retrying the merge
+ * resumes and completes rather than redoing. Consumers must not present this as "nothing was
+ * changed" — that is only true of refusals raised before the first window (which throw their
+ * original error, e.g. {@link MergeContentDuplicationError}).
+ */
+export class MergePartialProgressError extends Error {
+  readonly code = 'MERGE_PARTIAL_PROGRESS';
+  constructor(
+    readonly branchId: string,
+    /** Every change committed (and observed) before the failure, in rev order. */
+    readonly committedChanges: Change[],
+    /** The branch revision the merge is durably merged through. */
+    readonly mergedThroughRev: number | undefined,
+    override readonly cause: unknown
+  ) {
+    super(
+      `Merge of branch ${branchId} failed after committing ${committedChanges.length} changes ` +
+        `(merged through branch rev ${mergedThroughRev ?? 'unknown'}). The committed prefix is ` +
+        `permanent; retrying the merge resumes and completes it. Cause: ` +
+        `${cause instanceof Error ? cause.message : String(cause)}`,
+      { cause }
+    );
+    this.name = 'MergePartialProgressError';
+  }
+}
+
+/**
+ * Thrown when the merge frame cannot be aligned with the branch's change log: the source
+ * carries committed merge rows whose branch changes no longer exist below the log's oldest
+ * retained revision (a pruned/compacted branch log). This is permanent for the branch — the
+ * raw ops the frame needs are gone — and needs operator attention rather than a retry.
+ */
+export class MergeFrameAlignmentError extends Error {
+  readonly code = 'MERGE_FRAME_ALIGNMENT';
+  constructor(
+    readonly branchId: string,
+    readonly missingIds: string[],
+    detail: string
+  ) {
+    super(
+      `Merge frame alignment failed for branch ${branchId}: ${detail} Retrying cannot succeed; ` +
+        `the branch's raw change log no longer contains changes its committed merge rows reference ` +
+        `(ids: ${missingIds.slice(0, 5).join(', ')}${missingIds.length > 5 ? ', …' : ''}).`
+    );
+    this.name = 'MergeFrameAlignmentError';
+  }
+}
+
+/**
  * OT-specific branch manager implementation.
  *
  * Manages branches for documents using Operational Transformation semantics:
@@ -380,7 +474,15 @@ export class OTBranchManager implements BranchManager {
    * @returns The branches.
    */
   async listBranches(docId: string, options?: ListBranchesOptions): Promise<Branch[]> {
-    return await this.store.listBranches(docId, options);
+    const branches = await this.store.listBranches(docId, options);
+    // The merge frame is server-internal working state and can run to 256KB; it must not ride
+    // every branch-list sync to every client.
+    return branches.map(branch => {
+      if (!('mergeFrame' in branch)) return branch;
+      const rest = { ...branch };
+      delete rest.mergeFrame;
+      return rest;
+    });
   }
 
   /**
@@ -472,7 +574,7 @@ export class OTBranchManager implements BranchManager {
   }
 
   /**
-   * Merges changes from a branch back into its source document.
+   * Merges changes from a branch back into its source document, in bounded windows.
    *
    * Supports multiple merges — the branch stays open and `lastMergedRev` tracks
    * which branch revision was last merged. Subsequent merges only pick up new changes.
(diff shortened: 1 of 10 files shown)


src/server/OTBranchManager.ts
```diff
@@ -472,97 +574,426 @@ export class OTBranchManager implements BranchManager {
...
+        // a count heuristic, which a byte-trimmed or store-truncated read would fool.
+        if (previousBranchRev !== undefined && result.branchRev <= previousBranchRev) break;
+        previousBranchRev = result.branchRev;
+      }
+    } catch (error) {
+      // A failure after the first window committed is NOT "nothing happened": the prefix is
+      // permanent, the watermark points at it, and a retry resumes. Surface that distinctly
+      // so consumers stop telling users the merge left no trace. A failure before anything
+      // committed (including guard refusals, which run before the loop) throws unchanged.
+      if (committed.length > 0) {
```
matalina, line 722: If the **first** window's commit lands and its watermark write throws, `mergeBranchWindow` throws before returning, `committed` is still empty here, and the raw error propagates — so the very case this error exists for (commit landed, consumer told "no trace") presents as "nothing was changed" whenever the branch fits in one window, which is the common case. The `resumes a crashed windowed merge` test pins this behavior (`rejects.toThrow('simulated crash')`).

Suggestion: catch failures that occur *after* a successful `saveChanges` inside `mergeBranchWindow` and rethrow carrying that window's `windowCommitted`, so classification keys on "did a commit land" rather than "did a window return its result".

Related: `error.committedChanges` omits rows committed in the window that crashed — `failWatermarkOnce(store, 2)` reports `['b1']` while `b2` is also on the source. The retry heals it, but the field's docblock ("Every change committed (and observed) before the failure") overpromises; either the fix above covers both, or soften the docblock.

src/server/OTBranchManager.ts
```diff
@@ -472,97 +574,426 @@ export class OTBranchManager implements BranchManager {
...
+    slice = trimToByteBudget(slice, windowBytes);
+
+    if (slice.length === 0) {
+      // Nothing new to merge. Persist any progress the catch-up made (a pure resume), then
+      // surface everything the catch-up observed so a retried merge still returns the full,
+      // rev-dense result.
+      const nextCarry = await this.persistMergeProgress(branch, frame);
+      const resumedRows = byRev([...resumed.own, ...resumed.foreign]);
+      if (resumedRows.length === 0) return undefined;
+      return { changes: resumedRows, branchRev: frame.branchRev, carry: nextCarry };
```
matalina: Contract change worth a deliberate decision: a repeat merge with nothing new on the branch, but foreign source edits above the persisted frame's `sourceRev`, returns those foreign rows as the merge result — the old contract was `[]`. Rev-density only needs foreign rows that sit *below a resumed own row*; with no own rows there is nothing to keep dense. A consumer reporting "merged N changes" will miscount, and because the early-return in `persistMergeProgress` skips the write in this path, every repeat call returns the same rows again.

Consider returning `undefined` here when `resumed.own` is empty, so `mergeBranch` still answers `[]` for a true no-op. Note the `advances the watermark... follow-up merge is a no-op` test only passes because no foreign edit lands after the merge — a variant with one would pin whichever behavior you intend.

src/server/OTBranchManager.ts
```diff
@@ -472,97 +574,426 @@ export class OTBranchManager implements BranchManager {
...
+          committedIds.add(change.id);
+          committed.push(change);
+        }
+        carry = result.carry;
+        // The cursor must move every window, or the next read returns the same changes and the
+        // loop spins re-committing them. A store that silently drops the watermark write, or a
+        // concurrent merge that regressed it, stops here with the work done so far. Completion
+        // is detected by a window reading an empty slice (returning undefined above) — never by
+        // a count heuristic, which a byte-trimmed or store-truncated read would fool.
+        if (previousBranchRev !== undefined && result.branchRev <= previousBranchRev) break;
```
matalina: Both this stall break and `MAX_MERGE_WINDOWS` exhaustion return a partial merge as silent success — a store that drops watermark writes yields "merged" with the tail missing, which cuts against the care taken everywhere else in this PR (typed partial-progress error, refusing count heuristics). A `console.warn` naming the branch — or throwing `MergePartialProgressError`, since the committed prefix is in hand — would keep the "never silently partial" story consistent on the backstop paths too.

src/server/OTBranchManager.ts
```diff
@@ -472,97 +574,426 @@ export class OTBranchManager implements BranchManager {
...
+   */
+  private async persistMergeProgress(branch: Branch, frame: WorkingFrame): Promise<MergeCarry | undefined> {
+    if (frame.branchRev <= (branch.lastMergedRev ?? 0)) {
+      // Nothing newly covered — don't touch the record (or regress a concurrent advance).
+      return { frame, expectedWatermark: branch.lastMergedRev };
+    }
+    // Persisted programs are a JSON string: the natural array-of-arrays shape is unwritable
+    // as a field value in document stores like Firestore.
+    const programs = JSON.stringify(frame.programs);
+    const fits = programs.length <= MAX_PERSISTED_FRAME_BYTES;
```
matalina: `programs.length` counts UTF-16 code units, not bytes — CJK/emoji-heavy content undercounts UTF-8 by up to 3×, so a "256KB" frame can weigh ~768KB inside a Firestore doc capped at 1MB, alongside the rest of the branch record. The margin holds today, but either measure real bytes (`new TextEncoder().encode(programs).length`) or document the unit honestly. Same unit fib in `trimToByteBudget` and the `maxBytes` hint, where the stakes are lower.

src/server/OTBranchManager.ts
```diff
@@ -472,97 +574,426 @@ export class OTBranchManager implements BranchManager {
...
+          while (
+            runEnd + 1 < result.length &&
+            !sliceIds.has(result[runEnd + 1].id) &&
+            result[runEnd + 1].batchId !== branchId
+          ) {
+            runEnd++;
+          }
+          const run = result.slice(i, runEnd + 1);
+          advancedForeign.push(
+            ...advanceProgramsThroughQueue(
```
matalina: Test-coverage suggestion, not a bug I can demonstrate: the sequential-suffix fold across **multiple** foreign runs in one commit result (foreign row / own row / foreign row within a single window) is the subtlest code in the walk, and the existing tests only exercise single-run interleavings per window (`converges when a foreign change lands mid-merge` lands one row per window). One test injecting two foreign commits at different points inside a single window's commit — e.g. via a `saveChanges` hook on the source doc — would pin the suffix consumption across runs.

src/server/OTBranchManager.ts
```diff
@@ -380,7 +474,15 @@ export class OTBranchManager implements BranchManager {
...
    */
   async listBranches(docId: string, options?: ListBranchesOptions): Promise<Branch[]> {
-    return await this.store.listBranches(docId, options);
+    const branches = await this.store.listBranches(docId, options);
+    // The merge frame is server-internal working state and can run to 256KB; it must not ride
+    // every branch-list sync to every client.
+    return branches.map(branch => {
+      if (!('mergeFrame' in branch)) return branch;
+      const rest = { ...branch };
+      delete rest.mergeFrame;
```
matalina, line 504: This covers the list path — worth confirming before the pup pin bump that no *other* egress ships whole branch records to clients (branch subscription pushes, any get-branch RPC), or a 256KB frame rides out to every client anyway.

src/server/OTBranchManager.ts
```diff
@@ -472,97 +574,426 @@ export class OTBranchManager implements BranchManager {
...
+   * — and their raws still advance the frame: the branch's program included them.
+   *
+   * `batchId` is client-mintable, so id membership in the branch log is the authoritative
+   * classifier: a run row whose id is nowhere in the log is an ordinary foreign change
+   * wearing our batch id — it folds into the frame as a program instead of wedging the merge.
+   * Returns the reclassified-foreign ids. A genuine gap in the branch log (a pruned log whose
+   * oldest retained rev is past the frame's position) throws {@link MergeFrameAlignmentError}:
+   * the raw ops the frame needs are permanently gone.
+   */
+  private async advanceFrameThroughOwnRun(branchId: string, frame: WorkingFrame, run: Change[]): Promise<Set<string>> {
```
matalina, line 975: Note on cost: a foreign row wearing this batchId whose id is nowhere in the branch log makes this scan page through the entire remaining log before classifying it as an impostor — K forged rows cost K × O(branch) paged reads on the next merge. Collaborator-only exposure, so fine as-is, but worth a comment (or a cheap scan cap) if it ever shows up in a profile.

src/types.ts
```diff
@@ -145,6 +145,17 @@ export interface Branch {
...
+  /**
+   * The merge frame persisted by windowed merges: the source's concurrent (foreign) changes
+   * re-expressed in the branch's frame after every branch change at or below `lastMergedRev`.
+   * `sourceRev` is the source revision the frame covers — foreign changes at or below it are
+   * folded in; the merge folds newer ones on demand. Written atomically alongside
+   * `lastMergedRev` so the two never diverge; absent (legacy branches, or a frame too large
+   * to persist) the merge rebuilds it from the branch's raw change log. Server-managed like
+   * `lastMergedRev`; client-supplied values are stripped.
+   */
+  mergeFrame?: MergeFrame;
```
matalina: The clear path writes `mergeFrame: null` (and the tests need `as any` / `toBeNull()` to work with it), but the declared type doesn't admit null. Make it `mergeFrame?: MergeFrame | null` so the casts drop and readers know null is a real persisted state, distinct from absent.

matalina (changes requested):
The core design is sound and unusually well-tested — the frame/lift diamond, the carry validation, CAS-loss handling, and the guard's TOCTOU re-collection all trace correctly, and the fuzz + race coverage (both CAS and legacy stores, noop windows in both positions, forged batchId, pruned logs) is excellent. Requesting changes for one concrete blocker, plus two behavioral edges flagged inline that deserve a deliberate decision even if the answer is "as designed".

**Blocker — the new error classes aren't exported from the package.** `MergePartialProgressError` and `MergeFrameAlignmentError` are new public API — pup's `wrapMergeCommit` mapping is a stated DAB-930 follow-up — but `src/server/index.ts` isn't touched by this PR and the package `exports` map has no deep-import path, so consumers literally cannot import them to `instanceof`-match (the `code` string fields would allow duck-typing, but that shouldn't be the only option). Add both to the barrel next to `MergeContentDuplicationError`, and confirm the `MergeFrame` type reaches the root barrel.

Flagged inline:
- A window-1 commit→watermark crash still presents as "nothing was changed" — the exact consumer inference `MergePartialProgressError` exists to prevent, in the single-window (common) case.
- A no-op repeat merge can return pure foreign rows where the old contract returned `[]`.
- The cursor-stall break and `MAX_MERGE_WINDOWS` exhaustion return partial merges as silent success.
- Nits: UTF-16 code units vs bytes; `mergeFrame` null typing; a test gap on multi-run folds; frame egress beyond `listBranches`; impostor scan cost.

Two non-blocking notes with no inline anchor:
- The frame stays on closed branches forever (up to 256KB of dead weight per record) — consider clearing it in `closeBranch`.
- Performance shape for the record: the guard, when configured, is now ~4 paged passes over branch + source-since-frame per merge (memory-bounded, O(branch) reads), and a no-op merge of a legacy frameless branch pays a full source catch-up walk before discovering there's nothing to merge (previously one empty read). Both acceptable; just stating them so they're a known cost.

Verified against `main` while reviewing: `commitChanges`' `DuplicateChangeIdsError` resend loop behaves as the idempotency story assumes, `ListVersionsOptions.endBefore` exists, and the transform-walk semantics the frame extension leans on match `transformIncomingChanges`' existing diamond.

src/server/OTBranchManager.ts
jacwright: Fixed as suggested: failures after a successful commit rethrow through an internal fault carrying everything the window observed (its committed rows, the resumed prefix, folded foreign rows), so classification keys on whether a commit landed, not whether the window returned. The window-1 case now throws `MergePartialProgressError` — the updated `resumes a crashed windowed merge` test asserts `committedChanges` `['s2','b1','b2']` and `mergedThroughRev` 3 on a first-window crash. That also covers the docblock overpromise: the crashed window's rows ride in the error, and `mergedThroughRev` now documents what it actually means — content durably on the source, with the watermark healing on retry.

src/server/OTBranchManager.ts
jacwright: Agreed, and fixed with your suggestion: the empty-slice path returns `undefined` when nothing own was resumed, so a repeat merge answers `[]` again — rev-density only needs foreign rows beneath a resumed own row, exactly as you framed it. New test `answers [] for a repeat merge when only the source has moved` commits a foreign edit after the merge and asserts `[]` twice, pinning the repeat-reporting variant too.

src/server/OTBranchManager.ts
jacwright: Both backstops now throw, wrapped into `MergePartialProgressError` whenever anything committed: the cursor-stall error names the rev and blames the watermark write, the cap error names `MAX_MERGE_WINDOWS`. Completion is now signaled explicitly by the empty-slice read (a `drained` flag on the window result), so a merge that a concurrent winner finished doesn't false-positive as a stall. Unit tests cover a store that acks-but-drops the watermark write and a 600-change branch at window size 1 exhausting the cap.

src/server/OTBranchManager.ts
jacwright: Both now measure real UTF-8 bytes via `TextEncoder` — the persisted-frame cap and `trimToByteBudget`. The `ListChangesOptions.maxBytes` hint stays advisory; its docblock already tells callers to tolerate larger results, which covers stores approximating with cheaper units.

src/server/OTBranchManager.ts
jacwright: Added `folds multiple foreign runs from one window commit through the right suffixes`: the loser of a merge race parks on its slice read; f0 lands, the winner merges the slice, f1 lands; the loser's resend answers `[f0, b1, b2, f1]` in one result — the first run folds through the whole as-sent queue, the last through nothing. Verified by equivalence against the same history run sequentially, plus a follow-up merge through the surviving frame.

src/server/OTBranchManager.ts
jacwright: Confirmed on both sides. In patches, the RPC surface is exactly the five `branchManagerApi` methods and only `listBranches` returns records (stripped); nothing in `src/net` pushes branch records. In pup, `GET /docs/:docId/_branches` goes through the manager, so the strip covers it after the pin bump; the only direct `store.listBranches` call is presence scope computation, which reads `branch.id` server-side and returns nothing to clients. On your unanchored note about closed branches: close is app-level metadata in patches (there's no closeBranch), so clearing the frame when Dabble archives a review copy belongs in pup's close path — added to the DAB-930 follow-up list next to the `MergePartialProgressError` mapping.

src/server/OTBranchManager.ts
jacwright: Added the cost note to the docblock, including why there's no safe early cutoff (a genuine run's first match can sit arbitrarily deep past lift-noops committed under a different manager's window size) — so a future profiler finds the analysis next to the loop.

src/types.ts
jacwright: Done — `mergeFrame?: MergeFrame | null`, with the docblock stating null is a real persisted state (explicitly cleared, e.g. oversized) distinct from absent. The test casts are gone.

jacwright:
Thanks Alicia — every point landed, this is a sharper PR for it. All addressed in ea77791 (inline replies on each thread):

- **Blocker**: both error classes export from `@dabble/patches/server`; `MergeFrame` already reached the root barrel via `export type *`.
- **Window-1 crash**: partial progress now keys on "did a commit land" — an internal fault carries the crashed window's observed rows out to `MergePartialProgressError`, first window included, and `committedChanges` includes the crashed window's rows (your b2 case).
- **No-op repeat**: returns `[]` when only the source moved, per your suggested `undefined`-when-no-own-rows shape.
- **Backstops**: stall and window-cap throw partial progress instead of silent success; completion is an explicit drained signal so a concurrent winner isn't a false stall.
- Bytes are real UTF-8 now, `mergeFrame` admits `null`, the multi-run fold has its test, the impostor scan cost is documented at the loop, and the egress audit (patches + pup) is in the thread — with close-path frame clearing noted for pup.

Full suite/lint/format/type-check green. Re-requesting your review.

matalina (approved):
Approving. Every round-1 item is genuinely fixed rather than papered over, and two of the fixes are better than what I asked for: completion is now an explicit `drained` signal instead of a count heuristic (so a concurrent winner finishing the branch can't read as a stall), and partial-progress classification keys on "did a commit land" rather than "did the window return", which is what makes the first-window case fall out naturally instead of needing its own branch.

**Round-1 items verified**

- **Blocker:** both error classes export from `src/server/index.ts`; `MergeFrame` reaches the root barrel via `export type * from './types.js'`.
- Window-1 commit→watermark crash now throws `MergePartialProgressError` carrying the crashed window's rows (`['s2','b1','b2']`, `mergedThroughRev: 3`).
- No-op repeat answers `[]`, twice in a row — and correctly, since nothing persists on that path.
- Stalled cursor and `MAX_MERGE_WINDOWS` exhaustion both throw instead of returning a partial merge as success.
- UTF-8 bytes, `mergeFrame: MergeFrame | null`, the multi-run fold test, and the impostor scan cost note are all in.

**What I re-verified**

Full suite 3415 passed / 4 skipped. The three failing `tests/solid/*` suites fail identically on `main` (`file:///@solid-refresh` vite resolution) — pre-existing, not this PR. `tsc --noEmit` and eslint clean.

Downstream, since this PR makes the store's write-time guard load-bearing rather than a backstop:

- pup satisfies the promoted `saveChanges` contract — `FirestoreOTStore` writes `_changeIds/{id}` markers in the same batch as the changes and throws `DuplicateChangeIdsError`. `updateBranchIf` is implemented and forwarded on the composite store, so production gets the CAS watermark path, not legacy max-wins.
- pup applies its unbounded-read 413 only when `limit === undefined`. Every read this PR adds passes `limit`, so windowed merges won't trip pup#226's budget — and the guard's old unbounded `listChanges(startAfter: baseRev)` becoming paged is a straight improvement.
- pup's `listVersions` handles `endBefore`, `groupId`, `reverse` and `limit` for both shapes `copyBranchVersions` and `latestBranchVersionCopyId` use.

**Non-blocking, for a follow-up or a docblock line**

1. **`maxBytes` is a no-op against the only production backend.** pup's `listChanges` ignores `ListChangesOptions.maxBytes` (its `budget.maxBytes` is the separate `UnboundedReadBudget`), so a window still reads and decompresses up to `limit` (2,000) changes before `trimToByteBudget` discards the tail. `OTBranchManagerOptions.maxBytesPerMergeWindow`'s "caps memory for branches carrying large changes" is true of the transform/commit side but not the read side. Worth either softening that line or tracking the pup-side plumbing.

2. **`slice.length === 0` is read as drained unconditionally.** `ListChangesOptions.maxBytes` documents "always returning at least one change", but if a backend ever violates it the merge reports completion with the watermark unmoved — the branch looks merged and isn't. `trimToByteBudget` keeps at least one change for exactly this reason; a symmetric assert on the read side (empty slice while the branch tip is above `frame.branchRev`) would turn a store bug into a loud one instead of a silent under-merge. Cheap insurance given `saveChanges` was just promoted to MUST on the same reasoning.

3. **`mergeBranch` returns `committed` sorted by construction, not by contract.** It's built by appending each window's `byRev`-sorted result, which stays ascending only while every window's catch-up starts at or above the previous window's max observed rev. I traced the carry, `carryInvalid` and lost-CAS paths and could not construct a violation — so this is robustness, not a defect. But the consumer, `PatchesSync.applyMergeChanges`, takes `mergeChanges[0].rev` / `[length-1].rev` as the min/max and then applies the array in order, so a violation would mis-apply rather than route to the degraded snapshot path it was hardened for. A single `byRev(committed)` on the return makes the guarantee unconditional rather than emergent.

4. **The guard can now pin `mergeBaseRev` before it refuses.** `runContentDuplicationGuard` calls `resolveMergeBase`, which writes the pin on a clamped branch before the refusal throws. "A refused merge leaves no side effects" stays true of the source document, but is no longer true of the branch record — and the guard only needs a read floor for `sinceRev`, not a pin. Either read `branch.mergeBaseRev ?? branchedAtRev` there, or note the exception in the docblock, since the PR leans on that promise for pup's and dw3's copy.

5. **Minor:** `latestBranchVersionCopyId` adds a per-window `groupId` query that pup serves by fetching every branch-origin version for that branch on the source and sorting in memory. Bounded by one branch's version count, so fine — just flagging it as new per-window work.

6. **Confirmed from round 1:** an oversized/cleared frame drops the guard's `sinceRev` back to `mergeBase`, so a guard-enabled merge on such a branch pages the whole source log since the base on every attempt. The guard is env-gated in pup (`mergeDuplicationGuardFromEnv`), so this is live-configurable rather than hypothetical.

The frame/lift diamond, catch-up-as-one-path, and the `accountedRev` coverage accounting all still trace correctly on re-read, and the round-2 delta didn't disturb them. Ship it after the publish + pup pin bump.

PR #141, opened 2026-08-07
feat(client): surface pending changes dropped at hydration (DAB-854)

matalina:
## Problem

OTDoc's constructor recovery drops pending changes that fail strict apply against the hydrated committed state. The drop itself is required for liveness — a change the local state rejects would also be rejected server-side at flush, and a poison change at the head of the queue wedges every commit behind it. But the drop was **fully silent**, and the next pending persist makes the truncation permanent: user work destroyed with zero signal. This is one of the producing paths behind DAB-854 / DAB-947-class content loss.

A related blind spot: `hasPending` does not cover the optimistic tier (`_optimisticOps`), so between a keystroke's `change()` and its mint confirmation, `doc.state` differs from the committed frame while `hasPending` reads false. Consumers comparing `doc.state` against a committed store snapshot in that window see phantom divergences (the dabble-writer-3.0 equal-rev heal was doing exactly this).

## Changes

- **`BaseDoc.droppedPendingChanges`** — hydration drops are captured on the doc instead of vanishing; the constructor logs ids/revs (never content).
- **`Patches.onPendingDropped`** — emitted from `openDoc` when a hydration reported drops, modeled on `onChangeQuarantined`; this is the only moment the payload is still in hand, so consumers can shelve it. Structurally probed so custom doc classes without the field are safe; cleared in `close()`.
- **`BaseDoc.optimisticOpsCount`** — exposes the optimistic tier so audits can gate on it or compare committed frames instead.

All additive API; no behavior change to existing paths.

## Notes for review

While testing this I confirmed strict apply does **not** throw for plain `replace`/`remove` on missing paths — it throws for `@txt` ops on a non-Delta target, failed `test`, and `copy`/`move` with a missing source. So hydration drops predominantly eat **text ops whose target shape changed** — manuscript bodies. The new test uses that real shape.

## Testing

- New: drop capture + clean-hydration cases (`OTDoc.spec`), `optimisticOpsCount` window test, `onPendingDropped` emission ×2 (`Patches.spec`)
- Full suite: 3,390 passing (the 3 `tests/solid/*` suite failures are a pre-existing Windows env issue, identical on untouched main); type/lint/format clean

## Consumer

dabblewriter/dabble-writer-3.0#dab-854-equal-rev-audit subscribes to `onPendingDropped` (feature-detected) and shelves the payload via its unsaved-changes shelf. The loop closes when this ships in a release and dw3 bumps its pin.

### DIFF
--- a/src/client/BaseDoc.ts
+++ b/src/client/BaseDoc.ts
@@ -40,6 +40,15 @@ export abstract class BaseDoc<T extends object = object> extends ReadonlyStoreCl
    */
   protected _optimisticOps: JSONPatchOp[][] = [];
 
+  /**
+   * Pending changes discarded during hydration because they failed strict apply against
+   * the snapshot's committed state (see OTDoc's constructor recovery). Empty for a clean
+   * hydration. Populated by subclasses BEFORE the constructor returns, so `Patches.openDoc`
+   * can surface the loss (`onPendingDropped`) instead of it vanishing silently — dropped
+   * changes are permanent once the next pending write persists the truncated queue.
+   */
+  readonly droppedPendingChanges: Change[] = [];
+
   /** Current sync status of this document. */
   readonly syncStatus = store<DocSyncStatus>('unsynced');
 
@@ -76,6 +85,18 @@ export abstract class BaseDoc<T extends object = object> extends ReadonlyStoreCl
 
   /** Are there local changes that haven't been committed yet? */
   abstract get hasPending(): boolean;
+
+  /**
+   * Count of op batches applied optimistically to `state` but not yet confirmed into the
+   * pending tier. NOT covered by `hasPending`: between change() and its applyChanges()
+   * confirmation this is non-zero while `hasPending` can read false — so `state` differs
+   * from the committed frame in a window `hasPending` cannot see. Consumers comparing
+   * state against a committed snapshot (audits, divergence heals) must gate on this too,
+   * or compare committed frames instead.
+   */
+  get optimisticOpsCount(): number {
+    return this._optimisticOps.length;
+  }
 
   /**
    * Captures an update to the document, applies it optimistically to state,
--- a/src/client/BaseDoc.ts
+++ b/src/client/BaseDoc.ts
@@ -87,14 +87,15 @@ export abstract class BaseDoc<T extends object = object> extends ReadonlyStoreCl
   abstract get hasPending(): boolean;
 
   /**
-   * Count of op batches applied optimistically to `state` but not yet confirmed into the
-   * pending tier. NOT covered by `hasPending`: between change() and its applyChanges()
-   * confirmation this is non-zero while `hasPending` can read false — so `state` differs
-   * from the committed frame in a window `hasPending` cannot see. Consumers comparing
-   * state against a committed snapshot (audits, divergence heals) must gate on this too,
-   * or compare committed frames instead.
+   * Count of op BATCHES applied optimistically to `state` but not yet confirmed into the
+   * pending tier — one per un-confirmed change() call, NOT a count of individual ops
+   * (each batch may carry many). NOT covered by `hasPending`: between change() and its
+   * applyChanges() confirmation this is non-zero while `hasPending` can read false — so
+   * `state` differs from the committed frame in a window `hasPending` cannot see.
+   * Consumers comparing state against a committed snapshot (audits, divergence heals)
+   * must gate on this too, or compare committed frames instead.
    */
-  get optimisticOpsCount(): number {
+  get optimisticBatchCount(): number {
     return this._optimisticOps.length;
   }
 
--- a/src/client/OTDoc.ts
+++ b/src/client/OTDoc.ts
@@ -73,6 +73,14 @@ export class OTDoc<T extends object = object> extends BaseDoc<T> {
         }
         this._pendingChanges = valid;
         this.state = state;
+        // Hardcoded console.error rather than an onSkippedChange-style hook (the
+        // convention applyChangesForReconstruction uses): the constructor is invoked
+        // through ClientAlgorithm.createDoc(docId, snapshot), which has no options
+        // plumb-through, and no consumer can have subscribed to anything yet. The
+        // structured channel is Patches.onPendingDropped, emitted from openDoc right
+        // after construction — this log is the fallback signal for non-Patches hosts
+        // and for drops that occur before any subscriber exists. Ids and revs only,
+        // never content.
         console.error(
           `OTDoc(${id}): dropped ${this.droppedPendingChanges.length} of ${
             this.droppedPendingChanges.length + valid.length
--- a/src/client/OTDoc.ts
+++ b/src/client/OTDoc.ts
@@ -51,8 +51,16 @@ export class OTDoc<T extends object = object> extends BaseDoc<T> {
         this.state = applyChangesToState(this._committedState, this._pendingChanges);
       } catch {
         // Pending changes are corrupt (conflicting ops from accumulated sessions).
-        // Apply one-by-one, skipping changes that fail. Later changes created on
+        // Apply one-by-one, dropping changes that fail. Later changes created on
         // committed state may still apply even when earlier ones conflict.
+        //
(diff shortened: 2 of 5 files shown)


src/client/OTDoc.ts
```diff
@@ -51,20 +51,34 @@ export class OTDoc<T extends object = object> extends BaseDoc<T> {
...
             state = applyPatch(state, c.ops, { strict: true });
             valid.push(c);
           } catch {
-            // Skip this corrupt change
+            this.droppedPendingChanges.push(c);
           }
         }
         this._pendingChanges = valid;
         this.state = state;
+        console.error(
```
jacwright, line 84: A hardcoded `console.error` in a library, when the file next door already solved this exact problem with a hook.

`applyChangesForReconstruction` in `algorithms/ot/shared/applyChanges.ts` takes an `onSkippedChange` callback and *defaults* to `console.error` — and its docblock gives the same rationale you're giving here, that affected tuples need enumerating for a data-repair sweep. Same situation, same library, one convention already chosen.

It matters more here because `onPendingDropped` **is** the structured channel. A consumer that subscribes gets the payload and handles it, and then still gets an unconditional `console.error` it can't turn off. In dw3 that's every affected doc open, and if `CaptureConsole` is on it becomes a Sentry event per doc on top of the `pending-dropped-at-hydration` telemetry the consumer already emits.

Not a blocker — the constructor has no consumer handle to route through, which is presumably why it went this way. But it's worth either matching the `onSkippedChange` shape or saying in the comment why the constructor can't.

src/client/Patches.ts
```diff
@@ -424,6 +433,16 @@ export class Patches {
...
       baseDoc._setFlushAwaiter(() => this._changeQueues.get(docId));
 
+      // Hydration dropped pending changes (strict-apply failures — see the constructor
+      // recovery in OTDoc). Surface them BEFORE any consumer can persist the truncated
+      // queue: this signal is the only moment the dropped payload is still in hand.
+      // Probed structurally — createDoc's contract returns a PatchesDoc, which need not
+      // extend BaseDoc (custom doc classes, test doubles).
+      const dropped = (doc as { droppedPendingChanges?: Change[] }).droppedPendingChanges;
+      if (dropped && dropped.length > 0) {
+        this.onPendingDropped.emit(docId, [...dropped]);
```
jacwright, line 449: The emit lands before `this.docs.set(docId, …)`, so a handler runs while the doc is not yet registered and `_openingDocs` still holds the id.

That's the right position for the stated goal — nothing can persist the truncated queue before it, since `doc.onChange` is wired on the next line — so I'm not asking you to move it. But it's a constraint on handlers that isn't written down: a handler calling `getOpenDoc(docId)` gets `undefined`, and one that `await`s `openDoc(docId)` re-enters a `_openDocOnce` that hasn't resolved.

The dw3 consumer is fine (`_shelvePendingChanges` only writes IndexedDB and emits telemetry), so this is documentation rather than a defect. Worth a line on the signal's docblock saying the doc isn't registered yet, next to the "only moment the payload is in hand" note that explains why.

src/client/BaseDoc.ts
```diff
@@ -77,6 +86,18 @@ export abstract class BaseDoc<T extends object = object> extends ReadonlyStoreCl
...
 
+  /**
+   * Count of op batches applied optimistically to `state` but not yet confirmed into the
+   * pending tier. NOT covered by `hasPending`: between change() and its applyChanges()
+   * confirmation this is non-zero while `hasPending` can read false — so `state` differs
+   * from the committed frame in a window `hasPending` cannot see. Consumers comparing
+   * state against a committed snapshot (audits, divergence heals) must gate on this too,
+   * or compare committed frames instead.
+   */
+  get optimisticOpsCount(): number {
```
jacwright: Name says ops, value is batches — `_optimisticOps` is `JSONPatchOp[][]`, so this is the number of un-confirmed `change()` calls, not the number of ops in them.

The docblock says "op batches" so it's not undocumented, and for the gating use it's exactly right (`> 0` is the question anyone asks). The risk is only someone summing it or reporting it as an op count later — dw3 already stores it as `optimistic_ops` in its evidence payload, where the name will outlive the docblock. `optimisticBatchCount` would close it, or a one-word note that the number is batches.

jacwright (approved):
Approving. Turning a silent drop into a signal is the right shape, and the reasoning for keeping the drop is the part that makes it reviewable — you argue for liveness explicitly rather than leaving "why not just keep the change" for someone to rediscover.

Verified the two claims everything rests on. `hasPending` really is only `_pendingChanges.length > 0` on OTDoc, so the optimistic tier genuinely is invisible to it. And the recovery loop cannot log a vacuous zero: `applyChanges` uses the identical `applyPatch(state, ops, { strict: true })` the loop does, so a bulk throw guarantees at least one individual failure.

**3432 tests pass across 128 files here, zero failures** — the three `tests/solid/*` failures you saw don't reproduce on macOS, which supports your read that they're environment-specific rather than something reviewers need to discount.

Three notes inline, none blocking.

src/client/OTDoc.ts
matalina: Documented in c6ef175: the constructor can't take an onSkippedChange-style hook because createDoc(docId, snapshot) has no options plumb-through and no subscriber can exist at construction — onPendingDropped is the structured channel, and the log is the fallback for non-Patches hosts. If the double-signal (shelf telemetry + console noise under CaptureConsole) proves annoying in practice, gating the log on signal subscribers would be the follow-up.

src/client/Patches.ts
matalina: Added to the docblock in c6ef175: handlers run before docs.set, getOpenDoc(docId) returns undefined, awaiting openDoc(docId) re-enters an unresolved open — persist elsewhere and return.

src/client/BaseDoc.ts
matalina: Renamed to optimisticBatchCount in c6ef175, and the dw3 consumer's telemetry field is now optimistic_batches to match — done now while nothing published depends on either name.

jacwright (commented):
Follow-up to my approval, so whichever of the pair merges first catches it: **dw3#1339 now probes `optimisticBatchCount`, this branch still exposes `optimisticOpsCount`.**

She renamed the consumer side at `a165d19b` after my note that the accessor counts batches rather than ops. If this publishes under the current name, dw3's `optimistic_batches` records `-1` permanently — and `-1` is documented there as "the doc predates the accessor", so it looks like a legitimate legacy value instead of a broken probe. dw3's test can't catch it either; `pupSync.test.ts:529` sets the property on its own stub, so it passes under any name.

`optimisticBatchCount` is the better name and this is the cheaper side to change, since the accessor is new in this PR and isn't public API until you publish. Renaming `BaseDoc.ts:97` (and its `OTDoc.spec` test) closes it.

Still approved — this doesn't change my read of the change itself.

jacwright (approved):
Approving. All three closed, and the rename resolves the mismatch with dw3#1339 — `optimisticBatchCount` now matches what that branch probes at `pupSync.ts:442`, so `optimistic_batches` will carry real values instead of a permanent `-1`. `git grep optimisticOpsCount` comes back empty across `src` and `tests`, so nothing is left half-renamed. 3432 tests pass across 128 files.

**The `console.error` rationale is the right answer, and I checked it.** `ClientAlgorithm.createDoc<T>(docId, snapshot?)` really does take no options, so there is nowhere to thread an `onSkippedChange`-style hook through, and nothing can have subscribed at construction time. That's a genuine structural difference from `applyChangesForReconstruction`, not a shrug — and "fallback signal for non-Patches hosts and for drops before any subscriber exists" is a better justification than the hook would have been. Writing it down is what stops the next reviewer re-raising it.

The emit-timing paragraph on `onPendingDropped` is exactly the constraint I wanted stated: undefined from `getOpenDoc`, re-entry on `openDoc`, persist elsewhere and return. dw3's shelf already obeys it, so the docs now match the one consumer that exists.

Nothing further from me. This wants publishing so dw3 can pin it — the hydration-drop path stays silent in production until it does.

PR #142, opened 2026-08-07
DAB-930: Merge-result ordering, completion-signal verification, guard refusal purity

jacwright:
## TL;DR

Three pieces of cheap insurance from Alicia's approval notes on #140, landed before the npm publish so the pinned pup version gets them: merge results are now sorted by contract (not just by construction), a buggy store can no longer make a merge silently report completion with changes left behind, and a refused merge now leaves the branch record exactly as untouched as the document itself.

Follow-up to #140 (DAB-930), addressing @matalina's non-blocking approval notes 1–4. The pup-side notes (`maxBytes` read plumbing, `MergePartialProgressError` mapping, frame clear on close) stay with the pin-bump work.

## Changes

- **Result ordering is contractual** (note 3): `mergeBranch` returns `byRev(committed)`, and `MergePartialProgressError.committedChanges` gets the same treatment. Ascending order held by construction on every traced path, but `PatchesSync.applyMergeChanges` takes `result[0]`/`result[last]` as span bounds and applies in order — a violation would mis-apply rather than degrade, so the guarantee shouldn't be emergent.
- **The completion signal is verified** (note 2): an empty window slice — the merge's only completion signal — is checked against `getCurrentRev(branchId)`. A store that returns an empty read while the tip says rows remain now throws instead of ending the merge silently short with the watermark unmoved. A branch edit landing between the two reads is indistinguishable from that store bug, so one transient empty read triggers a re-read: rows resolve the race (the window proceeds with them), a second empty read is the store lying. Symmetric with `trimToByteBudget`'s keep-at-least-one and the `saveChanges` MUST promotion — the merge no longer trusts any single read it can't corroborate.
- **Guard refusals no longer pin** (note 4): `runContentDuplicationGuard` computes its resend-exclusion floor as `mergeBaseRev ?? min(branchedAtRev, tip)` instead of calling `resolveMergeBase`, which writes the clamp pin. A refusal leaves the branch record untouched — the pin (first-writer-wins CAS) happens only once a merge actually proceeds. An unpinned floor can only sit lower than the pin would, which only adds genuinely committed ids to the exclusion corpus — safe in the refusal direction.
- **`maxBytesPerMergeWindow` docs tell the truth** (note 1): the byte cap always binds the transform/commit side (in-memory trim), but binds the read side only where the store honours the `ListChangesOptions.maxBytes` hint — pup ignores it today, so its windows read up to `limit` changes before trimming. Read-side plumbing tracked with the pup work.

## Testing

- `refuses a clamped branch without pinning its merge base` — guard refusal on a branch whose `branchedAtRev` is ahead of the tip leaves `mergeBaseRev` unset.
- `rejects a store whose empty window read contradicts the branch tip` — the lying store fails loudly, watermark untouched.
- `re-reads once when an empty window read races a landing edit` — the benign race completes the merge normally.
- Unit-spec mocks updated so the branch's `getCurrentRev` answers from its mocked log (the completion check made the docId-agnostic tip visible).
- Full suite, lint, format, type-check green.

### DIFF
--- a/src/server/OTBranchManager.ts
+++ b/src/server/OTBranchManager.ts
@@ -103,10 +103,12 @@ export interface OTBranchManagerOptions {
    */
   maxChangesPerMerge?: number;
   /**
-   * Maximum serialized ops bytes per merge window — the second axis of the window bound
-   * (`maxChangesPerMerge` caps document count; this caps memory for branches carrying large
-   * changes). Passed to the store as a read hint (`ListChangesOptions.maxBytes`) and enforced
-   * in memory after the read. Defaults to {@link DEFAULT_MERGE_WINDOW_BYTES}.
+   * Maximum serialized ops bytes (UTF-8) per merge window — the second axis of the window
+   * bound (`maxChangesPerMerge` caps document count). Always caps the transform/commit side:
+   * enforced in memory after the read, so an over-budget read is trimmed before any work.
+   * Caps the read side only when the store honours the `ListChangesOptions.maxBytes` hint —
+   * a store that ignores it still reads up to `maxChangesPerMerge` changes per window before
+   * the trim. Defaults to {@link DEFAULT_MERGE_WINDOW_BYTES}.
    */
   maxBytesPerMergeWindow?: number;
 }
@@ -720,11 +722,13 @@ export class OTBranchManager implements BranchManager {
         cause = error.cause;
       }
       if (committed.length > 0) {
-        throw new MergePartialProgressError(branchId, committed, mergedThroughRev, cause);
+        throw new MergePartialProgressError(branchId, byRev(committed), mergedThroughRev, cause);
       }
       throw cause;
     }
-    return committed;
+    // Ascending rev order held by construction on every path traced, but the consumer takes
+    // result[0]/result[last] as the span bounds and applies in order — make it contractual.
+    return byRev(committed);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


matalina (approved):
Approving — all three hardening items are correctly implemented, each regression test would fail under the old code, and the claims in the description check out against the source. Notes below; the one real finding is narrow, non-blocking, and self-healing.

**Result ordering** — Correct and cheap. `byRev` sorts in place, but both call sites operate on the local `committed` accumulator that's immediately returned or thrown, so the mutation can't alias anything. Windows already produced sorted output, so this is a no-op sort at the boundary — the right call given the consumer takes `result[0]`/`result[last]` as span bounds.

**Completion-signal verification** — Sound; the two properties it depends on both hold:
- Placement after `trimToByteBudget` is safe because the trim keeps at least one change — it can never manufacture a false empty from a non-empty read.
- An honest store honoring the `maxBytes` hint can't false-positive either: the `ListChangesOptions.maxBytes` contract explicitly requires "always returning at least one change". So empty-read + tip-ahead genuinely narrows to {racing edit, store bug}, and the single re-read separates those exactly as the comment argues.
- A contract-violation throw after earlier windows committed gets wrapped into `MergePartialProgressError` by the existing catch — good interaction.
- Cost is one extra `getCurrentRev` on the drained window only. Negligible.

One item for the pup pin-bump checklist: this makes "`getCurrentRev(branchId)` agrees with `listChanges(branchId)`" an *observable* store requirement. If pup derives a branch's `currentRev` from a doc record rather than the change rows `listChanges` serves, any drift now throws mid-merge (loudly, which is the point — but worth verifying they read the same source of truth).

**Guard refusal purity** — The floor `branch.mergeBaseRev ?? min(branchedAtRev, getCurrentRev(source))` is numerically identical to `resolveMergeBase`'s result in every quiescent case, so behavior changes only in the intended way: no pin write on refusal. The regression test is real.

The caveat: "an unpinned floor can only sit lower than the pin" isn't airtight under concurrency. `resolveMergeBase`'s healthy path deliberately re-loads the branch after the tip read to prefer a concurrent merge's freshly written pin; the guard dropped that step. If a concurrent merge pins a clamped base and commits *between the guard's `loadBranch` and its `getCurrentRev`*, the guard's floor sits **above** the pin, the concurrent merge's committed rows fall below `sinceRev`, their ids miss the exclusion corpus, and the stale-watermark walk self-matches them against the head — a spurious refusal. Mitigating: it needs a clamped branch plus a tight interleaving; the failure is loud and retryable; the retry self-heals because a fresh `loadBranch` sees the pin. If you want it airtight, the fix is cheap and stays pure: when the `min()` path was taken, re-load the branch once after the tip read and prefer `fresh.mergeBaseRev` — the read-only half of what `resolveMergeBase` already does. Fine to land as-is or as a follow-up.

Minor, same path: a refused clamped branch no longer emits the clamp `console.warn`, so the data-integrity signal rests on the refusal error alone. Probably fine — refusal is already loud — just noting the signal moved.

**Docs** — Accurate against both the implementation and the `maxBytes` contract; the pup-ignores-the-hint disclosure matches reality.

**Tests** — The mock discriminator `docId === branchId && options?.maxBytes !== undefined` is precise, not brittle: only the window slice reads (and the re-read) pass `maxBytes`; `pageChanges` and `catchUpFrame` don't. The unit-spec rework keeping the mocked store self-consistent is exactly the property the new check observes. The guard-floor concurrency edge above has no test, consistent with it being outside the PR's claimed scope.

**Remaining** — The same trust-the-empty-read pattern still exists in `catchUpFrame`/`pageChanges`, but verifying only the window's empty read is the right scope: it's the only place an empty read silently terminates with the watermark short; the others fail loudly downstream. And the contract-violation throw is a plain `Error` where other merge failures are typed — consistent with leaving error mapping to the pup work, just flag it there if pup needs a distinct status for it.

PR #143, opened 2026-08-07
Bump to 0.27.0

matalina:
Release 0.27.0. Ships everything on main since 0.26.0:

- #139 — `fix(ot)`: estimate split points instead of compressing at every boundary (DAB-931)
- #140 — merge branches in bounded windows, carrying the OT frame between them (DAB-930)
- #141 — `feat(client)`: surface pending changes dropped at hydration (DAB-854) — adds `Patches.onPendingDropped`, `BaseDoc.droppedPendingChanges`, `BaseDoc.optimisticBatchCount`
- #142 — DAB-930: merge-result ordering, completion-signal verification, guard refusal purity

Minor bump: #141 adds public API.

**⛔ Publish gate: `npm publish` only after BOTH this PR and #142 are on main.** Jacob's ordering (#development, 15:29): publish + pup pin bump after #142 lands. Publishing from main with only this PR merged would ship 0.27.0 without #142's ordering guarantee, completion-signal verification, and guard-refusal fix.

After both merge: `npm publish` from a clean checkout of main (`prepare` builds dist). Then TWO pin bumps: **pup** (per Jacob) and **dabble-writer-3.0** (activates the hydration-drop shelving behind dabblewriter/dabble-writer-3.0#1339's feature-detected hook).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dabble/patches",
-  "version": "0.26.0",
+  "version": "0.27.0",
   "description": "Immutable JSON Patch implementation based on RFC 6902 supporting operational transformation and last-writer-wins",
   "author": "Jacob Wright <jacwright@gmail.com>",
   "bugs": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@dabble/patches",
-  "version": "0.26.0",
+  "version": "0.27.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@dabble/patches",
-      "version": "0.26.0",
+      "version": "0.27.0",
       "license": "MIT",
       "dependencies": {
         "@dabble/delta": "^1.2.5",


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dabble/patches",
-  "version": "0.26.0",
+  "version": "0.27.0",
```
jacwright, line 3: The release instruction in the description is a step short of what Jacob asked for.

It says "After merge: `npm publish` from a clean checkout of main". But **#142 is open, approved, and explicitly wanted before the publish** — his words in #development at 15:29 were "Publish + pup pin bump after it lands." It's `CLEAN`/`MERGEABLE` right now.

Merging *this* PR first is harmless — the version number doesn't need to change again, and #142 landing on top still goes out as 0.27.0. The hazard is only the window between the two: anyone following the description literally, before #142 merges, publishes a 0.27.0 without the merge-result ordering guarantee, the completion-signal verification, or the guard-refusal fix.

Worth a line in the body making the gate explicit — publish once **both** this and #142 are on main. Same reason the contents list is worth a second pass at publish time: once #142 lands, 0.27.0 will contain a fourth PR the description doesn't mention, and this body is the artifact people will read later to work out what a version shipped.

jacwright (approved):
Approving. The bump is correct and complete, which for a release PR is the whole job.

**Both version fields in `package-lock.json` moved** — the root at line 3 and `packages.""` at line 9 — alongside `package.json`. That's the failure this repo's neighbour hit last week: dw3#1316 shipped a release branch whose lockfile disagreed with its manifest, and CI caught it only because a dependent import broke. Nothing here disagrees. The other `0.25.2` strings in the lock are esbuild platform packages, untouched and unrelated.

**The contents list is accurate.** Everything on `main` since the 0.26.0 bump (`fea74d4`) is exactly #139, #140 and #141 — three merge commits, no strays. Minor rather than patch is right: #141 adds `Patches.onPendingDropped`, `BaseDoc.droppedPendingChanges` and `BaseDoc.optimisticBatchCount` to the public surface.

One note inline about the release instruction.

package.json
matalina: Right on both counts — #142 wasn't on my radar and the body would have led someone to publish inside the gap. Body updated: #142 added to the contents, and an explicit publish gate (both PRs on main first, per Jacob's #development note). Merging #142 first, then this.

PR #145, opened 2026-08-12
DAB-951: flush stragglers at their true baseRev instead of relabeling them

jacwright:
**TL;DR:** Fixes the second known way Dabble could permanently corrupt a document's history. When a user typed at the exact moment their app was catching up with the server (or typed in a second tab that hadn't caught up yet), one of their edits could be saved to the server against the wrong version of the document — producing a change that no device can ever apply, wedging the document for everyone in it (the same failure class we repaired 37 documents for under DAB-946). With this fix that edit is instead sent honestly, the server places it correctly, and the document stays healthy. Should ride the same patches release as #136/#137.

## Merge order

## The defect

`OTAlgorithm._withConsistentBaseRev` raised a straggler change's `baseRev` to the queue max **without transforming its ops** across the intervening committed changes `(trueBase, maxBase]`. The server then had nothing to transform it against — the ops committed verbatim in a frame they were never expressed in. That is committed poison: strict replay throws at it forever (DABBLE-PUP-5 detections on `Qp8r62vVKSKHcPE5` 8/7 and `jXvONedGC8Sz2JmU` 8/8 confirm it live in prod).

The mixed queue that triggers it forms whenever a mint lands while its doc is a frame behind the store: the single-tab reconcile/reload window (mint between `reconcilePending`'s store write and the doc re-import), or a follower tab minting before the writer's broadcast reaches it.

## Why not transform on the client

The transform the relabel skipped genuinely cannot be run at the flush seam: the committed span the straggler missed is already collapsed into local state (client stores don't retain committed changes), and the straggler's frame may also include *old copies* of pending predecessors that the store has since rebased in place. The one party holding every committed change past any `baseRev` is the server — so the fix is the issue's second option: **commit the straggler against its true baseRev** and let the server's normal commit transform do the work.

## The fix (three seams, one invariant)

A pending row's `baseRev` now stays truthful: it only advances when the row's ops are transformed across exactly the span being crossed.

1. **Flush seam** (`_withConsistentBaseRev`): a mixed-baseRev queue is sliced to its leading same-baseRev run — a coherent sequential program — and flushed at its true baseRev. Deferred rows go out in follow-up passes (still warns, greppable `[patches] Mixed baseRev`).
2. **Receive seam** (`applyServerChanges` / `reconcilePending` via new `_rebasePendingPreservingFrameDebt`): without this, `rebaseChanges`' unconditional renumber would relabel the deferred straggler to the new tip on the very next echo — re-minting the same poison one commit later. A row whose `baseRev` is behind the frame the incoming tail attaches to is neither walked nor relabeled: it keeps its ops and true frame (dropped only when the tail echoes its id), and is re-sequenced into its queue position. Frame-consistent queues take the plain `rebaseChanges` path unchanged. The preservation deliberately lives at the two queue-rebase call sites, not inside `rebaseChanges` — the ejection path and `OTDoc` feed that function synthetic carriers with meaningless baseRevs.
3. **Drain** (`PatchesSync.flushDoc`): when `getPendingToSend` still returns flushable rows after a flush, queue one serial-gated follow-up pass (the existing `reloadedMidFlush` pattern), so a deferred run doesn't sit unsynced until the next keystroke or reconnect.

`computePendingEjection` gets the same honesty fix: a stale-frame predecessor keeps its true baseRev through an ejection instead of being relabeled (it already warned on this trigger).

## Known bounded degradation

When a straggler's mint frame included old copies of pending predecessors that have since committed, its solo flush is transformed against those commits as foreign (the batch carries neither their ids nor their batchId), which can double-shift overlapping offsets — content lands slightly misplaced. That is the theoretical worst case of the conservative option; the previous behavior corrupted the document permanently in the same situation, and when the straggler's frame was clean (the common case: a follower tab with an empty queue, or the reconcile-window mint at queue head) the server transform is exactly right.

## Tests

- `tests/client/OTAlgorithm-frameDebt.spec.ts` (new, runs against the real server `commitChanges` + transform machinery, no mocks): reproduces the poison mint — a mixed queue whose straggler is unappliable if relabeled — and asserts both changes commit, the full committed log replays under strict apply, and client and server converge on the same head; plus receive-side preservation, echo-drop, and `reconcilePending` coverage, and a consistent-queue control.
- `tests/client/OTAlgorithm-getPendingToSend.spec.ts`: rewritten from asserting the relabel to asserting the slice (leading-run, straggler-at-head, adjacent same-frame stragglers stay one run).
- Full suite: 134 files, 3521 passed / 4 skipped. `tsc --noEmit`, eslint, prettier all clean.

## Release note

Should ship in the same patches release as #136 (poison skip floor) and #137 (durable-queue flush): #136 heals the read side for poison already committed, this stops the single-tab generator, #137 stops the multi-tab reorder generator.

### DIFF
--- a/src/algorithms/ot/shared/ejectPendingChange.ts
+++ b/src/algorithms/ot/shared/ejectPendingChange.ts
@@ -29,7 +29,8 @@ export interface PendingEjection {
  * inverse through the successors. Predecessors are untouched. Every survivor is then
  * renumbered contiguously off `committedRev`, preserving the OT pending invariant that all
  * pending share `baseRev === committedRev` with sequential revs (see
- * `OTAlgorithm._withConsistentBaseRev`).
+ * `OTAlgorithm._withConsistentBaseRev`). A predecessor minted a frame behind keeps its true
+ * baseRev instead — see the frame-debt note at the renumber below.
  *
  * The server accepts `newPending` as a valid poison-free queue and both sides converge on it
  * deterministically. Exact tie-resolution follows the same one-sided transform as a normal
@@ -107,24 +108,23 @@ export function computePendingEjection(
   // successor whose ops transformed away to nothing.
   //
   // A predecessor with a stale baseRev is the same mint/rebase race
-  // `OTAlgorithm._withConsistentBaseRev` re-stamps — warn on the same trigger rather than
-  // silently masking it here (the re-stamp can misplace ops if foreign changes landed in
-  // between; the strict probe above only vouches for the poison's frame, not the queue's).
+  // `OTAlgorithm._withConsistentBaseRev` defers at the flush seam. Its true baseRev is
+  // preserved here for the same reason: relabeling it to committedRev would commit its ops
+  // in a frame they were never transformed into (DAB-951). Warn — the strict probe above
+  // only vouches for the poison's frame, not the queue's.
   const staleBaseRevs = before.filter(change => change.baseRev !== committedRev);
   if (staleBaseRevs.length > 0) {
     console.warn(
-      `[patches] Ejection of ${changeId} is re-stamping ${staleBaseRevs.length} predecessor(s) from baseRev ` +
-        `${staleBaseRevs.map(c => c.baseRev).join(',')} to ${committedRev}. This indicates a mint/rebase ` +
-        `race (likely two client instances over one store) and can misplace ops if foreign ` +
-        `changes landed in between.`
+      `[patches] Ejection of ${changeId} has ${staleBaseRevs.length} predecessor(s) on older frame(s) (baseRev ` +
+        `${staleBaseRevs.map(c => c.baseRev).join(',')} vs committed ${committedRev}) — a mint/rebase race ` +
+        `(likely two client instances over one store). Their true baseRev is preserved; they flush separately.`
     );
   }
   let rev = committedRev;
-  const newPending = [...before, ...rebasedAfter].map(change => ({
-    ...change,
-    baseRev: committedRev,
-    rev: ++rev,
-  }));
+  const newPending = [
+    ...before.map(change => ({ ...change, rev: ++rev })),
+    ...rebasedAfter.map(change => ({ ...change, baseRev: committedRev, rev: ++rev })),
+  ];
 
   return { poison, newPending };
 }
--- a/src/algorithms/ot/shared/ejectPendingChange.ts
+++ b/src/algorithms/ot/shared/ejectPendingChange.ts
@@ -26,11 +26,11 @@ export interface PendingEjection {
  * server change: the ejected change genuinely *preceded* its successors, so its inverse is
  * the "already-happened" side those successors' position ties yield to. We invert the
  * ejected change against the state it applied to (committed + predecessors) and walk that
- * inverse through the successors. Predecessors are untouched. Every survivor is then
- * renumbered contiguously off `committedRev`, preserving the OT pending invariant that all
- * pending share `baseRev === committedRev` with sequential revs (see
- * `OTAlgorithm._withConsistentBaseRev`). A predecessor minted a frame behind keeps its true
- * baseRev instead — see the frame-debt note at the renumber below.
+ * inverse through the successors that share the ejected change's frame. Predecessors are
+ * untouched. Every survivor is then renumbered contiguously off `committedRev`, preserving the
+ * OT pending invariant that all pending share `baseRev === committedRev` with sequential revs
+ * (see `OTAlgorithm._withConsistentBaseRev`). A row minted a frame behind — on either side of
+ * the poison — keeps its true baseRev and sits out the walk; see the frame-debt notes below.
  *
  * The server accepts `newPending` as a valid poison-free queue and both sides converge on it
  * deterministically. Exact tie-resolution follows the same one-sided transform as a normal
@@ -99,32 +99,52 @@ export function computePendingEjection(
     // survivors are renumbered below — only the diamond walk matters. A fresh id keeps it
     // out of the successors' id set so rebaseChanges walks it rather than dropping it.
     const inverseCarrier = createChange(poison.baseRev, poison.rev, invertedOps, {});
-    rebasedAfter = rebaseChanges([inverseCarrier], after);
+    // Only successors minted in the poison's OWN frame were stacked on it. A successor from a
+    // lagging context (a different baseRev) never had the poison in frame, so the inverse —
+    // computed in the poison's frame — must not walk through it. Same rule the receive path
+    // applies (see `OTAlgorithm._rebasePendingPreservingFrameDebt`); such a row rides through
+    // untouched below.
+    rebasedAfter = rebaseChanges(
+      [inverseCarrier],
+      after.filter(change => change.baseRev === poison.baseRev)
+    );
   }
 
-  // Renumber survivors contiguously off committedRev. Predecessors already sit on
-  // committedRev with these very revs, so this is a no-op for them and only re-seats the
-  // rebased successors into the gap the poison left. rebaseChanges has already dropped any
-  // successor whose ops transformed away to nothing.
+  // Renumber survivors contiguously off committedRev, re-seating the rebased successors into
+  // the gap the poison left. rebaseChanges has already dropped any successor whose ops
+  // transformed away to nothing.
   //
-  // A predecessor with a stale baseRev is the same mint/rebase race
-  // `OTAlgorithm._withConsistentBaseRev` defers at the flush seam. Its true baseRev is
-  // preserved here for the same reason: relabeling it to committedRev would commit its ops
-  // in a frame they were never transformed into (DAB-951). Warn — the strict probe above
-  // only vouches for the poison's frame, not the queue's.
-  const staleBaseRevs = before.filter(change => change.baseRev !== committedRev);
+  // Every survivor keeps its OWN baseRev. On the invariant queue — all of it on committedRev —
+  // that is the same result as stamping committedRev on each. A row minted a frame behind (the
+  // same mint/rebase race `OTAlgorithm._withConsistentBaseRev` defers at the flush seam) keeps
+  // its true baseRev instead, on both sides of the poison: relabeling it would commit its ops
+  // in a frame they were never transformed into (DAB-951) — the very poison class this
+  // function exists to remove. Warn on both halves — the strict probe above only vouches for
+  // the poison's frame, not the queue's.
+  //
+  // Such a row is left with a rev/baseRev gap. Nothing reads that pair as a frame: rev only has
+  // to stay strictly increasing across the queue (it does — every survivor takes the next one),
+  // getPendingToSend slices the flush on baseRev and ranges the store read on rev
+  // independently, and the server reassigns rev on commit.
+  const staleBaseRevs = [...before, ...after].filter(change => change.baseRev !== committedRev);
   if (staleBaseRevs.length > 0) {
     console.warn(
-      `[patches] Ejection of ${changeId} has ${staleBaseRevs.length} predecessor(s) on older frame(s) (baseRev ` +
+      `[patches] Ejection of ${changeId} has ${staleBaseRevs.length} queue neighbour(s) on older frame(s) (baseRev ` +
         `${staleBaseRevs.map(c => c.baseRev).join(',')} vs committed ${committedRev}) — a mint/rebase race ` +
         `(likely two client instances over one store). Their true baseRev is preserved; they flush separately.`
     );
   }
+  const rebasedById = new Map(rebasedAfter.map(change => [change.id, change]));
   let rev = committedRev;
-  const newPending = [
-    ...before.map(change => ({ ...change, rev: ++rev })),
-    ...rebasedAfter.map(change => ({ ...change, baseRev: committedRev, rev: ++rev })),
-  ];
+  const newPending: Change[] = before.map(change => ({ ...change, rev: ++rev }));
+  for (const change of after) {
+    const rebased = rebasedById.get(change.id);
+    if (rebased) newPending.push({ ...rebased, baseRev: change.baseRev, rev: ++rev });
+    // Absent from the walk's output for one of two reasons: it was left out of the walk (a
+    // different frame — ride it through untouched), or it was walked and transformed away to
+    // nothing under the inverse (drop it, as before).
+    else if (change.baseRev !== poison.baseRev) newPending.push({ ...change, rev: ++rev });
+  }
 
   return { poison, newPending };
 }
--- a/src/client/ClientAlgorithm.ts
+++ b/src/client/ClientAlgorithm.ts
@@ -89,7 +89,9 @@ export interface ClientAlgorithm {
 
   /**
    * Gets pending data to send to the server.
-   * - OT: Returns all pending changes (may batch)
+   * - OT: Returns the pending queue's leading run of same-baseRev changes — usually the whole
+   *   queue; a mixed-baseRev queue flushes one frame per call (callers flush repeatedly until
+   *   drained, see PatchesSync.flushDoc's follow-up pass)
    * - LWW: Creates single Change from pendingFields (or returns existing)
    *
    * When `doc` (the open doc) is passed, OT trusts its in-memory pending as the source of truth
--- a/src/client/OTAlgorithm.ts
+++ b/src/client/OTAlgorithm.ts
@@ -240,7 +240,7 @@ export class OTAlgorithm implements ClientAlgorithm {
           const staleIds = new Set(staleServerChanges.map(c => c.id));
           pendingSet = pendingSet.filter(c => !staleIds.has(c.id));
         }
-        rebased = newServerChanges.length > 0 ? rebaseChanges(newServerChanges, pendingSet) : pendingSet;
+        rebased = this._rebasePendingPreservingFrameDebt(newServerChanges, pendingSet, committedRev);
         const result = await this.store.applyServerChanges(docId, serverChanges, rebased, tailRev);
         if (result !== 'conflict') {
           applied = true;
@@ -321,11 +321,13 @@ export class OTAlgorithm implements ClientAlgorithm {
       if (pending.length === 0) return;
       const tailRev = pending[pending.length - 1].rev;
 
-      // rebaseChanges drops pending the server already committed (matched by id) and transforms
-      // the survivors into the tail's frame — a pure op transform that never applies the tail, so
+      // Drops pending the server already committed (matched by id) and transforms the
+      // survivors into the tail's frame — a pure op transform that never applies the tail, so
       // it is safe even when the local committed state is corrupt (which is why the
-      // snapshot-reload recovery calling this exists at all).
-      const rebased = rebaseChanges(committedChanges, pending);
+      // snapshot-reload recovery calling this exists at all). The tail starts at the frame the
+      // queue sits on (see the interface contract), so tail[0].rev - 1 anchors the frame-debt
+      // check: a row minted a frame behind keeps its true baseRev instead of being relabeled.
+      const rebased = this._rebasePendingPreservingFrameDebt(committedChanges, pending, committedChanges[0].rev - 1);
 
       // Install the reconciled tail AND swap the pending queue in ONE store transaction, retrying
       // if a foreign mint raced the replace (R2).
@@ -538,25 +540,68 @@ export class OTAlgorithm implements ClientAlgorithm {
   }
 
   /**
-   * The server requires every change in one flush batch to share a baseRev. A torn reload —
-   * reconcilePending advances the store's committed tail (rebasing pending to the new frame),
-   * then saveDoc/import faults before the open doc follows — leaves the doc a frame behind the
-   * store, so the next mint stamps its baseRev off the stale doc while the store queue already
-   * moved on. Neither the in-txn rev mint nor the conflict-safe replace expresses baseRev, so
-   * normalize the outgoing batch here: bump any straggler up to the freshest frame present (the
-   * majority, already correctly rebased). The stragglers' ops are then a frame behind their
-   * label — the server transforms them as concurrent edits and every replica converges the same
-   * way; the commit echo rebases the stored copies into agreement. A consistent queue is a no-op.
+   * The server requires every change in one flush batch to share a baseRev — the batch is a
+   * sequential program expressed against that committed frame. The queue can carry rows from
+   * more than one frame: a mint lands while its doc is a frame behind the store (a torn reload,
+   * a follower tab that hasn't received the writer's broadcast yet), so its baseRev — and its
+   * ops — sit on an older frame than siblings the receive path already rebased.
+   *
+   * Those frames are not interchangeable. Relabeling a straggler to the newest frame commits
+   * its ops WITHOUT the transform across the intervening committed changes — committed history
+   * that can never apply (the DAB-946 poison class; DAB-951). The client cannot run that
+   * transform here: the committed span the straggler missed is already collapsed into local
+   * state. The server can — it holds every committed change past any baseRev — so flush one
+   * frame at a time: return the queue's leading run of same-baseRev changes at its TRUE
+   * baseRev and leave the rest pending (flushDoc queues a follow-up pass for them; see also
+   * {@link _rebasePendingPreservingFrameDebt}, which keeps a deferred row's frame honest
+   * across receives). A consistent queue is a no-op.
    */
   private _withConsistentBaseRev(docId: string, batch: Change[]): Change[] {
-    let maxBase = batch[0].baseRev;
-    for (const c of batch) if (c.baseRev > maxBase) maxBase = c.baseRev;
-    if (batch.every(c => c.baseRev === maxBase)) return batch;
+    const baseRev = batch[0].baseRev;
+    let end = 1;
+    while (end < batch.length && batch[end].baseRev === baseRev) end++;
+    if (end === batch.length) return batch;
     console.warn(
-      `[patches] Normalizing ${batch.filter(c => c.baseRev !== maxBase).length} pending change(s) for ${docId} to ` +
-        `baseRev ${maxBase} for a consistent flush (torn-reload straggler).`
+      `[patches] Mixed baseRev in pending queue for ${docId}: flushing ${end} change(s) at baseRev ${baseRev}, ` +
+        `deferring ${batch.length - end} on other frame(s) to a follow-up flush (DAB-951).`
     );
-    return batch.map(c => (c.baseRev === maxBase ? c : { ...c, baseRev: maxBase }));
+    return batch.slice(0, end);
+  }
+
+  /**
+   * Rebase the pending queue against a committed server tail without laundering frame debt.
+   * `frameRev` is the committed frame the queue sits on — the frame `serverChanges` extends.
+   *
+   * A row minted a frame behind (`baseRev < frameRev`, see {@link _withConsistentBaseRev}) is
+   * NOT in the frame this walk crosses: transforming it against `serverChanges` and relabeling
+   * it to the new tip — what {@link rebaseChanges} does to every survivor — would silently
+   * advance its label across the span it was already behind on, recreating the mislabeled
+   * commit the flush seam refuses. Such rows keep their ops and true baseRev (dropped only when
+   * `serverChanges` echoes their id — a true-baseRev flush coming back committed) and are
+   * re-sequenced into their queue position; the server transforms them across everything past
+   * their baseRev when they flush. Later rows minted by OTHER contexts were expressed without
+   * the straggler in frame, so the walk must not advance the server ops through it either —
+   * the straggler is skipped entirely, not walked. A frame-consistent queue (the invariant
+   * case) takes the plain {@link rebaseChanges} path unchanged.
+   */
+  private _rebasePendingPreservingFrameDebt(serverChanges: Change[], pending: Change[], frameRev: number): Change[] {
+    if (serverChanges.length === 0 || pending.length === 0) return pending;
+    const staleIds = new Set(pending.filter(c => c.baseRev < frameRev).map(c => c.id));
+    if (staleIds.size === 0) return rebaseChanges(serverChanges, pending);
+
+    const serverIds = new Set(serverChanges.map(c => c.id));
+    const rebased = rebaseChanges(
+      serverChanges,
+      pending.filter(c => !staleIds.has(c.id))
+    );
+    const rebasedById = new Map(rebased.map(c => [c.id, c]));
+    let rev = serverChanges[serverChanges.length - 1].rev;
+    const result: Change[] = [];
+    for (const c of pending) {
+      const row = staleIds.has(c.id) ? (serverIds.has(c.id) ? undefined : c) : rebasedById.get(c.id);
+      if (row) result.push({ ...row, rev: ++rev });
+    }
+    return result;
   }
 
   /**
--- a/src/net/PatchesSync.ts
+++ b/src/net/PatchesSync.ts
@@ -1175,10 +1175,13 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
       this._resplitBudgets.delete(docId);
       const stillHasPending = await algorithm.hasPending(docId);
       this._updateDocSyncState(docId, { hasPending: stillHasPending, syncStatus: 'synced' });
-      // Send the remainder the reload rebased, from the store rather than from `batches`. Sync is
+      // Send the remainder, from the store rather than from `batches`: after a mid-flush reload
+      // that rebased the queue, or when getPendingToSend still returns flushable rows — a
+      // mixed-baseRev queue flushes one frame per pass (see OTAlgorithm._withConsistentBaseRev),
+      // and without this pass a deferred run would wait for the next mint or reconnect. Sync is
       // serial-gated, so this queues exactly one follow-up pass instead of recursing, and that
       // pass flushes on the reloaded committedRev, which no longer answers docReloadRequired.
-      if (reloadedMidFlush && stillHasPending) void this.syncDoc(docId);
+      if ((reloadedMidFlush || pending.length > 0) && stillHasPending) void this.syncDoc(docId);
     } catch (err) {
       if (this._isDocDeletedError(err)) {
         await this._handleRemoteDocDeleted(docId);
--- a/src/net/PatchesSync.ts
+++ b/src/net/PatchesSync.ts
@@ -1055,6 +1055,16 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
         return; // Nothing to flush
       }
 
+      // The follow-up pass at the end of this method re-arms on every flush now, so it needs a
+      // progress signal — without one, any state where a run commits but the queue does not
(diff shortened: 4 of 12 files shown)


src/client/OTAlgorithm.ts
```diff
@@ -538,25 +540,68 @@ export class OTAlgorithm implements ClientAlgorithm {
...
+    const rebased = rebaseChanges(
+      serverChanges,
+      pending.filter(c => !staleIds.has(c.id))
     );
-    return batch.map(c => (c.baseRev === maxBase ? c : { ...c, baseRev: maxBase }));
+    const rebasedById = new Map(rebased.map(c => [c.id, c]));
+    let rev = serverChanges[serverChanges.length - 1].rev;
+    const result: Change[] = [];
+    for (const c of pending) {
+      const row = staleIds.has(c.id) ? (serverIds.has(c.id) ? undefined : c) : rebasedById.get(c.id);
```
matalina, line 601: **The preserved straggler is still handed to the open doc, where strict apply throws on it.**

Keeping the ops honest for the *server* is right. But the same untransformed row goes into the doc's pending queue on this path, and the doc replays pending strictly against the newly advanced committed state:

- `applyServerChanges` writes the store first (:246), then calls `otDoc.applyChanges([...serverChanges, ...rebased])` (:258).
- `OTDoc.applyChanges` sets `this._pendingChanges = rebasedPending` (`OTDoc.ts:298`) and calls `_recomputeState()` (:301).
- `_recomputeState`'s first line is `applyChangesToState(this._committedState, this._pendingChanges)` (`OTDoc.ts:177`) — the **strict** `applyChanges`, with no try/catch. Only the `_optimisticOps` loop below it is guarded.

Run this PR's own reproduction through it: the straggler is `add /items/3` minted against a 3-element frame, preserved untransformed while committed state advances to 1-2 elements. Strict apply throws `ApplyChangesError` out of `_recomputeState` → out of `applyServerChanges` — **after** the store transaction committed. Store advanced, doc not: the exact torn-reload asymmetry that mints stragglers in the first place.

Downstream is worse. `PatchesSync` reads that as an apply failure and reloads; on re-hydration `OTDoc`'s constructor strict-apply filter (`OTDoc.ts:66-74`) drops the straggler into `droppedPendingChanges` — the user's unsent edit, gone. So the change that was too precious to mislabel gets discarded locally instead.

Nothing in the suite catches this because every receive-side test passes `doc: undefined` and `OTInMemoryStore.getDoc` materializes committed rows only.

The frame debt has to be paid somewhere for the doc's optimistic state. Two shapes that keep the wire honest:

- keep the *stored* row at its true baseRev (as now), but hand the doc a transformed copy for its optimistic queue — the doc's queue is a display concern, the store's is the wire contract; or
- transform the straggler for both and record the true baseRev in a separate field the flush seam reads, so `baseRev` on the wire stays truthful without the ops being stranded in a dead frame.

src/client/OTAlgorithm.ts
```diff
@@ -538,25 +540,68 @@ export class OTAlgorithm implements ClientAlgorithm {
...
+   * `serverChanges` echoes their id — a true-baseRev flush coming back committed) and are
+   * re-sequenced into their queue position; the server transforms them across everything past
+   * their baseRev when they flush. Later rows minted by OTHER contexts were expressed without
+   * the straggler in frame, so the walk must not advance the server ops through it either —
+   * the straggler is skipped entirely, not walked. A frame-consistent queue (the invariant
+   * case) takes the plain {@link rebaseChanges} path unchanged.
+   */
+  private _rebasePendingPreservingFrameDebt(serverChanges: Change[], pending: Change[], frameRev: number): Change[] {
+    if (serverChanges.length === 0 || pending.length === 0) return pending;
+    const staleIds = new Set(pending.filter(c => c.baseRev < frameRev).map(c => c.id));
```
matalina, line 589: **When the whole queue is stale, nothing gets rebased at all.**

`staleIds` can cover every row. Then `rebased = rebaseChanges(serverChanges, [])` is `[]`, `rebasedById` is empty, and the loop below re-emits all of `pending` untouched. So this isn't "one straggler is preserved" — it's "the receive path stops transforming the queue entirely".

That's the torn-reload trigger this PR names, with an empty queue at the moment of the tear: the doc is a frame behind, every subsequent mint stamps the stale `baseRev`, so 100% of pending sits below `frameRev`. Each later echo then leaves the whole queue in a frame the committed state has moved past, and the debt compounds with every receive.

Combined with the strict `_recomputeState` above, the blast radius is the entire unsent queue rather than one change.

Worth bounding: if the stale set is the whole queue, there's no coherent optimistic state left to preserve, and routing to a reload is more honest than accumulating debt indefinitely. A cheap version is to cap how far behind `frameRev` a row may drift before the doc gives up on it.

src/algorithms/ot/shared/ejectPendingChange.ts
```diff
@@ -107,24 +108,23 @@ export function computePendingEjection(
...
   }
   let rev = committedRev;
-  const newPending = [...before, ...rebasedAfter].map(change => ({
-    ...change,
-    baseRev: committedRev,
-    rev: ++rev,
-  }));
+  const newPending = [
+    ...before.map(change => ({ ...change, rev: ++rev })),
+    ...rebasedAfter.map(change => ({ ...change, baseRev: committedRev, rev: ++rev })),
```
matalina: **The honesty fix stops at `before` — successors are still relabeled without the transform.**

`rebasedAfter` rows still get `baseRev: committedRev` stamped on. For a successor that was itself minted a frame behind, that is precisely the relabel-without-transform this PR exists to remove.

Concretely, queue `[fresh(N), poison(N), straggler(M)]` with M < N: `before = [fresh]`, `after = [straggler]`. `rebaseChanges([inverseCarrier], after)` transforms the straggler against the poison's inverse — itself computed in the *poison's* frame, not the straggler's — and then this line labels it `committedRev`. The result is a change whose ops were never transformed across `(M, committedRev]`: DAB-951 poison, minted by the ejection recovery path.

Ejection runs when a queue is already sick, so a mixed-frame queue is *more* likely here than in the steady state, not less.

The `staleBaseRevs` warn just above also only inspects `before`, so a stale successor isn't even reported — the operator sees a clean ejection.

```suggestion
    ...rebasedAfter.map(change => ({
      ...change,
      // Same frame-debt rule as the predecessors above: only relabel a row that was already
      // on the committed frame. A successor minted a frame behind keeps its true baseRev and
      // flushes separately (DAB-951).
      ...(change.baseRev === committedRev ? { baseRev: committedRev } : {}),
      rev: ++rev,
    })),
```

…and `staleBaseRevs` should be computed over `[...before, ...after]` so the warn covers both halves.

src/net/PatchesSync.ts
```diff
@@ -1175,10 +1175,13 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
...
       this._updateDocSyncState(docId, { hasPending: stillHasPending, syncStatus: 'synced' });
-      // Send the remainder the reload rebased, from the store rather than from `batches`. Sync is
+      // Send the remainder, from the store rather than from `batches`: after a mid-flush reload
+      // that rebased the queue, or when getPendingToSend still returns flushable rows — a
+      // mixed-baseRev queue flushes one frame per pass (see OTAlgorithm._withConsistentBaseRev),
+      // and without this pass a deferred run would wait for the next mint or reconnect. Sync is
       // serial-gated, so this queues exactly one follow-up pass instead of recursing, and that
       // pass flushes on the reloaded committedRev, which no longer answers docReloadRequired.
-      if (reloadedMidFlush && stillHasPending) void this.syncDoc(docId);
+      if ((reloadedMidFlush || pending.length > 0) && stillHasPending) void this.syncDoc(docId);
```
matalina: **The follow-up trigger loses its one-shot property, with no progress check to replace it.**

`reloadedMidFlush` could fire at most once per flush. `pending.length > 0 && stillHasPending` re-arms on every pass, and nothing verifies the previous pass actually drained anything.

Any state where a run commits but the queue doesn't shrink becomes a continuous spin: an echo that never carries the ids (`dropResolvedPending` failing to match), a straggler the server rebases away without echoing it back, a store write that silently no-ops. Serial-gating stops re-entrancy, not the loop — so it's an unbounded commit + IndexedDB cycle with no error and nothing user-visible.

This matters more with the slicing above, since a mixed queue is now *expected* to leave rows behind after a successful pass, which is exactly the shape that used to mean "something is wrong".

Capturing the queue head before the pass and only re-arming when it moved bounds it:

```ts
const headBefore = pending[0] && `${pending[0].id}@${pending[0].rev}`;
// ...after the loop
const next = await algorithm.getPendingToSend(docId, this.patches.getOpenDoc(docId) as PatchesDoc<any>);
const headAfter = next?.[0] && `${next[0].id}@${next[0].rev}`;
if ((reloadedMidFlush || (headAfter && headAfter !== headBefore)) && stillHasPending) void this.syncDoc(docId);
```

tests/client/OTAlgorithm-frameDebt.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  const foreign = (rev: number): Change => ({
+    ...createChange(rev - 1, rev, [{ op: 'remove', path: '/items/0' }]),
+    committedAt: Date.now(),
+  });
+
+  it('applyServerChanges: walks current-frame rows, leaves the straggler ops and baseRev intact', async () => {
+    const { store, algorithm, fresh, straggler } = await makeMixedQueue();
+
+    await algorithm.applyServerChanges(DOC_ID, [foreign(4)], undefined);
```
matalina, line 153: Every receive-side test here passes `undefined` for `doc` (:147, :166, and `flushOnce` at :33), so the open-doc half of the path never runs — which is where the preserved straggler actually causes trouble (see my note on `OTAlgorithm.ts:601`).

The store-level assertions can't catch it either: `OTInMemoryStore.getDoc` materializes `applyChanges(snapshot.state, committed)` — committed rows only, never pending. So `queue[1].ops` staying `add /items/3` and `baseRev` staying 1 both pass while that same row is unappliable in the doc's optimistic frame.

`makeMixedQueue`'s straggler should surface it immediately with a real doc:

```ts
const doc = algorithm.createDoc<any>(DOC_ID, (await algorithm.loadDoc(DOC_ID)) as any);
await algorithm.applyServerChanges(DOC_ID, [foreign(4)], doc);
expect(doc.state).toEqual(/* what should the user see here? */);
```

Whatever that assertion turns out to be is the missing half of the design — right now the PR specifies the wire contract precisely and leaves the local optimistic contract unstated.

src/client/OTAlgorithm.ts
```diff
@@ -538,25 +540,68 @@ export class OTAlgorithm implements ClientAlgorithm {
...
+    let end = 1;
+    while (end < batch.length && batch[end].baseRev === baseRev) end++;
+    if (end === batch.length) return batch;
     console.warn(
-      `[patches] Normalizing ${batch.filter(c => c.baseRev !== maxBase).length} pending change(s) for ${docId} to ` +
-        `baseRev ${maxBase} for a consistent flush (torn-reload straggler).`
+      `[patches] Mixed baseRev in pending queue for ${docId}: flushing ${end} change(s) at baseRev ${baseRev}, ` +
+        `deferring ${batch.length - end} on other frame(s) to a follow-up flush (DAB-951).`
+    );
+    return batch.slice(0, end);
```
matalina, line 568: Integration note for the three-PR release: stacked with #137 this multiplies store reads on the drain path.

#137 replaces the ranged store read in `getPendingToSend` with an unranged `store.getPendingChanges(docId)` inside `_collectPending`, and `flushDoc` already calls `getPendingToSend` once per batch. Adding this PR's per-frame follow-up pass makes it F frames × B batches **full pending-table scans** to drain one queue — on the reconnect path, on devices already showing multi-second IndexedDB stalls (DAB-766).

#137 also widens the batch to include foreign rows below the mirror tail, so mixed-baseRev queues become the ordinary multi-tab case rather than the torn-reload rarity this slicing was sized for. That's the right pairing — #137 surfaces more stragglers, this PR stops them being poison — but it means the follow-up loop will be exercised far more than the reproduction here suggests.

Both PRs touch adjacent regions of `getPendingToSend`/`_withConsistentBaseRev`, so a clean textual auto-merge won't tell you much. Worth stating the intended merge order in the description and re-running `OTAlgorithm-frameDebt.spec.ts` and `PatchesSyncQueueOrdering.spec.ts` on the combined tree before the release.

src/client/ClientAlgorithm.ts
```diff
@@ -89,7 +89,9 @@ export interface ClientAlgorithm {
 
   /**
    * Gets pending data to send to the server.
-   * - OT: Returns all pending changes (may batch)
+   * - OT: Returns the pending queue's leading run of same-baseRev changes — usually the whole
```
matalina, line 92: While you're in this docblock: the sentence two lines down is what #137 inverts, and it's being left untouched.

> When `doc` (the open doc) is passed, OT trusts its in-memory pending as the source of truth and only reads the store for foreign-tab mint deltas — no state materialization.

#137 makes store rows authoritative and merges the mirror in as the supplement, so after both land the published `ClientAlgorithm` contract states a rule neither implementation follows. This is the interface custom-algorithm and custom-store implementors build against.

Cheap to fix here since the lines above are already being rewritten — otherwise both PRs will each assume the other picked it up.

src/algorithms/ot/shared/ejectPendingChange.ts
```diff
@@ -107,24 +108,23 @@ export function computePendingEjection(
...
     );
   }
   let rev = committedRev;
-  const newPending = [...before, ...rebasedAfter].map(change => ({
-    ...change,
-    baseRev: committedRev,
-    rev: ++rev,
-  }));
+  const newPending = [
+    ...before.map(change => ({ ...change, rev: ++rev })),
```
matalina: Nit: the comment above ("Predecessors already sit on committedRev with these very revs, so this is a no-op for them") was the justification for the unconditional renumber, and it stops being true for exactly the rows this PR adds — a preserved stale predecessor sits on `baseRev = M ≠ committedRev`, so it gets `rev = committedRev + k` with an arbitrary rev/baseRev gap.

Nothing breaks today (`getPendingToSend` slices on baseRev, the server reassigns rev), but this is the one function where frame bookkeeping is the entire point, so a comment that misdescribes the new case is worth a line.

src/client/OTAlgorithm.ts
jacwright: Confirmed and fixed in b87ef19 — you traced it exactly, including the tail: the throw lands after the store transaction, and the reload's hydration filter then drops the straggler into `droppedPendingChanges`, so the row we refused to mislabel gets discarded locally instead.

I took neither of your two shapes, because both leak into the wire on this branch: `getPendingToSend` trusts the open doc's in-memory pending, so a transformed copy in the doc's queue is the copy that gets flushed — and it would be transformed across only part of the gap (the client no longer holds the span the straggler missed), then sent at either the true baseRev (double transform) or the new one (the relabel this PR removes). Your second shape needs a field on `Change`, which is a protocol change I don't want to make here.

What landed instead: the doc's queue stays byte-identical to the store's, and only the derived *state* skips the row. New `applyPendingForView(committedState, committedRev, pending)` behind the three seams that materialize committed+pending (`OTDoc` ctor, `_recomputeState`, `createStateFromSnapshot` for `import`). It strict-applies first, so a frame-consistent queue — and a deferred row that happens to still apply — is bit-for-bit unchanged; only when that throws are rows with `baseRev < committedRev` left out, and a failure that survives their removal still rethrows. The ctor's per-change drop loop also now passes frame-debt rows through to `valid` instead of `droppedPendingChanges`.

That answers the contract you said was unstated, so I wrote it down: **a row deferred to a later frame is not represented in optimistic state; it appears when the server commits it.** The user's edit is invisible for one flush cycle rather than lost, the queue never diverges from the store, and nothing changes on the wire.

src/client/OTAlgorithm.ts
jacwright: Verified, and the first half is right: `staleIds` can cover the whole queue, `rebaseChanges` gets `[]`, and every row is re-emitted with ops and baseRev untouched. That part is the design working — none of those rows is in the frame the walk crosses, so there is nothing to transform them against.

Where I think it lands differently from your read is progress. I wrote the degenerate case as a test (`a wholly-stale queue still drains`, `OTAlgorithm-frameDebt.spec.ts`): a queue at baseRev 1 and baseRev 2 under committedRev 3, doc open, driven through the real server. It drains in exactly two passes — one per distinct frame — the committed log comes out contiguous, and the client converges on the server head. The flush seam slices the leading same-baseRev run and sends it at its true baseRev regardless of how far behind it is, so "no rows were rebased" does not mean "no rows can leave".

The compounding you describe is real but it is *drift*, not debt that blocks: the gap between a row's baseRev and committedRev grows, and the cost of that is server-side — `commitChanges` does `listChanges(startAfter: baseRev)`, so a deeper drift means a wider transform window. Worth naming as a real trade (the old relabel made that read smaller by being wrong), but each pass pays a frame down and the follow-up now chains passes back-to-back, so it only accumulates while the client cannot send at all — at which point the fresh rows are equally stale.

So I have not added a drift cap. It is new policy that discards user work on a threshold, and the honest give-up path already exists and is server-authoritative: baseRev ahead of head, or a baseRev-0 continuation onto a doc this upload did not create, both answer 409 `scope:'doc'`, which `flushDoc` routes to a reload that rebases the whole queue onto the true head. Happy to revisit if you would rather bound it client-side, but I would want it as its own change with a number behind it.

The blast-radius half of your point I did fix — via your `:601` note. The view now omits every frame-debt row, so a wholly-stale queue renders as committed-only instead of throwing on the first row.

src/algorithms/ot/shared/ejectPendingChange.ts
jacwright: Fixed in b87ef19. You are right that the honesty fix stopped at `before`, and the `staleBaseRevs` warn now runs over `[...before, ...after]`.

One correction on the suggested diff: `rebasedAfter` rows have already been through `rebaseChanges`, which stamps `baseRev = serverChanges[last].rev` on every survivor — here that is the inverse carrier's rev, i.e. `poison.rev`. So `change.baseRev === committedRev` tests the carrier's stamp, not the row's original frame. I reassemble from the pre-rebase `after` rows instead and carry each survivor's own `baseRev` across:

```ts
const rebased = rebasedById.get(change.id);
if (rebased) newPending.push({ ...rebased, baseRev: change.baseRev, rev: ++rev });
```

On the invariant queue that is identical to stamping `committedRev` on each, so the normal path is unchanged.

I also went one step past the label. Walking a stale successor through the inverse is wrong for the same reason relabeling it is: a row from a lagging context never had the poison in frame, so the inverse — computed in the poison's frame — must not advance through it. The walk is now restricted to `after.filter(c => c.baseRev === poison.baseRev)`; rows on another frame ride through untouched. That is the same rule `_rebasePendingPreservingFrameDebt` applies on the receive path, and it is a sound discriminator: a row minted blind to the poison cannot have been stacked on it.

Test pins both halves in one queue (`computePendingEjection frame debt`): a straggler two frames back keeps `replace /items/1` and `baseRev 3`, while a same-frame successor still shifts to `/items/2` under the inverse.

Adjacent and pre-existing, not touched: `preState = applyChanges(committedState, before)` strict-applies *all* predecessors, so a stale predecessor whose ops miss the committed frame throws the whole ejection. That is the documented throw-and-latch posture and it is the safe direction, but stale rows now persist in the store rather than being healed on the next receive, so ejection will meet that shape more often than it used to.

src/net/PatchesSync.ts
jacwright: Fixed in b87ef19 — the one-shot property was genuinely lost and nothing replaced it.

I used your progress check with one change: no extra store read. `pending` is already re-fetched at the bottom of the batch loop, so the head comparison is free — which also keeps it out of the way of the F×B read amplification you flag on `:568`.

```ts
const headBefore = pending[0].id;            // before the batch loop
// ...
const drained = pending.length > 0 && pending[0].id !== headBefore;
if ((reloadedMidFlush || drained) && stillHasPending) void this.syncDoc(docId);
```

Compared by id rather than `id@rev`, deliberately: the receive path re-sequences revs on every echo, so a row that did not drain comes back with a new rev and `id@rev` would read that as progress and keep spinning. The queue only ever drains from the front, so the head id moving is the cheapest sound proof that at least one row left. The early-`break` paths keep their old behavior through the `reloadedMidFlush` disjunct.

New `tests/net/PatchesSyncFollowUpFlush.spec.ts` covers both directions: a store whose pending replace silently no-ops (your third example) commits exactly once and then parks with `hasPending: true` instead of cycling; and a three-row two-frame queue still drains in two passes driven entirely by the follow-up. I checked the spin test fails without the guard rather than trusting it — with the check reverted to `pending.length > 0` it runs away and trips the backstop.

tests/client/OTAlgorithm-frameDebt.spec.ts
jacwright: Right on both counts — `doc: undefined` everywhere and `getDoc` materializing committed rows only, so the store assertions could never see it. Added open-doc coverage in b87ef19, and I wrote it before the fix so it earns its keep: three of the four new tests failed on the old code, the middle one throwing `ApplyChangesError` out of `_recomputeState` exactly where you said it would.

Your snippet turned out to under-report the damage. With a real doc the straggler never even reaches `applyServerChanges` — `createDoc(loadDoc())` strict-applies the queue in the ctor, `add /items/3` misses the 2-element frame, and the row is already in `droppedPendingChanges` before any server change arrives. So the loss is one doc open away, not one echo away.

The four:

- **hydration** keeps the deferred row queued (`droppedPendingChanges` empty, both ids present) and renders `{items: ['c','W']}`
- **applyServerChanges** no longer throws; `doc.state` is `{items: ['W']}` and `doc.getPendingChanges()` deep-equals the store queue — that equality is the load-bearing one, since `getPendingToSend` trusts the doc and any divergence would flush the doc's copy instead of the honest row
- **end-to-end with the doc open**, two flush passes, asserting the 'M' comes back and `doc.state` equals the server head
- **wholly-stale queue** drains with the doc open (also covers your `:589` note)

To answer the question in your comment directly — `expect(doc.state).toEqual(/* what should the user see here? */)` — the answer is `{items: ['W']}`: the deferred row's content is **not** in the optimistic view, and appears when its commit echoes back. That is now the stated local contract, written up on `applyPendingForView`. The wire keeps the row, the view does not, and the two stay in sync on identity even while they differ on state.

src/client/OTAlgorithm.ts
jacwright: Acknowledged, and agreed on the pairing — #137 surfaces more stragglers, this PR stops them being poison.

Two things I could do cheaply, so I did:

- the follow-up progress check (your `:1184` note) adds **zero** store reads. Your suggested version called `getPendingToSend` again; I reused the `pending` the batch loop already re-fetched, so the F×B count is unchanged by that fix rather than made worse by it.
- the per-frame follow-up pass is not an extra read *per frame* on top of the existing drain — each pass is a `flushDoc` that would have happened on the next mint or reconnect anyway. What is new is that they now happen back-to-back instead of spread out, so the same reads land in a burst on the reconnect path. On DAB-766 devices that burst is the thing to watch.

One integration cost I do not think either PR has priced, worth adding to the release notes: an honest baseRev makes the *server* transform window wider. `commitChanges` does `listChanges(startAfter: baseRev)`, so a straggler flushing at its true baseRev makes the server read every change since that frame — where the old relabel made it read from the tip. Correct, but it is a real read-amplification trade on the server side, and #137 widening the mixed-baseRev case to the ordinary multi-tab shape means it will be hit far more than the torn-reload rarity here.

On merge order and the combined-tree run: I have not set that in the description, since the order is a call across all three PRs rather than mine to declare unilaterally — flagging to @jacobwright to sequence #136/#137/#145 and to run `OTAlgorithm-frameDebt.spec.ts` plus `PatchesSyncQueueOrdering.spec.ts` on the merged tree before release. Noting that `PatchesSyncQueueOrdering.spec.ts` does not exist on this branch — it arrives with #137 — so that half of the check can only run after both land.

src/client/ClientAlgorithm.ts
jacwright: Checked the branch state before touching it, and this one is #137's — deliberately leaving it alone rather than duplicating.

Where things stand: the sentence is unchanged on `main` and on this branch, and #137's pushed HEAD does not touch `ClientAlgorithm.ts` either. But #137 has the rewrite in its working tree right now, uncommitted — the replacement reads "The store's pending rows are the source of truth — it is the sole rev sequencer, so any context sharing it can mint at a rev the open doc's mirror already holds. When `doc` is passed, OT merges its in-memory pending in as a supplement (by change id, above the store tail)…", which is exactly the inversion you are asking for. So it is picked up, just not visible from any pushed ref yet.

Editing the same paragraph here would put two different rewrites of one paragraph on a collision course — my PR already rewrites the bullet immediately above it, so we would be touching adjacent lines in the same hunk with no textual signal that we disagreed. Your concern that "both PRs will each assume the other picked it up" is the right one, so to make it concrete rather than assumed: **#137 owns the `getPendingToSend` contract paragraph, #145 owns the `- OT:` bullet above it.** If #137 lands without it, that is a one-line follow-up on my side and I will take it.

src/algorithms/ot/shared/ejectPendingChange.ts
jacwright: Fixed in b87ef19. The comment is rewritten — the no-op justification is gone, and it now says every survivor keeps its own baseRev, which happens to *be* `committedRev` on the invariant queue and is why the normal path is unchanged.

Verified nothing depends on the gap, and wrote the result into the comment so the next reader does not have to re-derive it:

- **rev** only has to stay strictly increasing across the queue, and it does — every survivor takes the next `++rev` off `committedRev`, in queue order, on both sides of the poison
- **`getPendingToSend`** uses the two independently: it slices the flush on `baseRev` and ranges the store read on `rev` (`startAfterRev: tail`). Neither reads the pair as a frame. Same for `_collectPending`'s `rev > latestRev` filter, `replacePendingChanges`' tail, and the R2 conflict check, which all take a max-rev watermark
- **the server** reassigns `rev` on commit (`rev = baseRev + 1` over the batch), so the gap never leaves the client. It validates baseRev consistency *within* a batch and rejects baseRev ahead of head — neither is a rev/baseRev relation
- pinned with a test asserting revs stay strictly increasing above `committedRev` across a mixed-frame ejection, so a future change that starts inferring a frame from the pair trips on it

The gap does get wider with this change than when you flagged it, since successors can now carry it too.

src/client/OTAlgorithm.ts
matalina: Sequencing: **#145 → #137 → #136**, then both specs on the merged tree before cutting.

- **#145 first.** #137 alone makes production worse: it widens the batch to include foreign-context rows, pushing more mixed-baseRev changes through a relabel that on main still does not transform. Landing this PR first means the relabel is already a real transform when the batch widens, so there is no window where main is worse than today.
- **#137 second** — the multi-tab reorder generator, safe once the relabel is honest.
- **#136 last** — it heals the read side for poison already committed, and it should not be healing against still-active generators.

The combined-tree check can only happen after #137 lands, since `PatchesSyncQueueOrdering.spec.ts` arrives with it. So: merge #145, merge #137, run `OTAlgorithm-frameDebt.spec.ts` and `PatchesSyncQueueOrdering.spec.ts` together, then #136, then cut.

On the server read amplification — agreed it belongs in the release notes, but I would go past a note. `commitChanges` doing `listChanges(startAfter: baseRev)` against an honest baseRev widens pup transform windows, and pup documents read budgets and heap ceilings for `_changes` reads precisely because unbounded ones have hurt it before. #137 turning mixed-baseRev from a torn-reload rarity into the ordinary multi-tab case means this lands on the reconnect path at population scale. Worth a heads-up to whoever watches pup capacity, plus a measurement of the window width before we assume it stays bounded.

src/client/OTAlgorithm.ts
matalina: Agreed — no client-side cap. A threshold that discards user work is new policy, the 409 `scope:doc` reload is server-authoritative and already exists, and the wholly-stale-queue test draining in two passes answers the part I was actually worried about.

What I would still track is the cost you named yourself: deeper drift means a wider `listChanges(startAfter: baseRev)` on every straggler commit. Fine at torn-reload frequency; less obviously fine once #137 makes mixed-baseRev the ordinary multi-tab shape.

So rather than a cap, a measurement — drift depth (`committedRev - baseRev` at flush) as a metric. Cheap to emit, and it turns *each pass pays a frame down* from an argument into an observation. If it turns out to sit at one or two frames in practice, the question closes itself.

Not blocking.

src/algorithms/ot/shared/ejectPendingChange.ts
matalina: On the adjacent item you flagged at the end — `preState = applyChanges(committedState, before)` strict-applying every predecessor, so a stale-frame predecessor throws the whole ejection:

That wants a ticket rather than a closing paragraph. *Pre-existing but now more frequent* is the category that reliably gets lost, and you have identified the mechanism precisely: stale rows now persist in the store instead of being healed on the next receive, so ejection meets that shape more often than it used to.

It also sits on the quarantine path, which is already the last resort — a throw there latches with nothing behind it to recover.

Happy for it to stay out of scope here. Just not as a comment.

src/client/OTAlgorithm.ts
matalina: Concrete instance of why the combined-tree run matters — **#137 and #145 silently cancel each other in `flushDoc`**, and I do not think git will flag it.

The two branches disagree about whether the per-batch read exists:

- **#137 (`d442c65`)** deletes `pending = await algorithm.getPendingToSend(...)` at the bottom of the batch loop, having established it was vestigial — `batches` is split once before the loop, so it never fed the next batch.
- **#145 (`b87ef19`)** builds the follow-up progress check on exactly that read: `headBefore = pending[0].id` before the loop (`:1066`), then `const drained = pending.length > 0 && pending[0].id !== headBefore` after it (`:1196`).

On the merged tree the re-fetch is gone, so `pending` at `:1196` is still the flush-start array from `:1052` — the same array `headBefore` was taken from. `pending[0].id === headBefore` always, `drained` is always `false`, and the follow-up fires only on `reloadedMidFlush`.

Net effect: **#145 seam 3 is silently reverted.** A mixed-baseRev queue flushes its leading run and the deferred run then sits unsynced until the next mint or reconnect — precisely the gap the follow-up pass was added to close. No error, no test failure on either branch alone.

And the hunks are adjacent rather than overlapping (#137 deletes around `:1179-1180`, #145 adds at `:1066` and `:1189-1197`), so this most likely auto-merges clean with no conflict marker to prompt a second look.

The good news: `PatchesSyncFollowUpFlush.spec.ts` should catch it — its two-frame drain test depends on the follow-up firing. But only when run on the merged tree, which is the check that cannot happen until both are in.

A merged form that keeps both wins — #137 collapses K per-batch reads to one, #145 keeps a sound progress signal:

```ts
// after the batch loop, one read instead of one per batch
const remaining = (await algorithm.getPendingToSend(docId, this.patches.getOpenDoc(docId) as PatchesDoc<any>)) ?? [];
const drained = remaining.length > 0 && remaining[0].id !== headBefore;
if ((reloadedMidFlush || drained) && stillHasPending) void this.syncDoc(docId);
```

Whoever merges second should expect to write this by hand rather than trust the auto-merge.

src/algorithms/ot/shared/ejectPendingChange.ts
jacwright: Filed as DAB-1028 (S, High) with the mechanism and the DAB-951 frequency note. Agreed it belongs on a ticket — the quarantine-path latch is the part that makes it more than cleanup.

src/client/OTAlgorithm.ts
jacwright: Reconciled on #137's branch, in the shape you suggested: one post-loop read instead of one per batch (`b230f0d`).

```ts
// after the batch loop
pending = (await algorithm.getPendingToSend(docId, this.patches.getOpenDoc(docId) as PatchesDoc<any>)) ?? [];
```

Assigning back into `pending` rather than a new `remaining` local is deliberate: your `drained` line then reads the fresh queue with no edit of its own, so the two hunks stay non-overlapping and the auto-merge is correct instead of merely clean. #137 keeps its K-to-1 collapse, #145 keeps a sound progress signal, and the merged form is one read per flush rather than K.

Built the combined tree locally to check rather than reason about it (#137 merged onto #145, not pushed anywhere):

- `PatchesSyncFollowUpFlush.spec.ts` 2 passed, `OTAlgorithm-frameDebt.spec.ts` 9 passed, `PatchesSyncQueueOrdering.spec.ts` 5 passed
- full suite on the combined tree: 136 files, 3535 passed, 4 skipped; type, lint, format clean
- on #137's branch alone: 134 files, 3519 passed, 4 skipped

Your prediction that `PatchesSyncFollowUpFlush.spec.ts` catches the cancellation is right, and I confirmed it the only way that means anything: with the post-loop read deleted from the combined tree, `still re-arms while the queue is actually draining, one frame per pass` fails (`expected true to be false` on `hasPending`), while both branches stay green on their own. So the spec is a real guard on the merged tree, and only there.

One thing the merge does flag loudly, worth knowing before it happens: `src/client/OTAlgorithm.ts` conflicts in `_withConsistentBaseRev` (docstring and warn text), because #137 amended the paragraph #145 replaces wholesale. Take #145's side for both hunks; that is what I ran the combined suite against. `ClientAlgorithm.ts`, `PatchesSync.ts` and `OTAlgorithm-getPendingToSend.spec.ts` auto-merge clean.

jacwright: Sequencing acknowledged and recorded: **#145 -> #137 -> #136**, with `OTAlgorithm-frameDebt.spec.ts` and `PatchesSyncQueueOrdering.spec.ts` run together on the merged tree before cutting. It is now a "Merge order" section in all three PR descriptions, with your reasoning for each position, rather than living only in this thread.

The combined-tree run is already done once locally, ahead of the release: full suite 3535 passed / 4 skipped, both named specs green, and the `flushDoc` cancellation you found in the other thread fixed on #137's side so the auto-merge is correct and not just clean.

On the server read amplification: agreed it is more than a release note. Raising it with whoever watches pup capacity, with the drift-depth measurement from your `:589` note as the concrete ask, since `committedRev - baseRev` at flush is exactly what sets the `listChanges(startAfter: baseRev)` window width. Tracked outside these three PRs so it does not gate the release.

matalina: Correcting my own sequencing note above — I over-specified it.

All three go out in a single `@dabble/patches` publish, so no intermediate state of `main` ever reaches users. That makes the only hard rule a **release**-level one: do not publish a version containing #137 without #145. A single release satisfies it by construction.

The merge order I gave is tidiness, not a constraint — and one line of it was simply wrong. *#136 last, so it is not healing against still-active generators* is a release argument, not a merge argument, and it is moot when nothing publishes in between.

Practically: **#145 and #136 should merge now rather than wait.** #145 especially — it closes a generator that was still producing poison on 8/7 and 8/8, it is safe on its own, and landing it first puts the `flushDoc` reconciliation on #137 rebase, where you will already be working on the contract change. Holding it just delays a live fix and moves the same conflict work onto whoever merges second.

#137 is the only one actually gated.

matalina (approved):

PR #146, opened 2026-08-14
test(json-patch): pin that compose never folds a descendant into an ancestor (DAB-1040)

matalina:
Test-only — one case added to `tests/json-patch/compose.spec.ts`.

## Why

pup's comments-map-root guard (dabblewriter/pup#235) admits a write at a `comments` map root only when its value is `{}` — a populated one is indistinguishable from wiping every thread in the doc. dw3 mirrors that check client-side.

A client initialising an absent map commits the `{}` root **and** the first thread in ONE change. Both sides evaluate ops individually, so that pair passes only while composition leaves it as two ops. Were a descendant ever folded into its ancestor, the pair would arrive as a single *populated* root write, both sides would refuse it, and the doc would latch — the DAB-970 failure mode.

`composePatch` already does the right thing: it composes only identical paths and deletes ancestor/descendant entries rather than merging them. The existing `stops composing with higher paths` case pins the descendant-then-ancestor direction; nothing pinned the ancestor-then-descendant pair the guard actually rests on.

Pinning it here rather than in the two consumers means a future change to `composePatch` fails at the source, not downstream in pup and dw3.

## Testing

`tests/json-patch/compose.spec.ts` — 12 pass (11 existing + 1 new). No source changes.

### DIFF
--- a/tests/json-patch/compose.spec.ts
+++ b/tests/json-patch/compose.spec.ts
@@ -155,4 +155,28 @@ describe('composePatch', () => {
       { op: '@inc', path: '/x/y', value: 2 },
     ]);
   });
+
+  // Consumers depend on this one by name, so it is pinned separately from the
+  // chain-breaking case above: an ancestor write and a later descendant write
+  // must survive as TWO ops, with the ancestor's value untouched.
+  //
+  // pup's role guard admits a write at a `comments` map root only when its value
+  // is `{}` (a populated one is indistinguishable from wiping every thread in the
+  // doc), and dw3 mirrors that check client-side. A client initialising an absent
+  // map commits the `{}` root AND the first thread in ONE change. Both sides
+  // evaluate ops individually, so this pair passes only while it stays two ops.
+  // Were a descendant ever folded into its ancestor here, that pair would arrive
+  // as a single POPULATED root write, both sides would refuse it, and the doc
+  // would latch — so this is a cross-repo invariant, not a local detail.
+  it('never folds a descendant write into an ancestor write', () => {
+    expect(
+      composePatch([
+        { op: 'replace', path: '/docs/d1/body/comments', value: {} },
+        { op: 'replace', path: '/docs/d1/body/comments/tc1', value: { id: 'tc1', change: true } },
+      ])
+    ).toEqual([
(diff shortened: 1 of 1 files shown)


jacwright (approved):
Approving. This is the right response to the note I left on pup#235 — that guard leans on a property of `composePatch` that nothing stated, and a test is a better answer than the docblock line I suggested, because it fails when the property goes rather than when someone happens to read the comment.

**Control run:** I made `composePatch` actually fold a descendant into its ancestor (merging the child's value into the parent op and dropping the child) — the exact hazard — and it fails **`never folds a descendant write into an ancestor write`**, plus `stops composing with higher paths`. As shipped, 12/12 in the file and **3515 passed / 4 skipped across 133 files** repo-wide. CI green.

The two failing tests are worth distinguishing, because it explains why this one earns its place next to the existing coverage. `stops composing with higher paths` pins the *mechanism* — that ancestors and descendants break each other's chains. This pins the *consequence* consumers actually depend on: two ops in, two ops out, ancestor value untouched. A refactor that preserved chain-breaking while introducing folding elsewhere would slip past the first and be caught by the second.

Choosing the real `/docs/d1/body/comments` + `/comments/tc1` pair rather than an abstract `/a` + `/a/b` is what makes the comment land. Someone changing compose sees the actual shape that breaks, in the repo where the change would be made, rather than a cross-repo invariant they'd have to go looking for.

The comment is the valuable half and it is accurate on both sides: pup admits a `comments` map-root write only when the value is `{}` (`documents/roles.ts`, `isCommentTreeOp` → `isEmptyObject`), and dw3's client mints the `{}` bootstrap and the first thread in one change from `upsertChangeThread`. I verified that pair reaching pup as two ops when I reviewed pup#235; this now keeps it that way.

One thing worth knowing, since the comment says "consumers depend on this one by name": patches has no way to enforce that beyond this test, because the consumers are in other repos. That is fine — it is the same trade as any published-library invariant — but it does mean the comment is the only pointer back to why. Naming DAB-1040 in it, as you have, is what makes that recoverable.

Nothing else to raise. Test-only, no production change, and it closes a real gap in a guard that is about to ship.

PR #147, opened 2026-08-18
fix(net): authorize branch merge/update/delete against the source doc

matalina:
## Security fix

The JSON-RPC/WS auth wrapper authorizes `args[0]`. That's correct for `commitChanges(docId, …)`, but branch `merge`/`update`/`delete` take the **branch** id as `args[0]` while they write into (merge) or govern (update/delete) the **source** document. So the generic auth checked the *branch's* permissions, not the source's.

**Impact:** a principal who is `owner`/`write` on a review-copy branch but lacks write on the source (revoked, `comment`, `view`) could merge arbitrary content into the source manuscript — or update/delete the branch against a source they can't write — over the RPC/WS transport. The REST transport already gets this right (it checks write on the source doc named in the request URL); this brings RPC/WS in line.

## Fix

Kept within the library's scope boundary — Patches decides *which* document the consumer's `AuthorizationProvider` is asked about (a fact about the operation: merge writes to the source), never *whether* a user may act:

- **`ApiMethodDefinition`** gains an optional `authDoc(args, target)` resolver. When present, `register()` authorizes the document it returns instead of `args[0]`. `assertAccess`'s signature and the wire-docId validation are unchanged — non-`authDoc` methods and any subclassers are unaffected.
- **`BranchManager.getSourceDocId(branchId)`** (implemented on OT + LWW) resolves `Branch.docId` via `loadBranch`, throwing `StatusError(404)` on a missing/tombstoned branch so the access check **fails closed**. It's a `StatusError` (not a bare `Error`) because the resolver runs *before* authorization, so the throw reaches an as-yet-unauthorized caller — a bare `Error` would return `-32000` with a server stack trace.
- **`branchManagerApi`** declares merge/update/delete with `authDoc: (args, mgr) => mgr.getSourceDocId(args[0])`; create/list keep `args[0]` (already the source).

## Tests

- `JSONRPCServer.spec.ts`: authorizes the resolved **source**, not the branch; **denies a caller who can access the branch but not its source** (the review-copy escalation); **fails closed** with a `404` when the branch is unknown (provider not even consulted); non-`authDoc` methods still authorize `args[0]`.
- `branchUtils` + OT/LWW manager specs: api shape and `getSourceDocId` (returns source; throws `StatusError(404)` on missing/tombstone).

Full suite green on a clean `npm ci` (**3539 passed / 4 skipped**, `tests/solid` included); type-check / lint / format clean; CI `check` green.

## Release

- Version **bumped to `0.28.0`** in this PR (behavior change is minor under 0.x), so it can be published as soon as this merges.
- **Follow-up after publish, in order:** publish `0.28.0` → bump `@dabble/patches` to `0.28.0` in **pup** and **deploy pup** (this is where the exploit actually closes) → then bump the other consumers (**dabble-writer-3.0**, **dabble-admin**, and the caret consumers `dabble-rest` / `dabble-migrations` / `dabble-functions`, which is version alignment, not a security fix — none of them run the exposed RPC branch surface).

### DIFF
--- a/src/net/protocol/JSONRPCServer.ts
+++ b/src/net/protocol/JSONRPCServer.ts
@@ -13,8 +13,24 @@ export type MessageHandler<R = any> = (...args: any[]) => Promise<R> | R;
  * When `params` is declared, `register()` zips the names with the call arguments and
  * passes the resulting object to `AuthorizationProvider.canAccess` — required for
  * providers that validate request payloads (e.g. role-scoped commit rules on `changes`).
+ *
+ * `authDoc` names the document whose access governs the call when it is NOT the first
+ * argument. The access check defaults to authorizing `args[0]`, which is correct for
+ * `commitChanges(docId, …)` and friends; but branch `merge`/`update`/`delete` take the
+ * *branch* id as `args[0]` while they write into (merge) or govern (update/delete) the
+ * *source* document, so authorizing the branch would let anyone with branch access act on
+ * the source. When `authDoc` is present, `register()` resolves it (with the call args and
+ * the registered instance) and authorizes the returned docId instead. It may be async — a
+ * branch resolver reads the branch record to find its source — and a throw fails the access
+ * check closed.
  */
-export type ApiMethodDefinition = Access | { access: Access; params?: readonly string[] };
+export type ApiMethodDefinition =
+  | Access
+  | {
+      access: Access;
+      params?: readonly string[];
+      authDoc?: (args: readonly any[], target: any) => string | Promise<string>;
+    };
 
 /** Static API definition mapping method names to access levels */
 export type ApiDefinition = Record<string, ApiMethodDefinition>;
@@ -116,6 +132,7 @@ export class JSONRPCServer {
       }
       const access = typeof definition === 'string' ? definition : definition.access;
       const paramNames = typeof definition === 'string' ? undefined : definition.params;
+      const authDoc = typeof definition === 'string' ? undefined : definition.authDoc;
 
       this.registerMethod(method, async (...args: any[]) => {
         const docId = args[0];
@@ -130,7 +147,14 @@ export class JSONRPCServer {
         // building the named-params object (allocated on every call otherwise) in
         // that case — it would just be discarded unused.
         const params = this.auth ? buildNamedParams(paramNames, args) : undefined;
-        await this.assertAccess(access, ctx, method, args, params);
+        // The document whose access governs this call is args[0] by default, but a
+        // method may resolve it from the args instead (branch merge/update/delete take
+        // the branch id as arg 0 yet are governed by the source doc). Only the docId in
+        // slot 0 is read by assertAccess, so hand it an args list with the resolved id
+        // there — the method itself still runs on the real, unmodified args. Skip the
+        // (possibly async) resolve when there's no auth provider to consult.
+        const authArgs = this.auth && authDoc ? [await authDoc(args, obj), ...args.slice(1)] : args;
+        await this.assertAccess(access, ctx, method, authArgs, params);
         // _dispatch cleared the context during the await above; re-establish it
         // around the method's synchronous start so getClientId() works inside.
         setAuthContext(ctx);
--- a/src/net/protocol/JSONRPCServer.ts
+++ b/src/net/protocol/JSONRPCServer.ts
@@ -23,6 +23,16 @@ export type MessageHandler<R = any> = (...args: any[]) => Promise<R> | R;
  * the registered instance) and authorizes the returned docId instead. It may be async — a
  * branch resolver reads the branch record to find its source — and a throw fails the access
  * check closed.
+ *
+ * `params` and `authDoc` describe different things and never interfere: `params` reflects the
+ * request *payload* (built from the real, unresolved args, so a provider validating e.g.
+ * `changes` sees exactly what the client sent), while `authDoc` only redirects *which docId*
+ * the check governs. A method may in principle declare both.
+ *
+ * `authDoc` runs *before* any authorization check, so its throw reaches an as-yet-unauthorized
+ * caller: keep a resolver to a single cheap record read, and throw a `StatusError` (whose code
+ * and data are returned verbatim) rather than a bare `Error` (which leaks a server stack trace).
+ * Do not make it more expensive or more revealing than looking up the governed doc.
  */
 export type ApiMethodDefinition =
   | Access
--- a/src/server/LWWBranchManager.ts
+++ b/src/server/LWWBranchManager.ts
@@ -137,6 +137,17 @@ export class LWWBranchManager implements BranchManager {
   }
 
   /**
+   * Resolves a branch id to the source document it belongs to (`Branch.docId`), so the RPC
+   * layer can authorize merge/update/delete against the source rather than the branch. Throws
+   * on a missing or tombstoned branch, which fails the access check closed.
+   */
+  async getSourceDocId(branchId: string): Promise<string> {
+    const branch = await this.store.loadBranch(branchId);
+    assertBranchExists(branch, branchId);
+    return branch.docId;
+  }
+
+  /**
    * Merges a branch back into its source document.
    *
    * Supports multiple merges — the branch stays open and `lastMergedRev` tracks
--- a/src/server/LWWBranchManager.ts
+++ b/src/server/LWWBranchManager.ts
@@ -1,4 +1,5 @@
 import { JSONPatch } from '../json-patch/JSONPatch.js';
+import { StatusError } from '../net/error.js';
 import type { Branch, Change, CreateBranchMetadata, EditableBranchMetadata, ListBranchesOptions } from '../types.js';
 import type { BranchManager } from './BranchManager.js';
 import {
@@ -139,11 +140,14 @@ export class LWWBranchManager implements BranchManager {
   /**
    * Resolves a branch id to the source document it belongs to (`Branch.docId`), so the RPC
    * layer can authorize merge/update/delete against the source rather than the branch. Throws
-   * on a missing or tombstoned branch, which fails the access check closed.
+   * a `StatusError(404)` on a missing or tombstoned branch, which fails the access check
+   * closed. It is a `StatusError` — not `assertBranchExists`'s plain `Error` — because this
+   * runs *before* authorization, so the thrown value is surfaced verbatim to an unauthorized
+   * caller; a plain `Error` would return `-32000` carrying a server stack trace.
    */
   async getSourceDocId(branchId: string): Promise<string> {
     const branch = await this.store.loadBranch(branchId);
-    assertBranchExists(branch, branchId);
+    if (!branch || branch.deleted) throw new StatusError(404, `Branch ${branchId} not found`);
     return branch.docId;
   }
 
--- a/src/server/OTBranchManager.ts
+++ b/src/server/OTBranchManager.ts
@@ -1,4 +1,5 @@
 import type { Op } from '@dabble/delta';
+import { StatusError } from '../net/error.js';
 import { findLatestMainVersion } from '../algorithms/ot/server/getSnapshotAtRevision.js';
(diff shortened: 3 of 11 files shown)


src/net/protocol/JSONRPCServer.ts
```diff
@@ -130,7 +147,14 @@ export class JSONRPCServer {
...
         // that case — it would just be discarded unused.
         const params = this.auth ? buildNamedParams(paramNames, args) : undefined;
-        await this.assertAccess(access, ctx, method, args, params);
+        // The document whose access governs this call is args[0] by default, but a
+        // method may resolve it from the args instead (branch merge/update/delete take
+        // the branch id as arg 0 yet are governed by the source doc). Only the docId in
+        // slot 0 is read by assertAccess, so hand it an args list with the resolved id
+        // there — the method itself still runs on the real, unmodified args. Skip the
+        // (possibly async) resolve when there's no auth provider to consult.
+        const authArgs = this.auth && authDoc ? [await authDoc(args, obj), ...args.slice(1)] : args;
```
jacwright, line 166: The placement is right — the resolve has to precede `assertAccess`, since its whole job is to decide what gets authorized — but it does introduce a store read that happens **before any authorization check has run**, which is new.

Concretely: an authenticated caller with no rights on either the branch or the source can call `mergeBranch('projects/<guess>/content')` and reach `loadBranch` on an arbitrary id. `assertBranchExists` then distinguishes the cases in its message — `Branch with ID x not found.` versus `Branch x has been deleted.` versus proceeding to a `WRITE_FORBIDDEN:<source>` — so the error text is a branch-existence oracle for any signed-in user.

Low severity on its own, and closing it properly is awkward (you'd need to authorize the branch first, then the source, which doubles the provider round-trips for the common case). Not asking for that here. But it's worth a line in the `authDoc` docblock noting that the resolver runs pre-authorization, so nobody later writes one that does something more expensive or more revealing than a single record read.

The `this.auth && authDoc` short-circuit is the right guard, incidentally — with no provider configured there's nothing to authorize, so skipping a possibly-async store read is both correct and free.

One forward-looking note: `buildNamedParams(paramNames, args)` deliberately uses the real `args`, while `authArgs` carries the resolved id. That's the right split — `params` describes the payload, `args[0]` names the governed doc — but no method currently declares both `params` and `authDoc`, so the interaction is untested. A sentence in the type docblock saying which one `params` reflects would save the first person who combines them from having to work it out.

src/server/OTBranchManager.ts
```diff
@@ -596,6 +596,17 @@ export class OTBranchManager implements BranchManager {
...
   }
 
+  /**
+   * Resolves a branch id to the source document it belongs to (`Branch.docId`), so the RPC
+   * layer can authorize merge/update/delete against the source rather than the branch. Throws
+   * on a missing or tombstoned branch, which fails the access check closed.
+   */
+  async getSourceDocId(branchId: string): Promise<string> {
+    const branch = await this.store.loadBranch(branchId);
+    assertBranchExists(branch, branchId);
```
jacwright: `assertBranchExists` throws a plain `Error`, and on the new pre-authorization path that has a consequence the PR doesn't mention. (Applies equally to the identical `LWWBranchManager.getSourceDocId`.)

`processMessage`'s catch builds `rpcError(err?.code ?? -32000, err?.message ?? 'Server error', err?.code ? err?.data : err?.stack, …)`. `StatusError` carries a `.code`, so it yields a clean code plus its structured `data`. A plain `Error` has none, so it yields `-32000` **with `err.stack` as the payload** — a server stack trace returned over the wire.

Before this PR that only happened after an access check had already passed. Now `getSourceDocId` is the first thing a branch merge/update/delete touches, so an unauthorized caller poking an unknown branch id gets the stack back.

The fail-closed behaviour itself is correct and pinned — I confirmed dropping the assert fails `throws for a missing branch, so the access check fails closed`. It's only the error *shape* that's off. `throw new StatusError(404, ...)` here would keep the identical control-flow, produce a truthful status instead of a synthetic server error, and stop the stack from being emitted. It would also stop a benign race — someone merging a branch a collaborator just deleted — from reading as a 500 in pup's error reporting.

Existing callers would be unaffected in behaviour; `assertBranchExists` already throws on exactly these two conditions, and every current caller is inside an operation that should surface them as client errors anyway.

jacwright (approved):
Approved. The vulnerability is real and reachable in production, the fix is placed at the right layer, and every part of it is pinned. I verified the reachability and the REST comparison against pup rather than taking the description's word for it.

## The exploit path is live

- pup registers the branch manager through the **generic** path — `this.rpc.register(otBranchManager)` at `PupRPCServer.ts:66`, under the comment "Branch methods (listBranches, createBranch, updateBranch, deleteBranch, mergeBranch)". So it inherits exactly the `args[0]` auth this PR corrects; there is no pup-side override.
- Both transports are mounted: `POST /rpc` (`index.ts:139`) and a `WebSocketServer` over the same `rpcServer.rpc` (`index.ts:330-331`).
- `/rpc` does require an authenticated uid (401 otherwise), so the attacker must be a signed-in Dabble user — which the review-copy scenario already assumes, since they hold `owner`/`write` on a branch.

## The REST comparison holds exactly

`rest.ts` routes branch operations as `/docs/:docId/_branches/:branchId/...` and calls `assertWrite(authData, docId, 'mergeBranch')` on the **source** from the URL before `manager.mergeBranch(branchDocId)` — same for update (`:884`) and delete (`:897`). So RPC really was the odd one out, and this brings it in line rather than inventing a new rule.

The "signature and wire-docId validation unchanged" claim is also exact: `assertAccess` reads `args?.[0]` and nothing else, and `register`'s own `typeof docId !== 'string'` guard still runs against the real, unresolved `args[0]` before anything else happens.

## Controls (7 inversions, anchors asserted unique)

| change | fails |
| --- | --- |
| `updateBranch` reverted to plain `'write'` | 2, incl. `should share the standard branch api (source-authorized merge/update/delete)` |
| `deleteBranch` reverted | the same 2 |
| `mergeBranch` reverted | 3, incl. `authDoc resolves the source via the manager's getSourceDocId(args[0])` |
| **`authDoc` ignored in `register()`** | **3, incl. `denies a caller who can access the branch but not its source`** |
| `getSourceDocId` returns the branch id | `returns the branch record's source docId` |
| fail-closed assert dropped | `throws for a missing branch, so the access check fails closed` + the tombstone twin |

Baseline 172/172 on those four specs. The fourth row is the one that matters — the escalation itself is pinned by a test named after it, so the whole mechanism can't be refactored away silently.

## Correction: the `tests/solid` caveat doesn't reproduce

> The 3 pre-existing `tests/solid/*` failures are an unrelated `@solid-refresh` toolchain issue, present on `main`.

Not on a clean checkout. In a fresh worktree at `acef0aef` with a real `npm ci`, the **full suite is green with zero failures**: 135/135 files, **3539 passed / 4 skipped**. Running `tests/solid` alone is 35/35 across all three specs.

So that's a local `node_modules` artifact rather than a property of `main` — worth striking from the description before it becomes folklore that the solid tests always fail, because the next person who sees a real failure there will wave it through. (The stated 3504 doesn't match my 3539 either, which points the same way.)

## Two notes on the fail-closed path, neither blocking

Both stem from the same choice and are detailed inline: the resolve now runs **before** any authorization check, and `assertBranchExists` throws a plain `Error` rather than a `StatusError`. The combination means an unauthorized caller gets a server stack trace back. That pattern predates this PR, but this PR is what makes it reachable pre-authorization.

## Release

`0.28.0` in the PR is the right call for a 0.x behaviour change, and the follow-up sequencing is right too — **pup is where this actually closes**, since it's the one running the exposed RPC surface. Worth being explicit on the ordering when it lands: publish, bump pup, deploy pup, and only then dw3, since dw3's bump is cosmetic here while pup's is the fix.

`tsc --noEmit`, eslint and prettier clean on a fresh `npm ci`; CI `check` green.

src/server/LWWBranchManager.ts
```diff
@@ -136,6 +137,20 @@ export class LWWBranchManager implements BranchManager {
...
+   * Resolves a branch id to the source document it belongs to (`Branch.docId`), so the RPC
+   * layer can authorize merge/update/delete against the source rather than the branch. Throws
+   * a `StatusError(404)` on a missing or tombstoned branch, which fails the access check
+   * closed. It is a `StatusError` — not `assertBranchExists`'s plain `Error` — because this
+   * runs *before* authorization, so the thrown value is surfaced verbatim to an unauthorized
+   * caller; a plain `Error` would return `-32000` carrying a server stack trace.
+   */
+  async getSourceDocId(branchId: string): Promise<string> {
+    const branch = await this.store.loadBranch(branchId);
+    if (!branch || branch.deleted) throw new StatusError(404, `Branch ${branchId} not found`);
```
jacwright, line 150: This is correct and identical to the OT implementation, but unlike OT it is entirely untested — and the invariant it carries is the one this PR exists to establish.

Two controls, both green:

- reverting this line to `assertBranchExists(branch, branchId)` — the exact thing the docblock above now says not to do — fails nothing;
- replacing the whole method body with `return branchId` — which **reinstates the escalation**, authorizing the branch instead of its source — also fails nothing.

`tests/server/LWWBranchManager.spec.ts` only asserts `LWWBranchManager.api === branchManagerApi`, so it confirms the shared api shape (and therefore that `authDoc` is declared) but never that this manager's resolver returns the source. The OT spec's three `getSourceDocId` cases have no LWW counterpart.

Reachability bounds the severity, which is why this isn't blocking: pup registers only the OT manager (`if (otBranchManager) this.rpc.register(otBranchManager)`), so no Dabble surface exposes the LWW branch methods today. But `LWWBranchManager` is a public export (`@dabble/patches/server`), the shared `branchManagerApi` means any consumer registering it gets `authDoc` pointed at this method, and a security invariant that holds on one of two implementations by inspection only is the kind that drifts.

The fix is to port the three OT cases — returns the source docId, `StatusError(404)` on missing, `StatusError(404)` on tombstoned. The LWW spec already mocks `store.loadBranch`, so it should be close to copy-paste.

jacwright (approved):
Re-reviewed at `5557cf0e`. Both notes closed, and the `StatusError` change does slightly more than it claims. One asymmetry found in the process — detail inline, non-blocking.

**Controls (5 inversions, anchors asserted unique):**

| change | fails |
| --- | --- |
| OT `getSourceDocId` reverted to `assertBranchExists` | both `throws a StatusError(404) …` tests |
| OT: tombstone half of the guard dropped | `throws a StatusError(404) for a tombstoned branch` |
| **LWW `getSourceDocId` reverted to `assertBranchExists`** | **nothing** |
| **LWW `getSourceDocId` broken outright — returns `branchId`, never throws** | **nothing** |

Baseline 172/172 on the four specs; full suite still 135/135 files, 3539 passed / 4 skipped.

The OT side is pinned properly, including each half of the guard separately, and the RPC-level assertion on `response.error.code === 404` is the one that actually proves the wire shape rather than the thrown type. The LWW side has no `getSourceDocId` test at all — that's the asymmetry.

## The uniform message closes the oracle, which is worth naming

`if (!branch || branch.deleted) throw new StatusError(404, ...)` collapses what `assertBranchExists` reported as two distinguishable messages (`Branch with ID x not found.` versus `Branch x has been deleted.`) into one. So the pre-authorization probe I raised last round can no longer tell a tombstoned branch from a nonexistent one — a caller only learns "not a live branch", which is the same thing they'd learn from any 404. That's a real narrowing beyond the stack-trace fix, and it isn't mentioned anywhere; worth a line in the PR body, because the next person to "tidy" this back into `assertBranchExists` for consistency would undo it along with everything else.

## The docblock addition is the right call

Recording that `authDoc` runs pre-authorization, with the explicit "single cheap record read, throw a `StatusError`" constraint, puts the rule where the next resolver author will be standing. The `params` vs `authDoc` paragraph resolves the ambiguity I flagged in the right direction too — `params` from the real args so a provider validating `changes` sees what the client actually sent, `authDoc` governing only which docId is checked.

`tsc --noEmit`, eslint and prettier clean on a fresh `npm ci`; CI `check` green. Still approved.

PR #148, opened 2026-08-18
fix(client): shelve withheld rows on doc discard

jacwright:
TL;DR: When a shared document disappears out from under someone (access revoked, deleted from another device), everything they had typed but not yet synced is now handed to the app to shelve — including the one category that previously fell through: edits that never made it into local durable storage. Nothing a writer can see on screen is silently thrown away anymore.

Follow-up promised in #137 (thread: https://github.com/dabblewriter/patches/pull/137#discussion_r3808108358).

`getPendingToSend` answers "what is safe to put on the wire", and the send path is right to withhold a doc-only row the store never accepted. But `_handleRemoteDocDeleted` was reusing that answer for a different question — "what would vanish with this doc" — and the withheld rows are exactly the ones whose sole remaining copy is the mirror being discarded.

New `ClientAlgorithm.collectUnsyncedForDiscard(docId, doc?)`, a pure read:

- **OT**: `[...pending, ...withheld]` from `_collectPending`. No `_withConsistentBaseRev` (shelf payload, not a wire batch), no `UnstoredPendingError` (the doc is legitimately vanishing), and the once-per-doc report latch stays untouched for docs that survive.
- **LWW**: strictly more complete than the old call — it returns BOTH the in-flight sending change and a change built from ops accumulated behind it (the retry branch returns the in-flight change alone), and unlike `getPendingToSend` it neither mints a sending change nor clears pending ops on a doc that is about to be torn down.
- `_handleRemoteDocDeleted` uses it for the shelf payload. The `{ report }` option `getPendingToSend` grew in #137 has no remaining callers and never shipped in a release, so it is removed rather than left as dead API.
- One-line comment pinning the load-bearing `undefined` in `flushDoc`'s drained compare (from the #137 discussion) so nobody normalises it to `null`.

Tests: the queue-ordering spec's below-tail case now pins inclusion (`['later', 'orphan']`) instead of exclusion; a new LWW spec pins the pure-read contract and the both-halves payload; 3565 passing.


src/client/ClientAlgorithm.ts
```diff
@@ -108,11 +108,23 @@ export interface ClientAlgorithm {
...
    * what is durable — and OT reports it on {@link onError} rather than withholding it silently.
-   * `options.report: false` suppresses that report for callers reading the queue off the send
-   * path (e.g. building a shelf payload for a doc that is being discarded).
    * Returns null if nothing to send.
    */
-  getPendingToSend(docId: string, doc?: PatchesDoc<any>, options?: { report?: boolean }): Promise<Change[] | null>;
+  getPendingToSend(docId: string, doc?: PatchesDoc<any>): Promise<Change[] | null>;
+
+  /**
+   * Everything the doc still holds that the server does not — durable or not — for a doc being
```
RyanHavoc: Two overclaims in this docstring worth tightening.

"Everything the doc still holds … durable or not" — the un-minted optimistic tier is invisible to both implementations. Ops whose mint is still queued live only in `BaseDoc._optimisticOps`, not in the store or the pending mirror, so the collect misses them — and `closeDoc`'s drain then persists them just in time for `confirmDeleteDoc` to wipe them ([Patches.ts#L553-L568](https://github.com/dabblewriter/patches/blob/f67fdff364ca479184e4f834b8204aa50835cf04/src/client/Patches.ts#L553-L568)). For a write-latched doc the miss is deterministic — closeDoc's own comment says the retained ops "go with it". Either scope the claim to the store + pending mirror, or drain before collecting in `_handleRemoteDocDeleted`. (Quarantine is a second durable tier the shelf omits — separate note at the OT impl.)

"a collaborator losing access" — nothing routes revocation here: 403 is terminal and latches the doc at `'error'` ([error.ts#L40](https://github.com/dabblewriter/patches/blob/f67fdff364ca479184e4f834b8204aa50835cf04/src/net/error.ts#L40)), and every entry to `_handleRemoteDocDeleted` is strictly the 410/`docDeleted` class ([PatchesSync.ts#L1964-L1966](https://github.com/dabblewriter/patches/blob/f67fdff364ca479184e4f834b8204aa50835cf04/src/net/PatchesSync.ts#L1964-L1966)). Defensible if a downstream server signals revocation as a delete, but as written it reads like a wired-up trigger.

src/client/OTAlgorithm.ts
```diff
@@ -198,6 +192,20 @@ export class OTAlgorithm implements ClientAlgorithm {
...
+   * already computes: the sendable queue plus the withheld doc-only rows the send path refuses
+   * — refuses because they are not durable, which on a discard is the reason to include them.
+   * Raw rows, no {@link _withConsistentBaseRev}: this is a shelf payload, not a wire batch, and
+   * no report — the doc is legitimately vanishing, so the store-integrity alarm would misfire
+   * (and the once-per-doc latch stays untouched for any doc that survives).
+   */
+  async collectUnsyncedForDiscard(docId: string, doc?: PatchesDoc<any>): Promise<Change[]> {
+    const otDoc = doc as OTDoc<any> | undefined;
+    const { pending, withheld } = await this._collectPending(docId, otDoc, otDoc?.committedRev ?? 0);
+    return [...pending, ...withheld];
```
RyanHavoc: The union misses one durable tier: quarantined changes. They're user content the server refused, exposed by this same interface ([ClientAlgorithm.ts#L261](https://github.com/dabblewriter/patches/blob/f67fdff364ca479184e4f834b8204aa50835cf04/src/client/ClientAlgorithm.ts#L261)) and documented as never auto-discarded ([docs/quarantine.md#L48-L51](https://github.com/dabblewriter/patches/blob/f67fdff364ca479184e4f834b8204aa50835cf04/docs/quarantine.md#L48-L51)) — but they're absent from the shelf, and the discard path then destroys them (`OTInMemoryStore.confirmDeleteDoc` clears quarantine explicitly; on IndexedDB the rows are orphaned). If a doc with an ejected change is deleted remotely, the one mechanism built to keep refused content recoverable is dropped by the "last chance to shelve" payload.

Related store inconsistency while you're in here: `LWWInMemoryStore.confirmDeleteDoc` ([#L164-L166](https://github.com/dabblewriter/patches/blob/f67fdff364ca479184e4f834b8204aa50835cf04/src/client/LWWInMemoryStore.ts#L164-L166)) doesn't clear quarantine, unlike its OT sibling ([OTInMemoryStore.ts#L184-L187](https://github.com/dabblewriter/patches/blob/f67fdff364ca479184e4f834b8204aa50835cf04/src/client/OTInMemoryStore.ts#L184-L187)) — a remote-deleted LWW doc leaks its entries into `listQuarantinedChanges()` for the process lifetime.

src/client/OTAlgorithm.ts
```diff
@@ -198,6 +192,20 @@ export class OTAlgorithm implements ClientAlgorithm {
     return this._withConsistentBaseRev(docId, pending);
   }
 
+  /**
+   * See {@link ClientAlgorithm.collectUnsyncedForDiscard}. The union {@link _collectPending}
+   * already computes: the sendable queue plus the withheld doc-only rows the send path refuses
+   * — refuses because they are not durable, which on a discard is the reason to include them.
+   * Raw rows, no {@link _withConsistentBaseRev}: this is a shelf payload, not a wire batch, and
+   * no report — the doc is legitimately vanishing, so the store-integrity alarm would misfire
```
RyanHavoc: "no report — the doc is legitimately vanishing" only holds on the push path. The discovery routes reach `_handleRemoteDocDeleted` through catch blocks whose `try` already called `getPendingToSend` ([PatchesSync.ts#L813](https://github.com/dabblewriter/patches/blob/f67fdff364ca479184e4f834b8204aa50835cf04/src/net/PatchesSync.ts#L813)) — for a doc with a withheld row that emits `UnstoredPendingError` and consumes the latch *before* DOC_DELETED is thrown and caught ([#L1284-L1286](https://github.com/dabblewriter/patches/blob/f67fdff364ca479184e4f834b8204aa50835cf04/src/net/PatchesSync.ts#L1284-L1286)). So the offline-discovery delete — "a delete from another device", the scenario the original comment named — still fires the alarm this design says would misfire. The new tests can't observe it because they call `_handleRemoteDocDeleted` directly ([PatchesSyncQueueOrdering.spec.ts#L293](https://github.com/dabblewriter/patches/blob/f67fdff364ca479184e4f834b8204aa50835cf04/tests/net/PatchesSyncQueueOrdering.spec.ts#L293)). Worth either narrowing the claim to the notification path or accepting the alarm there deliberately — a test that reaches the shelf through `syncDoc` would pin whichever you pick.

src/client/LWWAlgorithm.ts
```diff
@@ -150,6 +150,27 @@ export class LWWAlgorithm implements ClientAlgorithm {
...
+   */
+  async collectUnsyncedForDiscard(docId: string): Promise<Change[]> {
+    return this._withDocLock(docId, async () => {
+      const changes: Change[] = [];
+      const sendingChange = await this.store.getSendingChange(docId);
+      if (sendingChange) changes.push(sendingChange);
+      const pendingOps = await this.store.getPendingOps(docId);
+      if (pendingOps.length > 0) {
+        const committedRev = await this.store.getCommittedRev(docId);
+        changes.push(createChange(committedRev, committedRev + 1, pendingOps));
```
RyanHavoc: This mints a fresh random id on every call (`createChangeId` = rev prefix + 4 random chars, [change.ts#L11-L13](https://github.com/dabblewriter/patches/blob/f67fdff364ca479184e4f834b8204aa50835cf04/src/data/change.ts#L11-L13)) and never persists it — which interacts badly with `_handleRemoteDocDeleted` having no re-entrancy guard: the `onDocDeleted` push ([PatchesSync.ts#L264](https://github.com/dabblewriter/patches/blob/f67fdff364ca479184e4f834b8204aa50835cf04/src/net/PatchesSync.ts#L264)) can overlap the 410-catch route for the same doc, both runs collect before either wipes (the window spans closeDoc's queue drain), and the two `onRemoteDocDeleted` emits then carry the same ops under *different* ids. The old path was accidentally dedupable — the retry branch returned the persisted sending change with a stable id. Cheapest fixes: give the built change a deterministic id (`createChange` takes an explicit one for exactly the reuse case), or hold an in-flight promise per doc in `_handleRemoteDocDeleted` so the second trigger awaits the first.

RyanHavoc (commented):
Ran a deep pass over this. The change itself is right — the dedicated method beats the `{report:false}` mode flag, and the shelf is strictly more complete than before (withheld rows ride it, LWW returns both halves, no truncation at frame boundaries). 13 inline notes: the substantive ones are contract-vs-implementation gaps on the new docstring (the un-minted optimistic tier, quarantined changes, and the no-alarm claim on discovery paths) plus one LWW id-stability regression under the pre-existing double-delete race. The rest are small. Nothing blocking.

src/client/OTAlgorithm.ts
```diff
@@ -198,6 +192,20 @@ export class OTAlgorithm implements ClientAlgorithm {
...
+   * (and the once-per-doc latch stays untouched for any doc that survives).
+   */
+  async collectUnsyncedForDiscard(docId: string, doc?: PatchesDoc<any>): Promise<Change[]> {
+    const otDoc = doc as OTDoc<any> | undefined;
+    const { pending, withheld } = await this._collectPending(docId, otDoc, otDoc?.committedRev ?? 0);
```
RyanHavoc, line 212: Narrow window where already-committed rows ride the shelf as `withheld`: on a multi-batch flush, `dropResolvedPending` removes the committed batch from the store and defers the doc re-sync to `_reloadDocFromServer` ([PatchesSync.ts#L1214-L1217](https://github.com/dabblewriter/patches/blob/f67fdff364ca479184e4f834b8204aa50835cf04/src/net/PatchesSync.ts#L1214-L1217)). If that reload throws DOC_DELETED, the catch lands in `_handleRemoteDocDeleted` with the doc still mirroring the batch — it classifies `rev <= latestRev` and is shelved as "what the server does not have". The harm is a lie in the payload (inflated "N unsaved changes were lost"), not resurrection, but the window is deterministic rather than a race. Passing the just-resolved batch ids through to the discard collect would close it.

src/net/PatchesSync.ts
```diff
@@ -1823,16 +1826,13 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
...
+    // Collect everything unsynced before cleanup so the app can shelve it. WITH the open doc: the
+    // rows most at risk here are the doc-only ones the store never accepted — after the close
+    // below they exist nowhere else, and this payload is the app's last chance to shelve them.
+    const pendingChanges = await algorithm.collectUnsyncedForDiscard(
+      docId,
+      this.patches.getOpenDoc(docId) as PatchesDoc<any>
+    );
 
     // Close doc if open
     const doc = this.patches.getOpenDoc(docId);
```
RyanHavoc: `closeDoc` without `force` here, vs. `deleteDoc`'s `{ force: true }` with the comment "the doc is being deleted, every consumer needs to release it" ([Patches.ts#L594-L597](https://github.com/dabblewriter/patches/blob/f67fdff364ca479184e4f834b8204aa50835cf04/src/client/Patches.ts#L594-L597)). With refCount > 1 (reachable by mixing direct `openDoc` calls with the vue/solid managed-docs surfaces) the mirror survives over a store `confirmDeleteDoc` just wiped, subsequent edits recreate an orphan buffer that never syncs — and the new comment above ("after the close below they exist nowhere else") is false in that state: the shelf carries withheld rows the live doc still shows, so restoring the shelf duplicates them. Pre-existing, but the withheld inclusion is what gives it teeth; `{ force: true }` seems right here for the same reason it's right in `deleteDoc`.

src/net/PatchesSync.ts
```diff
@@ -1823,16 +1826,13 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
...
-    // exist nowhere else, and this payload is the app's last chance to shelve them. But without
-    // the report: classifying would raise `UnstoredPendingError` — a store-integrity alarm — at
-    // the exact moment the doc legitimately vanished (a collaborator losing access, a delete from
-    // another device).
-    const pendingChanges =
-      (await algorithm.getPendingToSend(docId, this.patches.getOpenDoc(docId) as PatchesDoc<any>, {
-        report: false,
-      })) ?? [];
+    // Collect everything unsynced before cleanup so the app can shelve it. WITH the open doc: the
+    // rows most at risk here are the doc-only ones the store never accepted — after the close
```
RyanHavoc: "after the close below they exist nowhere else" — on the IndexedDB store that's true only of the doc-only mirror rows. `confirmDeleteDoc` deletes just the `docs` row ([IndexedDBStore.ts#L692-L696](https://github.com/dabblewriter/patches/blob/f67fdff364ca479184e4f834b8204aa50835cf04/src/client/IndexedDBStore.ts#L692-L696)); the actual data wipe lives in `deleteDoc` ([OTIndexedDBStore.ts#L163-L184](https://github.com/dabblewriter/patches/blob/f67fdff364ca479184e4f834b8204aa50835cf04/src/client/OTIndexedDBStore.ts#L163-L184)), which this path never calls. So every remote delete orphans the doc's snapshots, committed history, pending queue and quarantine — unreachable once the docs row is gone, unbounded growth for a client that sees many remote deletes. Pre-existing, but this is the only `confirmDeleteDoc` call site without a preceding wipe, and the premise stated in this comment is what the leak falsifies. Probably wants `untrackDocs`/`deleteDoc`-style cleanup here, as a follow-up.

src/client/LWWAlgorithm.ts
```diff
@@ -150,6 +150,27 @@ export class LWWAlgorithm implements ClientAlgorithm {
...
   }
 
+  /**
+   * See {@link ClientAlgorithm.collectUnsyncedForDiscard}. Unlike {@link getPendingToSend} this
+   * neither mints a sending change nor clears pending ops — the doc is being discarded, nothing
+   * will send — and it returns BOTH halves when both exist: the in-flight sending change and a
+   * change built from the ops that accumulated behind it (which the send path's retry branch
+   * deliberately holds back). Revs on the built change are informational for a shelf payload.
+   */
+  async collectUnsyncedForDiscard(docId: string): Promise<Change[]> {
```
RyanHavoc, line 163: This body is the same sending+pending fold `applyServerChanges` writes inline in the follower-heal branch ([#L237-L248](https://github.com/dabblewriter/patches/blob/f67fdff364ca479184e4f834b8204aa50835cf04/src/client/LWWAlgorithm.ts#L237-L248)) — same reads, same conditional halves, same order. A private `_collectLocalChanges(docId, baseRev?)` serves both: the heal passes the rev paired with its `getCommittedState` read, this method lets it default to `getCommittedRev`. Both call sites already hold `_withDocLock`, so the helper must not take it (re-entrant acquisition would deadlock — the lock chains on its own tail). Keeps the two consumers — the shelf and the open-doc rebuild — from ever diverging on what "the local layers above committed" means.

src/client/ClientAlgorithm.ts
```diff
@@ -108,11 +108,23 @@ export interface ClientAlgorithm {
...
+   * chance to shelve that content before the doc closes, so it answers a different question than
+   * {@link getPendingToSend}: not "what is safe to put on the wire" but "what would vanish".
+   *
+   * A pure read with none of the send path's work or side effects: no batch normalization, no
+   * store-integrity reporting (the doc is legitimately vanishing, so the alarm would misfire),
+   * no sending-change minting (LWW). It also includes what the send path deliberately withholds
+   * — for OT, doc-only rows the store never accepted survive nowhere but the mirror being
+   * discarded, making them the one category the shelf exists for.
+   */
+  collectUnsyncedForDiscard(docId: string, doc?: PatchesDoc<any>): Promise<Change[]>;
```
RyanHavoc: First genuinely new *required* member on this interface in a while — `reconcilePending` and the quarantine quartet all landed optional-and-guarded. The call site is unguarded inside a fire-and-forget subscription with no catch ([PatchesSync.ts#L264](https://github.com/dabblewriter/patches/blob/f67fdff364ca479184e4f834b8204aa50835cf04/src/net/PatchesSync.ts#L264)), so a partial implementation (the TabAlgorithm/WorkerAlgorithm proxy this file's header advertises, or an untyped consumer) throws before closeDoc/untrack/confirmDeleteDoc/emit and the deleted doc stays tracked forever, surfacing only as an unhandled rejection. TS consumers get a compile error, so this is a consistency question rather than a bug — but `collectUnsyncedForDiscard?` plus `?? []` at the call site makes the shelf degrade to empty instead of wedging the doc, for the cost of one `?`.

src/client/ClientAlgorithm.ts
```diff
@@ -108,11 +108,23 @@ export interface ClientAlgorithm {
...
+  /**
+   * Everything the doc still holds that the server does not — durable or not — for a doc being
+   * discarded (a remote delete, a collaborator losing access). This payload is the app's last
+   * chance to shelve that content before the doc closes, so it answers a different question than
+   * {@link getPendingToSend}: not "what is safe to put on the wire" but "what would vanish".
+   *
+   * A pure read with none of the send path's work or side effects: no batch normalization, no
+   * store-integrity reporting (the doc is legitimately vanishing, so the alarm would misfire),
+   * no sending-change minting (LWW). It also includes what the send path deliberately withholds
+   * — for OT, doc-only rows the store never accepted survive nowhere but the mirror being
```
RyanHavoc, line 132: Worth one more sentence here on payload shape: OT returns durable rows first, then withheld — not rev-ordered, and revs can collide across the buckets (the queue-ordering spec builds exactly that) — and LWW's two halves share a rev by construction. The "durable rows first" intent currently lives only in a test comment, while array order is the de facto contract: an app that re-persists the shelf through `savePendingChanges` gets revs re-stamped by array position. Something like: "Durable rows first, then withheld; not rev-ordered — revs are informational and may collide."

src/net/PatchesSync.ts
```diff
@@ -1138,6 +1138,9 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
       // held (an empty store queue merges to `[...merged]`), an id the store-side peek can never
       // return — which would read any non-empty post-flush queue as progress. Both ends of the
       // compare must read the same collection; an algorithm with no peek compares batch to batch.
+      // `undefined` (empty pre-flush store queue) is load-bearing: any non-empty post-flush queue
+      // then reads as drained — the mid-flight-arrival case, whose follow-up flushes that row.
+      // Don't normalise it to null; the compare below must distinguish "no head" from any real id.
```
RyanHavoc: The last sentence pins a constraint that doesn't exist: `headBefore`'s only consumer is the `!==` compare at [#L1281](https://github.com/dabblewriter/patches/blob/f67fdff364ca479184e4f834b8204aa50835cf04/src/net/PatchesSync.ts#L1281), where `null` and `undefined` behave identically against a non-empty id string — and the source value already *is* `null` (`peekPendingHead` returns `Change | null`); the `?.` on this line is what turns it into `undefined`. The first two added lines are the load-bearing part and they're right. Suggest dropping the "Don't normalise it to null" sentence so nobody hunts for a dependency that isn't there — especially since `head !== undefined` at [#L1266](https://github.com/dabblewriter/patches/blob/f67fdff364ca479184e4f834b8204aa50835cf04/src/net/PatchesSync.ts#L1266) is a different variable where undefined genuinely means something else (no `peekPendingHead`).

tests/client/LWWAlgorithm-collectUnsyncedForDiscard.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+  const DOC = 'discard-doc';
+  const op = (path: string, value: unknown): JSONPatchOp => ({ op: 'replace', path, value, ts: 1, rev: 1 });
+
+  function setup() {
+    const store = new LWWInMemoryStore();
+    const algorithm = new LWWAlgorithm(store);
+    return { store, algorithm };
+  }
+
+  it('returns the in-flight sending change AND a change from the ops behind it', async () => {
```
RyanHavoc, line 24: Three small coverage gaps in this file: (1) the empty-doc test below never tracks the doc, so it exercises the store's unknown-doc early return rather than a tracked-but-drained buffer — the shape `_handleRemoteDocDeleted` actually hits for a synced doc (cf. `algorithm-hasPending.spec.ts`, which covers the two cases separately); (2) the pure-read property is asserted only in the test *without* a sending change — this both-halves test checks nothing about store state afterwards, so a regression that mutates only in this branch stays green; (3) the setup comment's "only writes into an existing doc buffer" is `LWWInMemoryStore`-specific — `LWWIndexedDBStore.saveSendingChange` puts unconditionally, so the stated rationale won't survive a port of this spec to the persistent store.

src/net/PatchesSync.ts
```diff
@@ -1823,16 +1826,13 @@ export class PatchesSync extends ReadonlyStoreClass<PatchesSyncState> {
...
-    const pendingChanges =
-      (await algorithm.getPendingToSend(docId, this.patches.getOpenDoc(docId) as PatchesDoc<any>, {
-        report: false,
-      })) ?? [];
+    // Collect everything unsynced before cleanup so the app can shelve it. WITH the open doc: the
+    // rows most at risk here are the doc-only ones the store never accepted — after the close
+    // below they exist nowhere else, and this payload is the app's last chance to shelve them.
+    const pendingChanges = await algorithm.collectUnsyncedForDiscard(
+      docId,
+      this.patches.getOpenDoc(docId) as PatchesDoc<any>
```
RyanHavoc: The `as PatchesDoc<any>` is load-bearing for nothing — the param is already `doc?: PatchesDoc<any>` and the same call compiles uncast at [#L2041](https://github.com/dabblewriter/patches/blob/f67fdff364ca479184e4f834b8204aa50835cf04/src/net/PatchesSync.ts#L2041) (twin cast at [#L1121](https://github.com/dabblewriter/patches/blob/f67fdff364ca479184e4f834b8204aa50835cf04/src/net/PatchesSync.ts#L1121) if you clean both). And since `closeDoc` opens with the same existence check, lines 1837-1841 collapse to a bare `await this.patches.closeDoc(docId);`. Don't hoist a single `const` instead, though — a doc opened during the collect await must still be closed, so the fresh read (or the bare call) is the correct form.

src/client/ClientAlgorithm.ts
jacwright: Both tightened.

I tried the drain-before-collect first, and it is not safe: `BaseDoc.flush()` never settles for a write-latched doc. The latched handler returns without persisting and leaves the ops in `_optimisticOps`, while the `_changeQueues` entry stays put, so flush's `await tail` resolves immediately on the same already-settled promise every lap. It is not a slow flush, it is an unbounded microtask loop that starves the macrotask queue outright (a probe hung a vitest worker past 140s, and the test's own 20s timeout could not fire; the same probe on an unlatched doc settles in 3ms). Calling that from `_handleRemoteDocDeleted` would wedge the sync path on exactly the docs the shelf exists for, so I did not force it.

So the docstring now scopes the claim to the store's tiers (durable pending, withheld doc-only rows, quarantine) and says why the un-minted optimistic tier is out of reach. Worth a follow-up on flush itself: `retrySavingChanges` runs the same drain-to-fixed-point loop without the optimistic-ops condition and terminates fine, so a latch-aware flush looks reachable.

Revocation sentence is gone. It now reads as the 410/`docDeleted` class, notes a 403 latches at `'error'` and never routes here, and keeps revocation only as "if a downstream server signals it as a delete".

src/client/OTAlgorithm.ts
jacwright: Both halves done.

Quarantine now appends to the shelf in OT and LWW, after the durable and withheld rows, guarded with `?.` so a store without the method degrades quietly. Also added the note to docs/quarantine.md §4, since "only deleting the doc removes them" was the line the discard path was quietly falsifying.

`LWWInMemoryStore.confirmDeleteDoc` clears quarantine now, matching `OTInMemoryStore`.

Tests: `OTAlgorithm-getPendingToSend.spec.ts` "carries quarantined changes, which confirmDeleteDoc is about to drop", `LWWAlgorithm-collectUnsyncedForDiscard.spec.ts` "carries quarantined changes, which the discard is about to drop", `LWWInMemoryStore.spec.ts` "should drop the quarantine with the doc". All three fail on f67fdff.

src/client/OTAlgorithm.ts
jacwright: Narrowed, and pinned with a test through `syncDoc`.

The docstring no longer claims the alarm never fires for a deleted doc. It says skipping the report is right on the notification/push route, and that the discovery routes reach here from a catch whose `try` already called `getPendingToSend` and consumed the latch. What it now guarantees is the narrower thing that is actually true: the shelf read adds no report of its own and leaves the latch untouched for a surviving doc.

`PatchesSyncQueueOrdering.spec.ts` "reaches the shelf through syncDoc when the flush takes a 410" drives a real 410 out of `commitChanges` and asserts both the shelf payload and that the `UnstoredPendingError` did fire on that route, so the accepted alarm is now a pinned expectation rather than an unobserved one.

src/client/LWWAlgorithm.ts
jacwright: Took the in-flight-promise option, since it fixes the double emit for any overlapping pair of delete routes rather than just making the two payloads comparable.

`_handleRemoteDocDeleted` keeps a `Map<docId, Promise<void>>`; a second trigger for the same doc returns the first run's promise, and the entry is dropped in a `finally`. The body moved to `_discardRemoteDeletedDoc`.

Test: `PatchesSyncQueueOrdering.spec.ts` "shelves once when two delete routes name the same doc" fires two concurrent calls and asserts one collect and one emit. On f67fdff it reports 2 collects.

src/client/OTAlgorithm.ts
jacwright: Plumbed through. `flushDoc` accumulates a `resolvedIds` set as each batch is confirmed and dropped (both the `docReloadRequired` branch and the normal one), and passes it to `_handleRemoteDocDeleted`, which passes it to the collect. The parameter is optional on the interface and on both implementations, so every other call site is untouched.

I did not build the full multi-batch flush window in the harness. Constructing a reload that throws DOC_DELETED while the doc still mirrors the committed batch is a lot of scaffolding for a filter, so the coverage is a focused test on the filter itself: `OTAlgorithm-getPendingToSend.spec.ts` "drops ids the flush already resolved, so committed rows do not ride as withheld" builds the mirror-holds-a-committed-row state directly and asserts the row rides without the set and not with it. It fails on f67fdff. The plumbing from `flushDoc` into the collect is verified by the type checker and by the existing flush suite, not by a test that exercises that exact window.

src/net/PatchesSync.ts
jacwright: Agreed, changed to `closeDoc(docId, { force: true })` with the rationale spelled out next to it, pointing at `deleteDoc`'s. Folded in with the collapse from the note below.

src/net/PatchesSync.ts
jacwright: Not implementing the cleanup here, taking it as a follow-up as you suggested.

Fixed the false premise. The comment no longer says the rows "exist nowhere else" after the close. It now says what actually happens: the doc-only rows live only in the mirror the close evicts, and `confirmDeleteDoc` drops the doc's tracking row (the in-memory stores drop their buffers with it), which is what makes this payload the last chance to shelve any of it. The orphaning on the persistent stores is left for the follow-up rather than asserted away.

src/client/LWWAlgorithm.ts
jacwright: Extracted `LWWAlgorithm._collectLocalChanges(docId, baseRev?)`, used by `collectUnsyncedForDiscard` and the `applyServerChanges` follower-heal branch. It does not take `_withDocLock` and the docblock says why.

Two details kept deliberately: the committed-rev read stays lazy (only when there are pending ops), so the heal path's `baseRev` still means zero extra reads; and the heal branch's late `rev > doc.committedRev` re-check stays after the sending/pending reads, so a concurrent `applySnapshot` landing during them is still caught. Existing tests for both paths are green.

src/client/ClientAlgorithm.ts
jacwright: Done, for the cost of one `?`. `collectUnsyncedForDiscard?` on the interface, `await algorithm.collectUnsyncedForDiscard?.(...) ?? []` at the call site, and the docblock says the shelf degrades to empty rather than wedging the doc. Both concrete implementations stay required in their classes.

src/client/ClientAlgorithm.ts
jacwright: Added, near enough verbatim: "Durable rows first, then withheld, then quarantined; not rev-ordered — revs are informational and may collide across the buckets."

src/net/PatchesSync.ts
jacwright: Dropped the sentence. First two lines kept.

tests/client/LWWAlgorithm-collectUnsyncedForDiscard.spec.ts
jacwright: All three.

1. Added "returns an empty array for a tracked doc whose local layers have drained": tracks the doc, mints and confirms a sending change, then expects `[]`. The unknown-doc case stays as its own test right after it, so both shapes are covered separately.
2. The both-halves test now asserts store state afterwards: the sending change is still in the slot and the ops behind it are still pending. A regression that mutates only in that branch fails now.
3. Reworded the setup comment to name `LWWInMemoryStore` and to say what changes on a port: `LWWIndexedDBStore.saveSendingChange` puts unconditionally, so the priming write is what the port drops, not the assertions.

src/net/PatchesSync.ts
jacwright: Both casts gone (this one and the twin at #L1121); the type checker is happy without them. Lines 1837-1841 collapse to the bare call, no hoisted `const`, and it carries `{ force: true }` per the note above.

RyanHavoc (approved):

PR #149, opened 2026-08-19
fix(json-patch): drop @txt retain overruns instead of padding them with spaces (DAB-1064)

matalina:
Fixes the destructive half of DAB-1064: live co-authoring silently corrupting a manuscript.

## What was happening

A `@txt` change whose leading retain runs past the end of the document leaves the overrun stranded in the composed result as a bare `{ retain: n }` — structurally invalid, since a document is insert-only. `fixBadDeltaDoc` converted that into space inserts, deliberately, to keep the author's later offsets valid.

That trades a misplaced edit for a corrupted one. The spaces are indistinguishable from typed text, so the document grows past its own content and reports a length longer than the text it holds — and every subsequent offset then lands in the wrong place.

## How we know

Replaying a reporter's database export (the co-authoring session in which it happened) reproduces her screenshot character for character:

| rev | author | op | commit latency |
| --- | --- | --- | --- |
| 3395 | co-author | `retain 1258`, `insert "<I'm never "` (11 chars) | **7952 ms** |
| 3396 | co-author | `retain 1280`, `insert "impulsive.>"` | 3115 ms |

The document was **1269** characters long. `retain 1280` overruns by **11** — exactly the length of that author's own still-in-flight insert, counted twice. The 11 became 11 spaces, and the next change (also `retain 1280`) inserted itself *in front of* `impulsive.>`, shunting it to the end of the line.

Because the padding is committed, the damage survives refresh, tab close and re-login. The reporter's only recovery was retyping the sentence by hand — four times so far.

## The change

Drop the overrun rather than padding it. The offending change lands at the end of the document, where the author was typing anyway — a misplaced edit instead of a corrupted one.

This also makes the function self-consistent: *trailing* overruns were already dropped a few lines above; only mid-document ones became spaces.

## Tests

- `tests/json-patch/txt-retain-overrun.spec.ts` — new, built from the exact op sequence in the export. Covers: no invented characters, no non-insert op left in the document, document length stays honest, the following edit is not transposed, and an in-bounds retain still works (guarding against over-correcting).
- `tests/json-patch/json-patch.spec.ts` — the existing test asserted the padding by name. Updated to assert the drop, with the reasoning recorded.

Full suite: 3501 passed. `type:check`, `lint`, `format:check` clean. (Three `tests/solid/*.spec.tsx` files fail to load on `main` as well — pre-existing, unrelated.)

## Scope

**This does not fix the cause of the bad offset** — the client composing a change while its own previous change is un-acked and double-counting that insert's length. That is the next piece of work and is tracked on the ticket. Until it lands, an affected author may see their edit land at the end of the document instead of at the caret. That is the intended trade: a visible misplacement beats silent corruption.

No telemetry yet on the drop path (a `log()` call marks it); a counter for `retain > docLength` is the follow-up that will tell us the real fleet-wide rate.

### DIFF
--- a/src/json-patch/ops/text.ts
+++ b/src/json-patch/ops/text.ts
@@ -89,9 +89,20 @@ export const text: JSONPatchOpHandler = {
 };
 
 /**
- * Fix non-insert ops (retain/delete that overran the document)
- * Convert retains to space inserts to preserve cursor positions and subsequent edits
- * Ensure document ends with a newline
+ * Drop non-insert ops (retain/delete that overran the document) and ensure the document
+ * ends with a newline.
+ *
+ * A document is insert-only, so `compose` consumes every in-bounds retain. Anything left
+ * refers to content past the end of the document — i.e. the change was authored against a
+ * longer document than the one it landed on.
+ *
+ * These used to be converted to space inserts, to keep the author's later offsets valid.
+ * That trades a misplaced edit for a corrupted one: the spaces are indistinguishable from
+ * typed text, so the document silently grows past its own content and every later offset
+ * lands in the wrong place. DAB-1064 traced a reader's scrambled manuscript to exactly
+ * this — an 11-character overrun became 11 spaces, and the author's next edit inserted
+ * itself in front of the previous one. Dropping the overrun keeps the document honest;
+ * the offending change simply lands at the end, where the author was typing anyway.
  */
 function fixBadDeltaDoc(delta: Delta): Delta {
   // Find where trailing non-inserts start (these can be dropped)
@@ -106,21 +117,20 @@ function fixBadDeltaDoc(delta: Delta): Delta {
     delta.push({ insert: '\n' });
   }
 
-  // Check if we need to fix any middle-of-doc retains
+  // Check if we need to drop any middle-of-doc retains
   const hasNonInsertOps = delta.ops.some(op => op.insert === undefined);
   if (!hasNonInsertOps) {
     return delta;
   }
-  const newDelta = new Delta();
+  log(
+    'Dropping ops that overran the document in a @txt doc',
+    delta.ops.filter(op => op.insert === undefined)
+  );
 
+  const newDelta = new Delta();
   for (const op of delta.ops) {
     if (op.insert !== undefined) {
       newDelta.push(op);
-    } else if (op.retain) {
-      // Convert retain to spaces to preserve cursor positions
-      const insertOp: Op = { insert: ''.padStart(op.retain) };
-      if (op.attributes) insertOp.attributes = op.attributes;
-      newDelta.push(insertOp);
     }
   }
   return newDelta;
--- a/tests/json-patch/txt-retain-overrun.spec.ts
+++ b/tests/json-patch/txt-retain-overrun.spec.ts
@@ -1,0 +1,87 @@
+import { Delta } from '@dabble/delta';
+import { describe, expect, it } from 'vitest';
+import { applyPatch } from '../../src/json-patch/applyPatch.js';
+
+/**
+ * DAB-1064 — a `@txt` change whose leading retain runs past the end of the document.
+ *
+ * A client that composes a change while its own previous change is still in flight can
+ * count that insert's length twice, so the next change retains past the end of the doc.
+ * `Delta.compose` leaves the overrun stranded in the result as a bare `{ retain: n }`,
+ * which is structurally invalid in a document (documents are insert-only).
+ *
+ * The overrun must not become content. Materialising it as spaces inflates the document
+ * past its own text, so every later offset lands in the wrong place — which corrupts the
+ * manuscript rather than merely misplacing one edit.
+ *
+ * The sequence below is taken verbatim from a reporter's database export (co-authoring
+ * session, 2026-08-18): doc length 1269, two successive changes both retaining 1280.
+ */
+describe('@txt with a retain past the end of the document', () => {
+  // Reproduces the exact offsets from the export: 1258 chars of prior text, then the
+  // author types "<I'm never " (11 chars), leaving the document 1269 long.
+  const PRIOR = 'x'.repeat(1258);
+  const OPENING = "<I'm never ";
+  const docText = PRIOR + OPENING;
+
+  const textOf = (value: unknown) =>
+    new Delta(value as any).ops
+      .filter(op => typeof op.insert === 'string')
+      .map(op => op.insert as string)
+      .join('');
+
+  const applyTxt = (state: any, ops: any[]) => applyPatch(state, [{ op: '@txt', path: '/text', value: ops }]);
+
+  it('does not invent characters that nobody typed', () => {
+    const state = { text: new Delta().insert(docText).ops };
+
+    // retain 1280 is 11 past the end of the 1269-char document — exactly the length of
+    // the author's own preceding insert, counted twice.
+    const result = applyTxt(state, [{ retain: 1280 }, { insert: 'impulsive.>' }]);
+
+    expect(textOf(result.text)).toBe(`${docText}impulsive.>\n`);
+  });
+
+  it('leaves no non-insert op stranded in the document', () => {
+    const state = { text: new Delta().insert(docText).ops };
+
+    const result = applyTxt(state, [{ retain: 1280 }, { insert: 'impulsive.>' }]);
+
+    const ops = new Delta(result.text as any).ops;
+    expect(ops.every(op => op.insert !== undefined)).toBe(true);
+  });
+
+  it('keeps the document length honest so later edits land correctly', () => {
+    const state = { text: new Delta().insert(docText).ops };
+
+    const result = applyTxt(state, [{ retain: 1280 }, { insert: 'impulsive.>' }]);
+    const doc = new Delta(result.text as any);
+
+    // A document that reports a length longer than its own text poisons every
+    // subsequent offset; that is the mechanism behind the reported transposition.
+    expect(doc.length()).toBe(textOf(result.text).length);
+  });
+
+  it('does not transpose the next edit (the reported corruption)', () => {
+    let state: any = { text: new Delta().insert(docText).ops };
+
+    // Both changes retain 1280 — the second is correct in the author's own frame.
+    state = applyTxt(state, [{ retain: 1280 }, { insert: 'impulsive.>' }]);
+    state = applyTxt(state, [{ retain: 1280 }, { insert: ' Except for with Aster.' }]);
+
+    const text = textOf(state.text);
+
+    // Reported symptom: "impulsive.>" is shunted to the end, behind text typed after it.
+    expect(text).not.toMatch(/Except for with Aster\.impulsive\.>/);
+    expect(text).toContain('impulsive.> Except for with Aster.');
+  });
+
+  it('still preserves a retain that is within the document', () => {
+    const state = { text: new Delta().insert(docText).ops };
+
+    // Guard against over-correcting: an in-bounds retain is ordinary and must still work.
+    const result = applyTxt(state, [{ retain: 1269 }, { insert: 'impulsive.>' }]);
+
+    expect(textOf(result.text)).toBe(`${docText}impulsive.>\n`);
+  });
+});
--- a/tests/json-patch/json-patch.spec.ts
+++ b/tests/json-patch/json-patch.spec.ts
@@ -197,14 +197,17 @@ describe('JSONPatch', () => {
     expect(obj).toEqual({ test: true, sub: { name: 'Test' } });
   });
 
-  it('handles text delta that overruns document by converting retain to spaces', () => {
+  it('handles text delta that overruns document by dropping the overrun', () => {
     obj.text = new Delta().insert('Short\n');
     patch.text('/text', new Delta().retain(21).insert('Long').ops);
     obj = patch.apply(obj, { strict: true });
     // The retain(21) overruns the 6-char document, creating a retain(15) op after compose.
-    // The lenient behavior converts the retain to spaces to preserve cursor positions.
+    // That overrun is dropped rather than converted to spaces: padding would put 15
+    // characters nobody typed into the manuscript and leave the document reporting a
+    // length longer than its own text, so every later edit lands at a shifted offset
+    // (DAB-1064). Dropping it lands the insert at the end, where the author was typing.
     // Delta normalizes adjacent string inserts into one op. Ensures trailing newline.
-    expect(obj.text.ops).toEqual([{ insert: 'Short\n' + ''.padStart(15) + 'Long\n' }]);
+    expect(obj.text.ops).toEqual([{ insert: 'Short\nLong\n' }]);
   });
 
   it('handles text delta with trailing retain/delete overrun by dropping them', () => {


tests/json-patch/txt-retain-overrun.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    // Reported symptom: "impulsive.>" is shunted to the end, behind text typed after it.
+    expect(text).not.toMatch(/Except for with Aster\.impulsive\.>/);
+    expect(text).toContain('impulsive.> Except for with Aster.');
+  });
+
+  it('still preserves a retain that is within the document', () => {
+    const state = { text: new Delta().insert(docText).ops };
+
+    // Guard against over-correcting: an in-bounds retain is ordinary and must still work.
```
jacwright: This one doesn't guard what the comment says it guards, and it's worth knowing before it's relied on.

An in-bounds retain is fully consumed by `compose`, so nothing is stranded, `hasNonInsertOps` is false, and the function returns at the early exit — this case never reaches the drop at all.

Proved rather than argued: I made the mid-doc branch `throw` outright. Four of the five cases in this file fail; this one still passes.

```
=== mid-doc branch poisoned ===
 fails: does not invent characters that nobody typed
        does not transpose the next edit (the reported corruption)
        keeps the document length honest so later edits land correctly
        leaves no non-insert op stranded in the document
 passes: still preserves a retain that is within the document
```

The coverage you wanted does exist, just not here: `does not invent characters that nobody typed` asserts the insert *after* the overrun survives, so dropping too much fails it. That is the real over-correction guard.

No need to delete this case — an early-exit regression is worth a test too. But the comment claims it protects the drop, and it protects `compose`. If you want a case that actually exercises over-correction inside the branch, it needs a stranded op *and* something after it that must be kept, which is the first test's shape.

src/json-patch/ops/text.ts
```diff
@@ -106,21 +117,20 @@ function fixBadDeltaDoc(delta: Delta): Delta {
...
   }
 
-  // Check if we need to fix any middle-of-doc retains
+  // Check if we need to drop any middle-of-doc retains
   const hasNonInsertOps = delta.ops.some(op => op.insert === undefined);
   if (!hasNonInsertOps) {
     return delta;
   }
-  const newDelta = new Delta();
+  log(
```
jacwright: Worth knowing that this emits nothing anywhere we run it, so the description's "a `log()` call marks it" is doing less than it sounds like.

`utils/log.ts` is:

```ts
export let log: (...args: any[]) => void = () => undefined;
export function verbose(value: boolean) { log = value ? console.log : () => undefined; }
```

It's a no-op until something calls `verbose(true)`. I grepped both consumers on their mains: neither pup nor dabble-writer-3.0 calls it anywhere. So on the fleet this line runs and produces no record.

That doesn't argue for changing it here — the live binding means `verbose(true)` still works for local debugging, and that's a fair thing to keep. It does mean the follow-up counter you mention is the *only* way we'll ever learn the real rate, so it's load-bearing rather than nice-to-have. If you want a signal before that lands, this is the spot where a Sentry breadcrumb or a `console.warn` would cost nothing and would tell us within a day whether this is one unlucky co-authoring session or a steady drip across the fleet.

Either way the filtered array here is the right payload — it's the overrun lengths that identify the double-count.

src/json-patch/ops/text.ts
```diff
@@ -89,9 +89,20 @@ export const text: JSONPatchOpHandler = {
...
- * Convert retains to space inserts to preserve cursor positions and subsequent edits
- * Ensure document ends with a newline
+ * Drop non-insert ops (retain/delete that overran the document) and ensure the document
+ * ends with a newline.
+ *
+ * A document is insert-only, so `compose` consumes every in-bounds retain. Anything left
+ * refers to content past the end of the document — i.e. the change was authored against a
+ * longer document than the one it landed on.
+ *
+ * These used to be converted to space inserts, to keep the author's later offsets valid.
```
jacwright, line 99: Good docblock. Keeping the *reason* the padding existed, rather than deleting it as an obvious mistake, is what stops someone reinstating it the next time an author reports an edit landing in the wrong place — which is the symptom this change deliberately accepts.

One thing worth adding while you're here: the padding was strictly worse than it looks, because it applied to `retain(n, attributes)` too. A formatting op that overran the end produced *n spaces carrying that formatting* — invented characters that are also styled. Dropping handles both cases the same way, and saying so here closes off the "well, could we at least keep formatting retains?" question before someone asks it.

jacwright (approved):
Approved. The diagnosis is right, the trade is the right one, and the fix is genuinely pinned rather than merely tested.

**Controls (anchors asserted unique, full suite each, 136/136 files and 3536 passed at baseline):**

| change | fails |
| --- | --- |
| **revert to the space padding** | **3** — `does not invent characters that nobody typed`, `does not transpose the next edit`, and the rewritten `json-patch.spec` case |
| **skip the mid-doc drop** (stranded retain survives) | **4** — the two above plus `leaves no non-insert op stranded` and `keeps the document length honest` |
| remove the *trailing* overrun drop (pre-existing, untouched here) | 1 — `handles text delta with trailing retain/delete overrun by dropping them` |
| remove the `log()` call | nothing |

The first two rows are what matter: the old behaviour cannot come back without failing a test that names the corruption, and the two failure modes are pinned by different tests rather than one test standing in for both. The third row confirms the half you left alone is still covered.

I also read the change against the ops it can actually receive. A stranded `delete` was already dropped before this PR (the old `else if (op.retain)` never matched it), so the only behavioural change is retains, which is what the description says. A `retain(n, attributes)` overrun previously became *n formatted spaces*, which is the same bug wearing a nicer hat, and is now dropped with everything else. The trailing-newline guarantee survives the rebuild, since the `\n` pushed above is an insert.

## One correction to the description

> Three `tests/solid/*.spec.tsx` files fail to load on `main` as well — pre-existing, unrelated.

They don't. On a fresh `npm ci` in a clean worktree at `2d55734b`: **136/136 test files, 3536 passed, 4 skipped, 0 failed** — `tests/solid/context.spec.tsx` among the passes. `type:check`, `lint` and `format:check` are clean too.

This is the second PR carrying that same line (patches#147 had it, and a clean install disproved it there as well). It's a stale `node_modules` in the working checkout, not something in the repo — the solid specs fail to *load*, which is the signature of a missing or half-installed peer rather than a broken test. A real `npm ci` clears it. Worth fixing at the source, because it also explains the count: you reported 3501 where the suite actually has 3540.

## Two notes on the tests, neither blocking

Detail inline. Briefly: the case labelled as the over-correction guard never reaches the code this PR changes, and the `log()` marker is a no-op in every environment that runs this.

## What still has to happen for the reporter to stop losing text

This branch is at `0.27.0` and cut before the `0.28.0` release, so it's 16 commits behind — no drift on the files it touches, and it merges clean, but merging doesn't ship it. patches has no release-please, so the version bump and publish are manual, and then pup and dabble-writer-3.0 both need their pin moved before a single client stops padding. Until that chain finishes, the manuscript-corrupting path is still live for everyone.

Given that this is silent, durable data corruption, I'd cut `0.28.1` and move both pins as one piece of work rather than letting the publish wait on the follow-up that fixes the cause.

src/json-patch/ops/text.ts
```diff
@@ -106,21 +117,20 @@ function fixBadDeltaDoc(delta: Delta): Delta {
...
     return delta;
   }
-  const newDelta = new Delta();
+  log(
+    'Dropping ops that overran the document in a @txt doc',
+    delta.ops.filter(op => op.insert === undefined)
+  );
 
+  const newDelta = new Delta();
   for (const op of delta.ops) {
```
RyanHavoc: This changes how already-committed history materializes, not just new applies. OT never persists head state — reads stream the raw version blob plus raw changes ([getSnapshotAtRevision.ts L110-L119](https://github.com/dabblewriter/patches/blob/2d55734bb5034d77c4f938ed987804fd1b522cf3/src/algorithms/ot/server/getSnapshotAtRevision.ts#L110-L119)) and they're replayed through strict `applyPatch` ([applyChanges.ts L54-L56](https://github.com/dabblewriter/patches/blob/2d55734bb5034d77c4f938ed987804fd1b522cf3/src/algorithms/ot/shared/applyChanges.ts#L54-L56)), which lands here. So a log containing an overrun replays differently once this deploys.

Concrete case from the DAB-1064 doc itself: after the padding appeared, deleting the 11 visible spaces committed `[{retain:1269},{delete:11}]`. Under the new rule the padding never exists on replay, so that in-bounds delete erases `impulsive.>` — real typed text — and the log below never fires, because nothing overruns.

Exposure is bounded by persisted version blobs (the un-versioned tail, plus full-history replay fallbacks), but there's no version gate or repair sweep, and the settled-history premise in [algorithms.md L125-L127](https://github.com/dabblewriter/patches/blob/2d55734bb5034d77c4f938ed987804fd1b522cf3/docs/algorithms.md#L125-L127) ("the committed head is the truth and the replay merely reconstructs it") no longer holds for these docs. Worth a deploy plan: identify docs with stranded overruns in un-versioned tails before this lands.

src/json-patch/ops/text.ts
```diff
@@ -89,9 +89,20 @@ export const text: JSONPatchOpHandler = {
...
+ *
+ * These used to be converted to space inserts, to keep the author's later offsets valid.
+ * That trades a misplaced edit for a corrupted one: the spaces are indistinguishable from
+ * typed text, so the document silently grows past its own content and every later offset
+ * lands in the wrong place. DAB-1064 traced a reader's scrambled manuscript to exactly
+ * this — an 11-character overrun became 11 spaces, and the author's next edit inserted
+ * itself in front of the previous one. Dropping the overrun keeps the document honest;
+ * the offending change simply lands at the end, where the author was typing anyway.
  */
 function fixBadDeltaDoc(delta: Delta): Delta {
```
RyanHavoc: [`invert` (L73-L81)](https://github.com/dabblewriter/patches/blob/2d55734bb5034d77c4f938ed987804fd1b522cf3/src/json-patch/ops/text.ts#L73-L81) still mints the inverse from the raw ops in the pre-drop frame, so undo of exactly this change class no longer works. The inverse of `[{retain:1280},{insert:'impulsive.>'}]` is `[{retain:1280},{delete:11}]`; against the dropped doc the delete consumes only the trailing newline (which this function re-adds) — undo removes nothing. And if the author has typed more after the misplaced edit, the delete lands on that later real text instead. Under padding the same inverse deleted the insert correctly.

Same-size fix: clamp the change to `oldValue.length()` before `.invert(oldValue)`, plus a test pinning it. This is reachable via `JSONPatch.invert` and `invertPatch`, so it's not just interactive undo — see the ejection note below.

src/json-patch/ops/text.ts
```diff
@@ -106,21 +117,20 @@ function fixBadDeltaDoc(delta: Delta): Delta {
     delta.push({ insert: '\n' });
   }
 
-  // Check if we need to fix any middle-of-doc retains
+  // Check if we need to drop any middle-of-doc retains
   const hasNonInsertOps = delta.ops.some(op => op.insert === undefined);
```
RyanHavoc: Because the repair is silent — apply never returns an error string for an overrun — an overrunning change passes the eject probe at [ejectPendingChange.ts L82-L97](https://github.com/dabblewriter/patches/blob/2d55734bb5034d77c4f938ed987804fd1b522cf3/src/algorithms/ot/shared/ejectPendingChange.ts#L82-L97), which exists precisely to refuse an inverse for a change that doesn't apply cleanly in its own frame. Ejecting an overrunning poison now walks every surviving pending change through the mismatched inverse (see the `invert` note): the poison's text stays in the doc while the survivors are re-seated N characters off.

src/json-patch/ops/text.ts
```diff
@@ -106,21 +117,20 @@ function fixBadDeltaDoc(delta: Delta): Delta {
...
-  const newDelta = new Delta();
+  log(
+    'Dropping ops that overran the document in a @txt doc',
+    delta.ops.filter(op => op.insert === undefined)
+  );
 
+  const newDelta = new Delta();
   for (const op of delta.ops) {
     if (op.insert !== undefined) {
       newDelta.push(op);
```
RyanHavoc: A single change with two stranded inserts glues them together and deletes the gap between them ([L130-L135](https://github.com/dabblewriter/patches/blob/2d55734bb5034d77c4f938ed987804fd1b522cf3/src/json-patch/ops/text.ts#L130-L135)): doc `'The end of chapter one.\n'` (24 chars), value `[{retain:30},{insert:'Hello'},{retain:5},{insert:'world'}]` → `'…Helloworld\n'`, where the same edits applied as separate changes give `'…Hello\nworld\n'`.

That op shape isn't exotic — it's exactly what `Delta.compose` produces when two overrunning bursts are squashed by client batching (`[{retain:30},{insert:'Hello'}]` composed with `[{retain:40},{insert:'world'}]` yields it verbatim). Padding preserved the separation. None of the new tests cover more than one stranded insert — worth a case.

src/json-patch/ops/text.ts
```diff
@@ -89,9 +89,20 @@ export const text: JSONPatchOpHandler = {
 };
 
 /**
- * Fix non-insert ops (retain/delete that overran the document)
- * Convert retains to space inserts to preserve cursor positions and subsequent edits
- * Ensure document ends with a newline
+ * Drop non-insert ops (retain/delete that overran the document) and ensure the document
+ * ends with a newline.
+ *
+ * A document is insert-only, so `compose` consumes every in-bounds retain. Anything left
```
RyanHavoc, line 95: [`compose` (L82-L88)](https://github.com/dabblewriter/patches/blob/2d55734bb5034d77c4f938ed987804fd1b522cf3/src/json-patch/ops/text.ts#L82-L88) keeps the overrun while `apply` now drops it, so squashed and sequential application of the same edits diverge: doc `'x'.repeat(19)+'\n'`, `retain(25)+insert('ABC')` then `retain(28)+insert('DEF')` → sequential gives `…ABC\nDEF\n`, squashed gives `…ABCDEF\n`. Under padding both were byte-identical.

`composePatch` is public API ([composePatch.ts L7-L9](https://github.com/dabblewriter/patches/blob/2d55734bb5034d77c4f938ed987804fd1b522cf3/src/json-patch/composePatch.ts#L7-L9)) and micro's pending-field consolidation composes text deltas the same way ([micro/ops.ts L83-L85](https://github.com/dabblewriter/patches/blob/2d55734bb5034d77c4f938ed987804fd1b522cf3/src/micro/ops.ts#L83-L85)) — and micro has no `fixBadDeltaDoc` at all, so a stranded retain there is stored verbatim.

src/json-patch/ops/text.ts
```diff
@@ -89,9 +89,20 @@ export const text: JSONPatchOpHandler = {
...
+ * refers to content past the end of the document — i.e. the change was authored against a
+ * longer document than the one it landed on.
+ *
+ * These used to be converted to space inserts, to keep the author's later offsets valid.
+ * That trades a misplaced edit for a corrupted one: the spaces are indistinguishable from
+ * typed text, so the document silently grows past its own content and every later offset
```
RyanHavoc, line 101: Rollout skew: a replica on an old patches version pads while upgraded ones drop, so identical committed streams materialize into documents differing by the overrun length — and nothing detects a pure length skew (no checksum or state-hash comparison in the sync layer; only `ApplyChangesError` is loud, and this never errors). pup floats `^0.22.0` while dw3 pins an exact version, so a window where both rules coexist is guaranteed. Worth a coordinated-deploy note in the release.

src/json-patch/ops/text.ts
RyanHavoc: `log` is `() => undefined` unless something calls `verbose(true)` — nothing in src or any consumer does ([log.ts L1-L5](https://github.com/dabblewriter/patches/blob/2d55734bb5034d77c4f938ed987804fd1b522cf3/src/json-patch/utils/log.ts#L1-L5)) — so in production this drop is invisible, and the PR body's "a log() call marks it" doesn't hold. Two more gaps: trailing overruns are popped at L109-L111 *before* this line, so that shape is never logged at all; and the `.filter()` argument is allocated eagerly even with the logger off.

The repo's convention for recoverable-but-anomalous events is `console.warn`/`console.error` (e.g. [applyChanges.ts L133-L135](https://github.com/dabblewriter/patches/blob/2d55734bb5034d77c4f938ed987804fd1b522cf3/src/algorithms/ot/shared/applyChanges.ts#L133-L135)); a warn with the overrun length would give the follow-up telemetry ticket something to size against, given the producer bug is still live.

src/json-patch/ops/text.ts
```diff
@@ -89,9 +89,20 @@ export const text: JSONPatchOpHandler = {
...
+ * lands in the wrong place. DAB-1064 traced a reader's scrambled manuscript to exactly
+ * this — an 11-character overrun became 11 spaces, and the author's next edit inserted
+ * itself in front of the previous one. Dropping the overrun keeps the document honest;
+ * the offending change simply lands at the end, where the author was typing anyway.
```
RyanHavoc: With a realistic newline-terminated doc the dropped edit lands *after* the final newline — its own paragraph — not where the author was typing: `'abc\n'` + `[{retain:10},{insert:'X'}]` → `'abc\nX\n'`. The new spec doesn't catch this because its `docText` has no trailing newline ([spec L22-L26](https://github.com/dabblewriter/patches/blob/2d55734bb5034d77c4f938ed987804fd1b522cf3/tests/json-patch/txt-retain-overrun.spec.ts#L22-L26)), a state this very function guarantees can never persist. With a real 1269-char doc the overrun is 10, not 11, and the expectation at [L41-L43](https://github.com/dabblewriter/patches/blob/2d55734bb5034d77c4f938ed987804fd1b522cf3/tests/json-patch/txt-retain-overrun.spec.ts#L41-L43) changes shape.

src/json-patch/ops/text.ts
```diff
@@ -106,21 +117,20 @@ function fixBadDeltaDoc(delta: Delta): Delta {
...
   if (!hasNonInsertOps) {
     return delta;
   }
-  const newDelta = new Delta();
+  log(
+    'Dropping ops that overran the document in a @txt doc',
+    delta.ops.filter(op => op.insert === undefined)
+  );
 
+  const newDelta = new Delta();
```
RyanHavoc: Now that both halves drop non-inserts, the function makes four passes (`while`-pop, `.some()`, the log's `.filter()`, this rebuild) for "keep the inserts" — one pass can collect the dropped ops and reuse them for the log. One caveat if taking that: the rebuild must keep going through `Delta.push`, which merges adjacent inserts ([json-patch.spec.ts L209-L211](https://github.com/dabblewriter/patches/blob/2d55734bb5034d77c4f938ed987804fd1b522cf3/tests/json-patch/json-patch.spec.ts#L209-L211) depends on the merged `'Short\nLong\n'`); `new Delta(ops.filter(...))` is not equivalent.

tests/json-patch/txt-retain-overrun.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+ * manuscript rather than merely misplacing one edit.
+ *
+ * The sequence below is taken verbatim from a reporter's database export (co-authoring
+ * session, 2026-08-18): doc length 1269, two successive changes both retaining 1280.
+ */
+describe('@txt with a retain past the end of the document', () => {
+  // Reproduces the exact offsets from the export: 1258 chars of prior text, then the
+  // author types "<I'm never " (11 chars), leaving the document 1269 long.
+  const PRIOR = 'x'.repeat(1258);
+  const OPENING = "<I'm never ";
```
RyanHavoc: These strings are verbatim from the reporter's manuscript — `impulsive.>` and `' Except for with Aster.'` include a character name — and the header records the dated database-export provenance ([L16-L19](https://github.com/dabblewriter/patches/blob/2d55734bb5034d77c4f938ed987804fd1b522cf3/tests/json-patch/txt-retain-overrun.spec.ts#L16-L19)). This repo is public, so an identifiable fragment of an unpublished novel becomes permanent public git history on merge. Only the offsets (1258/11/1269/1280) carry test value — neutral strings lose zero coverage. Cheap to fix now, effectively impossible after.

RyanHavoc (commented):
Deep review pass — 13 notes inline, roughly most severe first. The direction is right: padding is the corruption mechanism, and the apply-path sanitizer is the correct layer (server-side rejection would need document state at commit, which the architecture deliberately avoids). But several paths still assume the padded frame — invert/undo, committed-history replay, poison ejection, and composePatch — and the new test fixture carries verbatim reporter text into a public repo. Details inline.

tests/json-patch/txt-retain-overrun.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+    // subsequent offset; that is the mechanism behind the reported transposition.
+    expect(doc.length()).toBe(textOf(result.text).length);
+  });
+
+  it('does not transpose the next edit (the reported corruption)', () => {
+    let state: any = { text: new Delta().insert(docText).ops };
+
+    // Both changes retain 1280 — the second is correct in the author's own frame.
+    state = applyTxt(state, [{ retain: 1280 }, { insert: 'impulsive.>' }]);
+    state = applyTxt(state, [{ retain: 1280 }, { insert: ' Except for with Aster.' }]);
```
RyanHavoc: This passes by arithmetic coincidence: the insert length equals the overrun (11 == 11), so after the first apply the doc is 1281 chars (1269 + 11 + `\n`) and this second `retain 1280` is **in-bounds** — it never exercises the drop path ([L68-L70](https://github.com/dabblewriter/patches/blob/2d55734bb5034d77c4f938ed987804fd1b522cf3/tests/json-patch/txt-retain-overrun.spec.ts#L68-L70)). Make change 1 insert `'ab'` instead: change 2 overruns by 8, gets dropped, and lands after the terminal newline on a new paragraph — the symptom this test's name asserts against. A case at the client's actually-advanced frame (`retain 1291`) would pin the real successor behavior: dropping misplaces the first edit *and* every successor until resync, where padding self-corrected after one.

tests/json-patch/txt-retain-overrun.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+    // Reported symptom: "impulsive.>" is shunted to the end, behind text typed after it.
+    expect(text).not.toMatch(/Except for with Aster\.impulsive\.>/);
+    expect(text).toContain('impulsive.> Except for with Aster.');
+  });
+
+  it('still preserves a retain that is within the document', () => {
+    const state = { text: new Delta().insert(docText).ops };
+
+    // Guard against over-correcting: an in-bounds retain is ordinary and must still work.
+    const result = applyTxt(state, [{ retain: 1269 }, { insert: 'impulsive.>' }]);
```
RyanHavoc: Three ways this suite can't fail ([L79-L85](https://github.com/dabblewriter/patches/blob/2d55734bb5034d77c4f938ed987804fd1b522cf3/tests/json-patch/txt-retain-overrun.spec.ts#L79-L85)):

1. `retain(1269)` equals the doc length — a boundary where no stranded op is ever produced — so this passes even if the drop loop discarded in-bounds retains and truncated everything after them. A mid-document `retain(600)+insert(...)` asserting the trailing 669 chars survive is the guard this claims to be.
2. `applyTxt` (L33) omits `{strict: true}`; the default path skips a failed op with a `console.error`, so "leaves no non-insert op stranded" and "keeps the document length honest" both pass on the untouched input if `@txt` stops applying entirely.
3. The length assertion (L62) is tautological for string-only fixtures, and false for docs with embeds — `textOf` skips non-string inserts while `Op.length` counts an embed as 1.

Also uncovered: stranded `{delete: n}`, attributed retains, and any embed on the drop path.

tests/json-patch/txt-retain-overrun.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+ * session, 2026-08-18): doc length 1269, two successive changes both retaining 1280.
+ */
+describe('@txt with a retain past the end of the document', () => {
+  // Reproduces the exact offsets from the export: 1258 chars of prior text, then the
+  // author types "<I'm never " (11 chars), leaving the document 1269 long.
+  const PRIOR = 'x'.repeat(1258);
+  const OPENING = "<I'm never ";
+  const docText = PRIOR + OPENING;
+
+  const textOf = (value: unknown) =>
```
RyanHavoc, line 36: Fifth copy of `textOf` ([L27-L31](https://github.com/dabblewriter/patches/blob/2d55734bb5034d77c4f938ed987804fd1b522cf3/tests/json-patch/txt-retain-overrun.spec.ts#L27-L31); others in rebaseChanges.spec.ts, ejectPendingChange.spec.ts:26, transformIncomingChanges.spec.ts:14, and `docBody` in OTBranchManager.integration.spec.ts), and `applyTxt` re-declares the `@txt` builder three other specs already declare. `tests/fuzz` already hosts shared helpers cross-imported by several specs, so an established home exists. The identical state + `retain 1280` fixture is also rebuilt in five `it` blocks — one describe-scoped result would keep the load-bearing 1280/1269 relationship in one place.

src/algorithms/ot/shared/applyChanges.ts
```diff
@@ -126,7 +126,13 @@ export function applyChangesForReconstruction<T>(state: T, changes: Change[], op
...
     const change = changes[i];
     try {
-      state = applyPatch(state, change.ops, { strict: true });
+      // `legacyTextOverrunPadding` belongs to this function and nowhere else. A committed log
+      // may contain a `@txt` change that overran the document, and the edits after it were
+      // authored against the padding that overrun produced. Replaying under the live rule —
+      // which drops the overrun — applies those later edits to a document that never had the
+      // padding, so an in-bounds delete lands on real prose. Settled history has to replay
+      // under the semantics it was written with. See ApplyJSONPatchOptions.
+      state = applyPatch(state, change.ops, { strict: true, legacyTextOverrunPadding: true });
```
jacwright: This is the line I'd like changed, because one of this function's callers isn't replaying history for display — it's minting a new document.

`OTBranchManager.createBranch` (`OTBranchManager.ts:549`) materializes the branch point through `getStateAtRevision(..., { reconstruction: {} })`, wraps the result in a root `replace`, and commits it:

```js
const rootReplace = createChange(0, 1, [{ op: 'replace', path: '', value: stateAtRev }], …);
await this.store.saveChanges(branchDocId, initChanges);
```

So with the flag on here, the padding is written into a brand-new document as committed content. Measured at this SHA:

```
live head (OTDoc.applyChanges, no flag) : "Hello\nX\n"
branch base (createBranch, reconstruction): "Hello\n              X\n"
                                             ^ 14 invented spaces, now real text
```

The branch is 14 characters longer than the document it was branched from, at the same revision. And because review copies are branches and branches merge back, those characters can return to the source as ordinary prose.

The reason the docblock's argument doesn't cover this: it's about later edits in the *same log* having been authored against the padded text. A branch init takes one snapshot and starts a **new** log — nothing downstream of it was authored against the padding, so the condition that makes padding the faithful choice is absent exactly here.

Suggestion: promote it to an explicit field on `ReconstructionOptions` that `PatchesHistoryManager` and `PatchesHistoryClient` set, and `createBranch` doesn't. That encodes the real distinction — replay you *show* versus replay you *persist* — in the type, instead of leaving it as a property of which function you happened to call.

Control, for the record: removing the flag from this line fails **5 tests**, including both behavioural reconstruction cases. The contract is genuinely pinned, which is why I'm asking for the caller to opt out rather than for the flag to go away.

src/json-patch/types.ts
```diff
@@ -40,6 +40,19 @@ export interface ApplyJSONPatchOptions {
...
+   * overran the document: the overrun is materialised as padding spaces instead of dropped.
+   *
+   * That padding is the DAB-1064 corruption mechanism, so live applies must never opt in. It
+   * survives for replay because committed history has to reproduce what clients actually
+   * computed when the change was first applied — a doc whose log contains an overrun, and then
+   * an edit made against the padded text, only replays coherently under the old rule. Changing
+   * the semantics under replay would silently rewrite settled documents (see
+   * `applyChangesForReconstruction`).
+   */
+  legacyTextOverrunPadding?: boolean;
```
jacwright, line 55: This docblock is the right place to record the constraint, so it's worth tightening one phrase in it.

> HISTORICAL-RECONSTRUCTION ONLY … Changing the semantics under replay would silently rewrite settled documents

"Settled" is doing the work here, and it's accurate for the history surfaces. The gap is that `applyChangesForReconstruction` also feeds `OTBranchManager.createBranch`, where the output isn't settled history being read — it's the seed content of a new document being written (`saveChanges(branchDocId, initChanges)`).

If the flag stays as-is, this comment is the one thing standing between the current behaviour and someone adding a third reconstruction caller that also persists. I'd make it say what must not happen rather than which function may ask:

> Only replay whose output is **displayed** may set this. Replay whose output is **persisted as new state** (branch initialization, snapshot seeding) must not — the padding would become authored content in a document that never had it.

That framing also answers the next question by itself, which the current wording doesn't: a reader who knows `createBranch` calls reconstruction has no way to tell from here whether that's intended.

src/json-patch/ops/text.ts
```diff
@@ -89,14 +89,44 @@ export const text: JSONPatchOpHandler = {
...
+ * a formatting op that overran produced n spaces *carrying that formatting*. Invented
+ * characters that are also styled. Dropping handles both shapes the same way, which is why
+ * "could we at least keep formatting retains?" is not a rescue.
+ *
+ * What the drop costs: since a stored document always ends in a newline, the overrunning
+ * edit lands *after* that newline — its own paragraph at the end of the document, not at
+ * the author's caret. That is the accepted trade. A visibly misplaced edit the author can
+ * see and move beats silent corruption of text they have already written.
+ *
+ * `legacyPadding` keeps the old behaviour, and ONLY historical reconstruction may ask for it.
```
jacwright, line 116: Small thing, and only worth doing if you're touching this docblock anyway.

> `legacyPadding` keeps the old behaviour, and ONLY historical reconstruction may ask for it.

Verified that the legacy branch below reproduces `origin/main` exactly — trailing non-inserts popped, mid-document retains padded with their attributes preserved, stranded deletes dropped by falling through. That fidelity is the whole point, so it's good that it's exact rather than approximately equivalent.

The sentence I'd revisit is this one, for the same reason as the `types.ts` comment: "historical reconstruction" names a function, and the constraint is really about what happens to the output. `createBranch` is historical reconstruction by that name and persists the result as new content.

Separately — the paragraph explaining what the drop *costs* (the edit lands after the document's trailing newline, as its own paragraph, rather than at the caret) is a genuinely good addition. It's the part a support person will need when an author asks why their sentence moved, and it would have been easy to leave as "the edit lands at the end."

matalina:
You're right, and the reproduction made it obvious. Fixed in `00341b5`.

`createBranch` persisting reconstructed state is the case my "reconstruction is its only caller" line didn't cover, and the reasoning for why the padding rationale doesn't extend there is exactly right: the justification is log-internal coherence, and a branch init starts a fresh log, so the one condition that makes padding correct is absent precisely there.

Took option (1). `legacyTextOverrunPadding` now lives on `ReconstructionOptions`, defaults off, and each caller decides:

| caller | opts in? | why |
| --- | --- | --- |
| `PatchesHistoryClient.scrubTo` | yes | renders this log |
| `PatchesHistoryManager` (state-before-version stream) | yes | renders this log |
| `createBranch` | **no** | seeds a new document; branch base now matches what clients compute |
| branch-doubling guard (`OTBranchManager:1254`) | no | compares text, never persists it |

Defaulting off also picks the safer failure: a caller that forgets gets a historical view differing from what the author saw, rather than corruption baked into live content. New test pins the branch case specifically — that seeded output has no padding *and* is byte-identical to what `applyChanges` produces for the same log, since "branch disagrees with parent" is the failure worth naming.

## One more with the opposite hazard, and it's outside this repo

`buildVersionState` is exported and no caller in patches uses it — pup builds the blobs. A version blob is a base for **further replay of the same log**, so it needs the padding: build a blob with live semantics from a log containing an overrun, and the changes recorded after it — authored against that padding — misapply on top of the blob. Same failure as the one you caught, opposite direction.

I've documented it on `buildVersionState` where a consumer would look, but I deliberately haven't changed pup. Worth deciding whether that opt-in should be part of the same release, since it's the one remaining path where the default is wrong for the caller.

## On the test counts

Thanks for re-checking, and no apology needed — the misdiagnosis was reasonable given a clean `npm ci` really does pass on your side. `3510 + 35 + 4 = 3549` reconciling with your total is what convinced me too. It's now 3511 on Windows with this commit.

Everything else you verified matches what I have: legacy mode byte-identical to `origin/main`, the warn's firing surface (only the genuine mid-document overrun shape reaches it), and the flag off by default.

The `compose` / `invert` / eject-probe notes from the earlier pass are tracked in DAB-1119 rather than folded in here. And the mid-document double-count — valid in-bounds offset, no overrun, no warning, no guard anywhere — is now written up prominently on DAB-1064 as its own section, since that's the one this PR can't help with.

src/server/OTBranchManager.ts
```diff
@@ -546,6 +546,12 @@ export class OTBranchManager implements BranchManager {
       // lenient-era commit is skipped — matching what version builds and
       // pre-strict clients computed for the same log — rather than making the
       // source doc permanently un-branchable. Skips log via console.error.
+      //
+      // Deliberately WITHOUT `legacyTextOverrunPadding`: this state is persisted as the new
+      // branch's first change, so it is authored content, not a rendering of the source log.
+      // The branch's history begins here — nothing downstream was written against the source's
+      // overrun padding — and padding it would bake invented characters into the branch as
+      // ordinary text, which a merge could then carry back into the source (DAB-1064).
       const { state: stateAtRev } = await getStateAtRevision(this.store, docId, rev, { reconstruction: {} });
```
jacwright, line 555: This is right, and the comment above it is the clearest statement of the rule anywhere in the PR — "this state is persisted as the new branch's first change, so it is authored content, not a rendering of the source log."

The gap is that nothing enforces it. Control:

```js
getStateAtRevision(this.store, docId, rev, { reconstruction: { legacyTextOverrunPadding: true } })
```

```
Tests  3546 passed | 4 skipped (3550)
```

The whole suite passes with the previous revision's bug restored. `does not pad unless the caller opts in` covers the library default, which is the right test to have, but it can't see what this line chose — so the decision here rests entirely on the comment surviving future edits.

Worth closing because the failure mode is invisible: nothing throws, no test reddens, and the symptom shows up later as a review copy that doesn't match its parent. That's a long way from the edit that caused it.

You already have the pattern for the cheap version — `applyChanges.spec.ts` asserts the exact options object passed down, which is what caught my attention as a good move in the last round. The same assertion on `getStateAtRevision` in `OTBranchManager.spec.ts` pins this line specifically.

The stronger version, if `OTBranchManager.integration.spec.ts` makes it easy: branch a doc whose committed log contains an overrun and assert the branch base equals the source's live head. That asserts the user-facing property rather than the argument, and it's the same invariant your new reconstruction test uses — which I thought was the best-chosen assertion in this diff.

src/algorithms/ot/server/buildVersionState.ts
```diff
@@ -252,6 +252,18 @@ export async function getStateBeforeVersionAsStream(
  * @param docId - The document ID.
  * @param version - The version metadata.
  * @param changes - The changes included in this version.
+ * A version blob is a base for FURTHER replay of the same log, so it is a rendering of that
```
jacwright, line 251: This paragraph is correct and it's the one that needs to escape the docblock, because the consumer it's addressed to doesn't currently do it.

pup builds every version blob through here:

```js
// pup src/store/FirestoreStore.ts:100
const state = await buildVersionState(store, docId, version, changes, {
  reconstruction: { onSkippedChange: skipped => reportSkippedChange('version-build', docId, skipped) },
});
```

No `legacyTextOverrunPadding`. That was fine while padding was unconditional; with the default flipped it means pup's blobs change semantics the moment its pin moves.

And it isn't confined to history views — `getSnapshotStream` hands the stored blob to clients as the base for ordinary document loads:

```js
const { rawState, versionRev } = await getLatestMainVersion(store, docId, rev);
const changes = await store.listChanges(docId, { startAfter: versionRev, … });
```

I want to be careful not to overstate the blast radius: I haven't worked through every blob-boundary case, and a build whose base is an already-padded parent blob may well come out consistent. What I can say is that for a doc with an overrun in its log, whether that range's blob was built with padding determines whether the author's later delete of the stray spaces removes padding or eats prose — and nothing about that decision is visible from pup's call.

Two suggestions:

- Put the pup change in the shipping chain alongside the manual publish and the pin bumps, rather than leaving it as a note here. This repo publishes by hand, so there's no CI step that would catch a consumer that didn't follow the docblock.
- Consider whether the default is the right way round for this one function specifically. Everywhere else, off-by-default fails safe — the cost is "a historical view that differs from what the author saw," as `ReconstructionOptions` puts it. Here the cost is a stored blob that later changes misapply onto, which is a different order of thing. Not asking you to change it, but it's the one place where the safe default may point the other way.

Separately, small: the new paragraph landed between the existing `@param` entries and the list got re-emitted below it, so `@param store` / `docId` / `version` / `changes` each appear twice (lines 250-266). Moving this paragraph above the first `@param` sorts it.

src/algorithms/ot/shared/applyChanges.ts
```diff
@@ -81,6 +81,28 @@ export interface ReconstructionOptions {
...
+   *   and an in-bounds delete authored against the padding lands on real prose instead.
+   * - Seeding a new document from a past revision (branch/review-copy creation) → leave OFF.
+   *   The new document's history begins at the snapshot, so nothing downstream depends on
+   *   the padding — and padding it would bake invented characters into a branch as ordinary
+   *   authored text, which can then merge back into the source (DAB-1064).
+   *
+   * When in doubt, leave it off: the cost is a historical view that differs from what the
+   * author saw, not corruption of live content.
+   */
+  legacyTextOverrunPadding?: boolean;
```
jacwright, line 105: This docblock is the part of the change I'd most want to keep intact. Naming the distinction as *rendering this log* versus *seeding a new one* is what makes the right answer derivable at a new call site, where "reconstruction vs live" left the reader to guess — and guessed wrong once already.

The "when in doubt, leave it off" line is doing real work too, and it's correctly asymmetric: off costs a historical view that differs from what the author saw, on costs invented characters committed as authored text. Stating which way the failure leans is more useful than stating the rule.

One observation rather than a request. That asymmetry holds for every caller in this repo, but not quite for `buildVersionState` — a blob built off is not a view that merely looks wrong, it's a base that later changes misapply onto. Its own docblock says as much. If that's right, then the guidance here is sound for the general case but has one documented exception, and it may be worth a clause saying so, since the sentence currently reads as universal.

Verified while checking this: removing the opt-in from `PatchesHistoryClient` fails 2 tests, so the rendering side of the boundary is pinned. The seeding side isn't, which I've left a note about on `OTBranchManager.ts:555`.

matalina:
All three addressed in `0b91c9e`.

## 1. The call-site decision is now pinned

You were right that only the default was covered. New case in `OTBranchManager.spec.ts` branches a document whose committed log contains an overrun and asserts the persisted branch base equals what `applyChanges` computes for the same log.

Ran your control before trusting it:

```
flag flipped on at OTBranchManager:555  →  1 failed  (seeds the branch with live semantics…)
reverted                                →  3512 passed
```

I took your framing on the assertion — comparing against the live head rather than "has no double spaces" — because it states the property that actually matters and keeps holding if the padding is ever spelled differently.

## 2. pup prerequisite recorded

Verified against pup's source; the call at `FirestoreStore.ts:100` is exactly as you quoted, with no `legacyTextOverrunPadding`.

Written onto DAB-1064 as a **ship prerequisite** with its own heading, next to the manual publish and the pin bumps — including your point that `getSnapshotStream` serves the blob for live document loads, not only scrubbing, which is what makes it more than a history-view concern. The action recorded is: add the flag to that call in pup and land it **with or before** the pin bump, not after. I kept your caveat that the blast radius isn't fully traced, since "silent change to stored state" is the argument on its own.

Deliberately not changing pup from this PR.

## 3. Duplicated `@param` block

Mine, from the previous commit — the new prose landed inside the existing param list rather than above it, so four params were emitted twice. Moved above the first `@param`.

---

Suite is 3512 on Windows with this commit (3546 + the new case on a clean install, by your arithmetic). `type:check`, `lint`, `format:check` clean.

On the last point in your review — I'd add that the reason the call sites needed to say *why* is that I got this wrong once already by reasoning about the function name instead of what the output is used for. The comment at `createBranch` exists because "reconstruction" was the wrong thing to key on, and the next person deserves to know that rather than rediscover it.

tests/server/OTBranchManager.spec.ts
```diff
@@ -197,6 +199,46 @@ describe('OTBranchManager', () => {
       expect(result).toBe('generated-id');
     });
 
+    it('seeds the branch with live semantics, not the log-rendering padding (DAB-1064)', async () => {
```
jacwright, line 202: This closes the gap properly. Verified by re-applying the exact regression:

```
× seeds the branch with live semantics, not the log-rendering padding (DAB-1064)
  AssertionError: expected 'Hello\n              X\n' to be 'Hello\nX\n'
```

Last round that same edit passed 3546/3546, so this is the difference between a decision held by a comment and one held by the suite.

Two things I'd point out as worth keeping if this test is ever refactored:

Driving it through `createBranch` and recovering the root `replace` from the `createChange` calls means it pins the *outcome* — what actually gets persisted — rather than which options object was passed down. An options assertion would have caught this particular regression too, but would go green if the threading changed shape while still padding. This one can't.

And `expect(textOf(branchBase.text)).toBe(textOf(liveHead.text))` is the assertion carrying the real weight. The two literal checks above it would both pass for a branch base that was wrong in some other way; deriving the expected value from `applyChanges` on the same log states the invariant itself — a branch and its parent must agree — instead of a value that happens to encode it today.

